PR #5, opened 2015-03-06
use ament_lint_auto

dirk-thomas:
Update `rosidl_generator_c` and `rosidl_parser` to use `ament_lint_auto` (depends on ament/ament_lint#2).

Note: adding all current linters will result in a lot of failing tests.

Connects to ros2/ros2#3

### DIFF
--- a/rosidl_cmake/CMakeLists.txt
+++ b/rosidl_cmake/CMakeLists.txt
@@ -4,6 +4,11 @@ project(rosidl_cmake NONE)
 
 find_package(ament_cmake REQUIRED)
 
+if(AMENT_ENABLE_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
+endif()
+
 ament_package(
   CONFIG_EXTRAS "rosidl_cmake-extras.cmake"
 )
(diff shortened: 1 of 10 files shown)


wjwwood:
ros2#3

wjwwood:
+1

tfoote:
+1

esteve:
+1

dirk-thomas:
I would recommend to *not* merge this but instead use it as a template and apply it to packages once the  tests actually (have a chance to) pass. Currently the number of failing tests is just to high and would create a lot of noise but very few help until we adjust the linter configurations.

dirk-thomas:
Forget my last comment - we should just merge it - otherwise nobody is going to try it from the branch of this PR. I also added it for the other packages in this repo. Until we have CI it shouldn't bother much - and then we should really get the tests to zero failures.

esteve:
What card/ticket is this connected to? ros2#3 is not a valid link.

dirk-thomas:
ros2/ros2#3

wjwwood:
I think it cannot be done retroactively, or at least I was unable to. You can see I added the connection text to the original description trying to get it to do so.

esteve:
@wjwwood it worked, this PR shows up on the ros2/ros2#3 card on my waffle

wjwwood:
Maybe it has to be the person who created it? Oh well seems to be working, awesome.

PR #6, opened 2015-03-06
Added support for RTI Connext

esteve:
This PR enables the RTI Connext middleware components

Connects to ros2/rmw_connext#3


tfoote:
+1

PR #7, opened 2015-03-12
Windows support

wjwwood:
Fixes for compiler warnings and errors. 

Connects to ros2/ros2#9

### DIFF
--- a/rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
+++ b/rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
@@ -133,9 +133,7 @@ def value_to_cpp(type_, value):
     for single_value in value:
         cpp_value = primitive_value_to_cpp(type_, single_value)
         cpp_values.append(cpp_value)
-    cpp_value = '{%s}' % ', '.join(cpp_values)
-    if type_.array_size is not None and not type_.is_upper_bound:
-        cpp_value = '{%s}' % cpp_value
+    cpp_value = '{{%s}}' % ', '.join(cpp_values)
     return cpp_value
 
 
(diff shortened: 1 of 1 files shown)


esteve:
+1

dirk-thomas:
lgtm

wjwwood:
Please have a look at 32d0441, it solves ros2/rosidl#8 for me on Windows.

wjwwood:
The Windows CI job is green: http://54.183.26.131:8080/job/ros2_batch_ci_windows/

dirk-thomas:
Please add a comment with the link to the MS knowledge base article above the line to describe why it is done that way.

I was surprised for a second how the job could be green. It looks like does neither compile not run any tests: https://github.com/ros2/ros2/issues/6#issuecomment-82748716

wjwwood:
> Please add a comment with the link to the MS knowledge base article above the line to describe why it is done that way.

There's no knowledge base article, it actually just makes more sense I think. You just have to think about the elements in the vector as being initialized separately from the vector, like this:

```c++
std::vector<int> a = {-5};  // Is implicitly {{-5}}
std::vector<int> b = {-5, 5}; // Is implicitly {{-5}, {5}}
```

This is perhaps a better way to think about it:

```c++
struct A {
  int foo;
  char bar;
};

std::vector<struct A> vec = {{1, 'a'}, {2, 'b'}};
```

But in the case of single values for types which can convert to `size_t`, the Windows compiler is sort of correct because it's not clear if `vector<int>({5})` should be expanded to `vector<int>(5)` or `vector<int>({{5}})`. And the former becomes ambiguous with the third std::vector constructor because it can be converted to a `size_t`:

http://en.cppreference.com/w/cpp/container/vector/vector (constructor 3)

We're already doing this syntax for our `std::arrays`, so my patch just removes the array specific logic and wraps values in `{}` as well as a containing set of `{}` for the container structure (vector or array).

PR #9, opened 2015-03-25
Fix get type support handle

wjwwood:
This pull request changes how the implementation specific version of the `get_type_support_handle` function is included. Each vendor specific implementation provides a custom version of this function in a header called `rosidl_generator_cpp/get_type_support_handle_impl.hpp` and the `rosidl_generator_cpp` package provides a header called `rosidl_generator_cpp/get_type_support_handle.h` which the client library imports. A different `rosidl_generator_cpp/get_type_support_handle_impl.hpp` is included based on the include paths set when building a library or executable, allowing you to select an implementation that way.

I've also renamed the implementation specific version of the function to `get_type_support_handle_impl`, to reduce confusion and improve searching, but that function is defined in the various implementations not here.

Connects to ros2/ros2#26

### DIFF
--- a/rosidl_generator_cpp/include/rosidl_generator_cpp/MessageTypeSupport.h
+++ b/rosidl_generator_cpp/include/rosidl_generator_cpp/MessageTypeSupport.h
@@ -17,13 +17,22 @@
 #define __rosidl_generator_cpp__MessageTypeSupport__h__
 
 #include <rosidl_generator_c/message_type_support.h>
+#include <rosidl_generator_c/visibility_control.h>
 
 namespace rosidl_generator_cpp
 {
 
 template<typename T>
+ROSIDL_PUBLIC
 const rosidl_message_type_support_t * get_type_support_handle();
 
 }  // namespace rosidl_generator_cpp
+
+// Tail include the implementation for this function.
+// This include is provided by the implementation and therefore only
+// one version of this header will get included from the path.
+// This header maps the get_type_support_handle function to an implementation
+// specific version of that function.
+#include <rosidl_generator_cpp/get_type_support_handle_impl.hpp>
 
 #endif  // __rosidl_generator_cpp__MessageTypeSupport__h__
--- a/rosidl_generator_c/include/rosidl_generator_c/message_type_support.h
+++ b/rosidl_generator_c/include/rosidl_generator_c/message_type_support.h
@@ -16,12 +16,14 @@
 #ifndef ROSIDL_GENERATOR_C_ROSIDL_GENERATOR_C_MESSAGE_TYPE_SUPPORT_H_
 #define ROSIDL_GENERATOR_C_ROSIDL_GENERATOR_C_MESSAGE_TYPE_SUPPORT_H_
 
+#include "visibility_control.h"
+
 #ifdef __cplusplus
 extern "C"
 {
 #endif
 
-typedef struct rosidl_message_type_support_t
+typedef struct ROSIDL_PUBLIC rosidl_message_type_support_t
 {
   const char * typesupport_identifier;
   const void * data;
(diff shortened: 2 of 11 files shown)


esteve:
+1

We need to address `get_service_type_support`, but we can do that in a future PR.

tfoote:
+1

rosidl_generator_cpp/include/rosidl_generator_cpp/MessageTypeSupport.h
```diff
@@ -17,13 +17,22 @@
...
 const rosidl_message_type_support_t * get_type_support_handle();
 
 }  // namespace rosidl_generator_cpp
 
+// Tail include the implementation for this function.
+// This include is provided by the implementation and therefore only
+// one version of this header will get included from the path.
+// This header maps the get_type_support_handle function to an implementation
+// specific version of that function.
+#include <rosidl_generator_cpp/get_type_support_handle_impl.hpp>
```
dirk-thomas: I am not convinced that this is a good idea.
Basically it makes the package not "includable" anymore without a specific implementation header on the path.
These kind of "inverse" dependencies should be avoided.

I think we should turn this around: user land code should import the implementation file.
And `rosidl_generator_cpp` should only contain the forward definition as well as a dummy implementation file containing an `#error` statement that lets the user know that he needs a package providing an implementation of this.

wjwwood: I started out with what you are describing, but I changed to this for a few reasons. I'm willing to change it again, but I think some of the rational are compelling:

Being not includable without the correct include path set has actually helped to find places where this is used and the include path should be set, but isn't. I can't think of a situation where you would want to include this header, but didn't or couldn't setup the include path to something reasonable in order to get the implementation header.

The alternative strategy also results in non-includable files unless special include paths are set. Currently the consumer of this header includes just this header and must have the include paths set. In the other arrangement, consumers of this header always need to additionally include the implementation header, which can only be done with special include paths setup. So you include two headers instead of one with the same constraints.

The diff is pretty trivial, basically you remove this include line and comments and add an identical include line everywhere this header is included.

dirk-thomas: I think we can easily keep the advantages you describe:

* single include for user land code
* fail nicely when no implementation is available

but still avoid the disadvantage of the current approach:

* header of this package is not includable
* header of this package references a file which doesn't exist (within the scope of this package and its dependencies)

This can be done by:

* user land includes the impl header
  * if no implementation is on the include path it will pick up the impl header file from this package giving a really good error using `#error`
  * otherwise use the impl header of the implementation package, this had can itself include the header from this package containing the forward declaration

wjwwood: Ok, I get the idea I think, but this comment doesn't make sense to me:

> if no implementation is on the include path it will pick up the impl header file from this package giving a really good error using #error

There is no `get_type_support_handle_impl.hpp` in this package, so if user land code (just `rclcpp` for now) includes `rosidl_generator_cpp/get_type_support_handle_impl.hpp` and the include path is not set to one of the implementation's special path, it will not find any header, so there is nothing to put a `#error` in.

dirk-thomas: Sorry, difficult to describe without real code.

I was suggesting to create the file `rosidl_generator_cpp/get_type_support_handle_impl.hpp` in this package and only include an `#error` directive letting the user know that it lacks a proper implementation of this and possibly point him to some packages which provide it.

wjwwood: Ok, that makes sense to me. I'll try that out, but it will take some time to test it. After lunch.

PR #15, opened 2015-04-21
fix visibility of service typesupport on Windows

dirk-thomas:
Connects to ros2/ros2#27.


wjwwood:
+1 (with ros2/rosidl_dds#15)

tfoote:
+1

PR #17, opened 2015-05-01
add comparison operators

dirk-thomas:

### DIFF
--- a/rosidl_generator_cpp/resource/msg_Struct.h.template
+++ b/rosidl_generator_cpp/resource/msg_Struct.h.template
@@ -2,7 +2,7 @@
 @#
 @# ROS message struct code generation for C++
 @#
-@# EmPy template for generating <msg>.h files
+@# EmPy template for generating <msg>_Struct.h files
 @#
 @###############################################
 @# Start of Template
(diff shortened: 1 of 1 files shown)


rosidl_generator_cpp/resource/msg_Struct.h.template
```diff
@@ -106,6 +106,26 @@ for field in spec.fields:
...
+  bool operator==(const @(spec.base_type.type)_ & other) const
+  {
+@[for field in spec.fields]@
+    if (this->@(field.name) != other.@(field.name)) {
+      return false;
+    }
+@[end for]@
+    return true;
+  }
+  bool operator!=(const @(spec.base_type.type)_ & other) const
```
esteve: You could reuse the `==` operator like this:

```cpp
bool operator!=(const @(spec.base_type.type)_ & other) const
{
  return !(*this == other);
}
```

esteve:
+1 either with my proposed change or not.

rosidl_generator_cpp/resource/msg_Struct.h.template
wjwwood: Makes sense, rather not have duplicated and then inverted logic.

wjwwood:
+1, ditto.

rosidl_generator_cpp/resource/msg_Struct.h.template
dirk-thomas: Good point to get rid of the duplicate logic. I updated it to use the following since we don't want lhs conversion here:

    return !this->operator==(other);

tfoote:
+1

Issue #12, opened 2015-04-15
Improve validation and error message when parsing invalid srv


tfoote:
When I had a typo in a service referencing an non-existant message as a subtype, it made it all the way through the generation and failed on missing include directories when actually trying to compile the c++ type support. 

We should find a way to validate this earlier and catch that it's referencing a message which does not exist and provide an error message to that effect before running generation for n languages and trying to compile some of them. 

For the moment this is a low priority but should be fixed before release. 

For reference this is the error shown for an invalid reference to Parameters which should be Parameter
```
In file included from /home/tfoote/work/ros2/parameters/build/rcl_interfaces/rosidl_generator_cpp/rcl_interfaces/GetParametersResponse.h:4:0,
                 from /home/tfoote/work/ros2/parameters/build/rcl_interfaces/rosidl_generator_cpp/rcl_interfaces/GetParameters_Service.h:5,
                 from /home/tfoote/work/ros2/parameters/build/rcl_interfaces/rosidl_typesupport_introspection_cpp/rcl_interfaces/GetParameters_ServiceTypeSupport_Introspection.cpp:10:
/home/tfoote/work/ros2/parameters/build/rcl_interfaces/rosidl_generator_cpp/rcl_interfaces/GetParametersResponse_Struct.h:12:39: fatal error: rcl_interfaces/Parameters.h: No such file or directory
 #include <rcl_interfaces/Parameters.h>
                                       ^
compilation terminated.
[  7%] Generating C++ type support for PrismTech OpenSplice
make[2]: *** [CMakeFiles/rcl_interfaces__typesupport_introspection_cpp.dir/rosidl_typesupport_introspection_cpp/rcl_interfaces/GetParameters_ServiceTypeSupport_Introspection.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/rcl_interfaces__typesupport_introspection_cpp.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
INFO com.rti.ndds.nddsgen.Main Running rtiddsgen version 2.2.0.0, please wait ...
WARN com.rti.ndds.nddsgen.antlr.auto.IdlLexer ParameterDescription_.idl line 35  preprocessor directive not supported. It will be ignored
WARN com.rti.ndds.nddsgen.antlr.auto.IdlLexer Parameter_.idl line 30  preprocessor directive not supported. It will be ignored
INFO com.rti.ndds.nddsgen.Main Done
[...]

PR #18, opened 2015-05-22
enable checking message dependencies

dirk-thomas:
Connects to ros2/rosidl#12


tfoote:
+1

dirk-thomas:
If there are any further comment I will address them after the PR is being merged.

PR #23, opened 2015-06-17
register interfaces with ament index

dirk-thomas:
Connects to ros2/rosidl#22

@ros2/owners Please review.

### DIFF
--- a/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
+++ b/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
@@ -110,6 +110,18 @@ macro(rosidl_generate_interfaces target)
     ${_idl_files}
   )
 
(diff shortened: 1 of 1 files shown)


esteve:
+1

tfoote:
+1

wjwwood:
+1

Just to be clear the contents of the file are just the names of the interfaces, but not the file ext or the path? Couldn't that cause problems with overlapping interfaces, e.g. Empty.msg and Empty.srv in the same package?

dirk-thomas:
Therefore the content of the ament index contains the filenames including the extension.

mikaelarguedas:
@dirk-thomas  do we still need this branch?

dirk-thomas:
No, deleted. Thanks for pointing it out.

PR #25, opened 2015-06-17
use more strict rules for identifiers

dirk-thomas:
The patch enables the stricter rules (instead of using the relaxed one to be able to read all ROS 1 messages). It also makes the regular expressions even more strict to match the spec from the article.

Please review.

### DIFF
--- a/rosidl_parser/rosidl_parser/__init__.py
+++ b/rosidl_parser/rosidl_parser/__init__.py
@@ -47,14 +47,14 @@ PRIMITIVE_TYPES = [
     'time',  # for compatibility only
 ]
 
-VALID_PACKAGE_NAME_PATTERN = re.compile('^[a-z][a-z0-9_]*$')
-# TODO replace relaxed patterns used for compatibility
-# VALID_FIELD_NAME_PATTERN = re.compile('^[a-z][a-z0-9_]*$')
-VALID_FIELD_NAME_PATTERN = re.compile('^[A-Za-z][A-Za-z0-9_]*$')
-# TODO replace relaxed patterns used for compatibility
-# VALID_MESSAGE_NAME_PATTERN = re.compile('^[A-Z][A-Za-z0-9]*$')
-VALID_MESSAGE_NAME_PATTERN = re.compile('^[A-Za-z][A-Za-z0-9]*$')
-VALID_CONSTANT_NAME_PATTERN = re.compile('^[A-Z][A-Z0-9_]*$')
+VALID_PACKAGE_NAME_PATTERN = re.compile('^[a-z]([a-z0-9_]?[a-z0-9]+)*$')
+VALID_FIELD_NAME_PATTERN = re.compile('^[a-z]([a-z0-9_]?[a-z0-9]+)*$')
+# relaxed patterns used for compatibility with ROS 1 messages
+# VALID_FIELD_NAME_PATTERN = re.compile('^[A-Za-z][A-Za-z0-9_]*$')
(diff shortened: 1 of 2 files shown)


wjwwood:
+1

tfoote:
+1

esteve:
+1

wjwwood:
So I've finally run into the messages which lead us to have the "less" strict regex. For example, sensor_msgs/CameraInfo:

http://docs.ros.org/api/sensor_msgs/html/msg/CameraInfo.html

It has fields called `D`, `K`, `R`, and `P`. These are "symbols" that represent elements of the image rectification algorithm. You could call them scientific or mathematical terms. So, the question is what to do about these? I guess we could update CameraInfo's field names, and I'd say that the plain english versions of them might even be more readable and informative, but I'm not sure about enforcing this across all messages. It seems like that might be too strict in some cases.

If we decide to keep this enforcement then I'll need help figuring out how to change CameraInfo correctly. At the very least I think changing CameraInfo in this way will require iteration with the community and I'm sure it will make sense to them why we are doing this. I have to admit that "To match the style" might not be justification enough, even for me.

dirk-thomas:
For the fields in the `CameraInfo` message I would rename them to lower case. I would assume that everybody familiar with the symbols will directly understand there meaning.

wjwwood:
That might be the best solution in this case. However, +0 to changing `CameraInfo` like this. I think we need to discuss it with the community.

tfoote:
From discussion in meeting we'll go ahead with this and see how painful the conversions are with this policy in place.

wjwwood:
For me to do, update messages that violate this stricter policy (`CameraInfo`).

gerkey:
Feedback is that the benefit from consistent style is outweighed by the cost of having to update existing message definitions and code using them.  So we'll provide a linter that you can opt out of using.

Open question: how should we resolve existing violations of this style in widely used messages (of which CameraInfo is one)?  @esteve to reply to the email thread about this.

gerkey:
For now, we'll keep the strict version of the generators, and we'll update ROS2 versions of CameraInfo and DisparityImage to follow the new style, just lower-casing them.

In the future, we can consider adding a feature to the generators to make them less strict, if there's enough demand for it.

Also, in the future we can consider proposals for changing those (and other) field names to something more descriptive.

@wjwwood to update the messages.

Then, this PR is ready to merge.

Then, @dirk-thomas to update the bridge to handle this new mapping.

wjwwood:
Thread, starting with @esteve's message: https://groups.google.com/d/msg/ros-sig-ng-ros/YAyfgrvUvs0/ULthxTF51GwJ

esteve:
Given that several members of the ROS community have expressed their concerns about this proposal, I think we should close this PR and work on the linter instead.

wjwwood:
I've opened the pr to change the offending messages: https://github.com/ros2/common_interfaces/pull/6

wjwwood:
I will change the branch name in my pr and run CI to double check before merging.

wjwwood:
I create a duplicate branch for ros2/common_interfaces#6 to match this one's branch name and queued a CI for it: http://ci.ros2.org/job/ros2_batch_ci_linux/67/

wjwwood:
I rebased this branch and started a new CI job: http://ci.ros2.org/job/ros2_batch_ci_linux/68/

wjwwood:
The job passed, but with two failing tests. The tests are just style related though:

http://ci.ros2.org/job/ros2_batch_ci_linux/68/testReport/

After that is fixed up, I think this is ready to go.

dirk-thomas:
I updated the style in dc91c2fc59db605f669953421f2a5ecef7a8373c
Should be ready to go.

Issue #26, opened 2015-07-09
[rosidl_generator_c] fails on static array of non-primitive type


dirk-thomas:

ernestmc:
I would like to start working on this issue. How can I test the code to see this failure?

jacquelinekay:
@wjwwood will contact you about starting this issue

ernestmc:
I've added back the line referenced in #27 and when building everything I get the following error:
```
Generating C++ code for ROS interfaces
Traceback (most recent call last):
  File "/home/ecorbellini/repos/osrf/ros2_ws/install/lib/rosidl_generator_c/rosidl_generator_c", line 25, in <module>
    sys.exit(main())
  File "/home/ecorbellini/repos/osrf/ros2_ws/install/lib/rosidl_generator_c/rosidl_generator_c", line 20, in main
    args.generator_arguments_file,
  File "/home/ecorbellini/repos/osrf/ros2_ws/install/lib/python3.4/site-packages/rosidl_generator_c/__init__.py", line 53, in generate_c
    minimum_timestamp=latest_target_timestamp)
  File "/home/ecorbellini/repos/osrf/ros2_ws/install/lib/python3.4/site-packages/rosidl_cmake/__init__.py", line 89, in expand_template
    interpreter.file(h)
  File "/usr/lib/python3/dist-packages/em.py", line 2326, in file
    self.safe(scanner, done, locals)
  File "/usr/lib/python3/dist-packages/em.py", line 2368, in safe
    self.parse(scanner, locals)
  File "/usr/lib/python3/dist-packages/em.py", line 2388, in parse
    token.run(self, locals)
  File "/usr/lib/python3/dist-packages/em.py", line 1403, in run
    interpreter.execute(self.code, locals)
  File "/usr/lib/python3/dist-packages/em.py", line 2565, in execute
    exec(statements, self.globals)
  File "<string>", line 7, in <module>
KeyError: 'PrimitivesBounded'
Exception ignored in: <bound method Interpreter.__del__ of <empy pseudomodule/interpreter at 0x2b4086ab6b70>>
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/em.py", line 2063, in __del__
  File "/usr/lib/python3/dist-packages/em.py", line 2126, in shutdown
  File "/usr/lib/python3/dist-packages/em.py", line 2619, in finalize
  File "/usr/lib/python3/dist-packages/em.py", line 2169, in push
[...]

PR #27, opened 2015-07-09
exclude arrays of non-primitive types for now

dirk-thomas:
Temporary workaround for #26.

The patch does not generate any code for static arrays of non-primitive types anymore because the template does not work in that case.

### DIFF
--- a/rosidl_generator_c/resource/msg__struct.h.template
+++ b/rosidl_generator_c/resource/msg__struct.h.template
@@ -42,8 +42,10 @@ for field in spec.fields:
 @{
 for field in spec.fields:
     if field.type.is_array and field.type.array_size is not None:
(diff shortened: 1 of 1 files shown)


wjwwood:
+1, were you using the C generated code for something?

dirk-thomas:
No, but the generator fails due to a Python error in the template (`KeyError`).

wjwwood:
I see, there was not context for the change (no description of the error). Sounds good to me.

jacquelinekay:
+1

tfoote:
The TODO would be clearer if you used the negative logic and a `continue` instead of reindenting the print. Or if it will need a different code path, put the TODO in a placeholder else block. 

+!

PR #28, opened 2015-07-13
refactor generators, fix timestamp of generated files

dirk-thomas:
Refactoring to pass all generator arguments through a JSON / YAML file. This will avoid problems with the length of commands on Windows.

Unification of expanding templates by moving the implementation to `rosidl_cmake`. This will address ros2/ros2#44 once the same patch has been applied to all generators.

Fix problem of reinvoking generator targets because the generated files where skipped (since the content did not change) but their timestamp was older then other target dependencies which led to rebuilds even in the install step.

### DIFF
--- a/rosidl_cmake/cmake/rosidl_write_generator_arguments.cmake
+++ b/rosidl_cmake/cmake/rosidl_write_generator_arguments.cmake
@@ -1,0 +1,117 @@
+# Copyright 2015 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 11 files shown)


tfoote:
+1, you might separate the commits next time the parallel refactoring makes it harder to read.

wjwwood:
+1

dirk-thomas:
The CI finally passed with all the related PRs:
* http://ci.ros2.org/job/ros2_batch_ci_linux/31/
* http://ci.ros2.org/job/ros2_batch_ci_osx/16/

It significantly speeds up the local builds compared to when the generators are always rerun - even during the install step. Therefore I will try to merge them soon.

The Windows CI builds do not fail during the generator steps anymore (with a random character missing) since the command lines are now much shorter. Instead we are having linker errors now: http://ci.ros2.org/job/ros2_batch_ci_windows/11/

jacquelinekay:
+1

PR #29, opened 2015-07-15
Add "has_bounded_size()" to generated message code

jacquelinekay:
Adds helper functions to `rosidl_parser` to check if a message specification has a bounded size and incorporates them into the message template.

It looked to me like `rosidl_parser` does not recursively store the type information of nested types. So currently nested types call `has_bounded_size` recursively in the generated C++ code on their non-primitive sub-fields.

### DIFF
--- a/rosidl_generator_cpp/resource/msg__struct.hpp.template
+++ b/rosidl_generator_cpp/resource/msg__struct.hpp.template
@@ -15,6 +15,7 @@
 @{
 from rosidl_generator_cpp import escape_string
 from rosidl_generator_cpp import msg_type_to_cpp
+from rosidl_generator_cpp import cpp_type_of_array 
 from rosidl_generator_cpp import MSG_TYPE_TO_CPP
 from rosidl_generator_cpp import value_to_cpp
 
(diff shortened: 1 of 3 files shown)


dirk-thomas:
I would propose to make this property available at compile time so that code can choose at compile time (using `std::enable_if`) how to handle a message.

jacquelinekay:
alright, in d86e2e2
<https://github.com/ros2/rosidl/commit/d86e2e2b1ed7a9d45e0f46079fafd3a6189fe0b2>
I changed rosidl_parser to pass around a list of interface dependency files
and construct MessageSpecification objects for each non-primitive type that
is encountered. Then in MessageSpecification.has_bounded_size() I
recursively call the function for non-primitive nested types. That way the
property is known at compile time for the generated message code. The
implementation is a bit inefficient since there are potentially many
redundant MessageSpecification objects in rosidl_parser, but I think it's
ok for now.

Style fixes and migration of the test into rosidl_generator_cpp forthcoming.

dirk-thomas:
The generator should *not* need to parse all recursive dependencies to generate the code for a message. It simply slows down the generation process significantly.

It should also not be necessary since the information can be composed from the information of all submessages and the information of all fields in the message.

jacquelinekay:
I don't see how to determine the bounded size property for a message
without the changes I've made in this PR.

On master, rosidl_parser parses the message file and creates the following
objects:

MessageSpecification, which has an array of Fields. Each Field has a Type.
Type contains the metadata necessary to determine if the message type is
bounded or not (is it a bounded array or string).

However, non-primitive Type objects (representing submessages of the
MessageSpecification) do not appear to contain information about their
fields, which could be bounded or unbounded. It appears to me that the
parser needs to parse the message specification for the submessage to know
this.

The only alternative I can think of to this approach is to store message
metadata as it's encountered by the parser and pull metadata from the cache
to determine if a submessage of a message is bounded.

If there's something I'm missing or a strategy that I've overlooked please
let me know.

dirk-thomas:
The `bounded_size` is an attribute of a message. It either has it (with a specific value) or doesn't have it (if it is unbounded).

For every message that attribute can be expressed by aggregating the information from all fields. If any field does not have a bounded size the whole message has no bounded size. If all fields do have a bounded size the whole message has a bounded size equal to the sum of the bounded sizes of all fields.

It should be possible to express this using `std::enable_if`. So if the message is unbounded the templates don't generate a `bounded_size` symbol. Otherwise the message has that symbol and use the symbols provided by all submessage types to build the sum (at compile time).

By making the information available at *compile time* it allows to write code which selects a more efficient code path at compile time rather then querying the bounded size at runtime and handling it with `if` / `else`.

dirk-thomas:
Wrong button pressed, sorry.

jacquelinekay:
I think we've had a lot of miscommunication which is leading me down a false path, so let's clarify some things:

I agree with everything you said about `bounded_size`, but this pull request doesn't even calculate or declare the `bounded_size` symbol, only the `has_bounded_size` symbol.

I was planning on introducing the `bool has_bounded_size` symbol or function in this PR and then later introducing a `bounded_size` symbol that gets computed from all sub-fields.

You seem to be saying that there should be no `has_bounded_size` symbol. Instead there should be a `bounded_size` representing the bounded size of a message that does not get generated for unbounded messages.

My reply to your initial post, however, was about how to determine IF the message has a bounded size, in the Python code or the generated C++ code. It's necessary to look at all recursive sub-messages of a message to determine if it's bounded or unbounded. My [initial approach](https://github.com/ros2/rosidl/blob/75848ac6bfde86011943e7476878f6875f99e0e9/rosidl_generator_cpp/resource/msg__struct.hpp.template#L153) made recursive calls of `has_bounded_size` in the C++ generated msg code. I changed it to recursively store metadata in the Python parser in d86e2e2 based on my misinterpretation of what you said.

I will try another approach after I learn some more about `std::enable_if` and metaprogramming.

PR #30, opened 2015-07-17
Determine `has_bounded_size` at compile time for messages and services

jacquelinekay:
This pull request enables the generated message code to determine if a message is bounded or not at compile time. It adds a test to `rosidl_generator_cpp` that will not compile on failure, and tweaks the CMakeLists to allow for the project to register itself as an ament extension before it is installed, in order to build the messages for the test.

Future work:
Dynamically sized, bounded arrays and strings must be implemented. Bounded dynamic arrays and strings can be specified in IDL but these traits are not enforced by the message's representation in memory. This could be accomplished by providing a custom allocator with a maximum size.

The `bounded_size` symbol will be added to compute the maximum size of a message. Currently `sizeof` is accurate for bounded messages, since any message with strings or vectors is considered unbounded. Once dynamically sized and bounded data structures are enabled, `sizeof` will be inaccurate and `bounded_size` will be necessary to expose this information.


rosidl_generator_cpp/CMakeLists.txt
```diff
@@ -11,12 +11,56 @@ ament_export_include_directories(include)
 ament_python_install_package(${PROJECT_NAME})
 
 if(AMENT_ENABLE_TESTING)
-  find_package(ament_lint_auto REQUIRED)
-  ament_lint_auto_find_test_dependencies()
```
dirk-thomas: This should be preserved.

rosidl_generator_cpp/CMakeLists.txt
```diff
@@ -11,12 +11,56 @@ ament_export_include_directories(include)
...
+    "msg/StaticArrayNestedBounded.msg"
+    "msg/StaticArrayPrimitives.msg"
+    "msg/StaticArrayBounded.msg"
+  )
+
+  set(service_files
+    "srv/Empty.srv"
+    "srv/PrimitivesBounded.srv"
+    "srv/DynamicArray.srv"
+  )
```
dirk-thomas: I would consider removing the services since there i nothing different between them (requests / responses) and messages.

rosidl_generator_cpp/CMakeLists.txt
```diff
@@ -11,12 +11,56 @@ ament_export_include_directories(include)
...
+
+  include(cmake/register_cpp.cmake)
```
dirk-thomas: Is this file missing in the PR?

jacquelinekay: yes, sorry about that, added in 483884e

rosidl_generator_cpp/resource/msg__meta.hpp.template
```diff
@@ -0,0 +1,77 @@
+// generated from rosidl_generator_cpp/resource/msg__meta.hpp.template
```
dirk-thomas: Should we call this file `*_traits.hpp` since it contains a type trait?

rosidl_generator_cpp/resource/msg__meta.hpp.template
```diff
@@ -0,0 +1,77 @@
...
+#endif  /* __ALL_MSG__BASIC_TEMPLATES */
+
+#ifndef __@(spec.base_type.pkg_name)__@(subfolder)__@(get_header_filename_from_msg_name(spec.base_type.type))__TEMPLATES
+#define __@(spec.base_type.pkg_name)__@(subfolder)__@(get_header_filename_from_msg_name(spec.base_type.type))__TEMPLATES
+
+#include "@(spec.base_type.pkg_name)/@(subfolder)/@(get_header_filename_from_msg_name(spec.base_type.type))__struct.hpp"
+
+@{
+template_string=""
+if len(spec.fields) == 0:
```
dirk-thomas: Better: `if not spec.fields:`

rosidl_generator_cpp/resource/msg__meta.hpp.template
```diff
@@ -0,0 +1,77 @@
...
+@[end for]@
+
+  // TODO(jacquelinekay): for now, all dynamically sized arrays and strings are considered unbounded
+  template<>
+  struct has_bounded_size<std::string> { static const bool value = false; };
+
+  template<typename T>
+  struct has_bounded_size<std::vector<T>> { static const bool value = false; };
+
+#endif  /* __ALL_MSG__BASIC_TEMPLATES */
```
dirk-thomas: Should this block be moved to a separate header file rather then embedded in every generated message?

rosidl_generator_cpp/resource/msg__struct.hpp.template
```diff
@@ -22,6 +22,7 @@ cpp_namespace = '%s::%s::' % (spec.base_type.pkg_name, subfolder)
 cpp_class = '%s_' % spec.base_type.type
 cpp_full_name = '%s%s' % (cpp_namespace, cpp_class)
 cpp_full_name_with_alloc = '%s<ContainerAllocator>' % (cpp_full_name)
+
```
dirk-thomas: Please avoid white space only changes.

rosidl_generator_cpp/resource/msg.hpp.template
```diff
@@ -15,6 +15,7 @@
 #ifndef __@(spec.base_type.pkg_name)__@(subfolder)__@(get_header_filename_from_msg_name(spec.base_type.type))__hpp__
 #define __@(spec.base_type.pkg_name)__@(subfolder)__@(get_header_filename_from_msg_name(spec.base_type.type))__hpp__
 
+#include "@(spec.base_type.pkg_name)/@(subfolder)/@(get_header_filename_from_msg_name(spec.base_type.type))__meta.hpp"
```
dirk-thomas: Should we move this line one down (same as in the service one) to indicate that it uses the struct?

rosidl_generator_cpp/CMakeLists.txt
dirk-thomas: I was referring to `cmake/register_cpp.cmake`.

dirk-thomas:
That looks good.

I think we should add support for the fields with a known upper boundary too. Seems like a low hanging fruit since the rosidl API already provides all the information.

May be it makes sense to provide two different traits: `bounded_size` and `fixed_size`?

rosidl_generator_cpp/CMakeLists.txt
jacquelinekay: OK, I will remove the other `find_packages` too since `ament_lint_auto_find_test_dependencies()` will find them for me, correct?

rosidl_generator_cpp/CMakeLists.txt
jacquelinekay: yeah, that too... 21cf726

rosidl_generator_cpp/resource/msg__meta.hpp.template
jacquelinekay: yes, that is a much better name for it

rosidl_generator_cpp/package.xml
```diff
@@ -17,5 +17,7 @@
   <exec_depend>rosidl_parser</exec_depend>
 
   <test_depend>ament_lint_auto</test_depend>
-  <test_depend>ament_lint_common</test_depend>
```
dirk-thomas: This still needs to be preserved.

rosidl_generator_cpp/resource/msg__meta.hpp.template
jacquelinekay: I'm all for this, but it creates a package dependency for all generated messages. Do you have a suggestion for which package should provide definitions for all C++ messages? Is `rosidl_generator_cpp` the appropriate place?

rosidl_generator_cpp/resource/msg.hpp.template
dirk-thomas: Another question: should the default header include the trait or should that be included explicitly by the user?

wjwwood: I'd say that is should, since in the use case where the classes in rclcpp, like subscription, are using the traits, the user would then need to know when that's going to happen and preemptively include the traits for the right message as the rclcpp classes cannot include the file when it is needed. So including it in the default message header seems like the best strategy in my opinion.

rosidl_generator_cpp/resource/srv.hpp.template
```diff
@@ -12,6 +12,7 @@
 #ifndef __@(spec.pkg_name)__srv__@(get_header_filename_from_msg_name(spec.srv_name))__hpp__
 #define __@(spec.pkg_name)__srv__@(get_header_filename_from_msg_name(spec.srv_name))__hpp__
 
+#include "@(spec.pkg_name)/srv/@(get_header_filename_from_msg_name(spec.srv_name))__traits.hpp"
```
dirk-thomas: The service struct only references the request and reponse message types. Why do they need a trait?

jacquelinekay: The `has_bounded_size` trait for services is just a convenience for checking that both the request and response messages for the service are bounded.

rosidl_generator_cpp/resource/msg.hpp.template
jacquelinekay: I'm actually working on an example of how this trait will be used in `rclcpp` right now:

`MessageMemoryStrategy<MessageT>` is injected into `Subscription<MessageT>`
`BoundedMsgMemoryStrategy<MessageT>` inherits `MsgMemoryStrategy`
`BoundedMsgMemoryStrategy<MessageT>` has some `enable_if` logic that prevents compilation if the user tries to use the template specialization where `MessageT` is unbounded.

Like William said, the user has to remember to include the `traits` header for the message when using `BoundedMessageMemoryStrategy`.

this might be OK since message type traits are "advanced" user topics and not needed in most use cases.

But I'm inclined to agree with William, since the connection between the `traits` header and the object where it is will be used is not obvious and the potential source of many headaches.

jacquelinekay:
I like the idea. What I've currently implemented is "has_fixed_size", not
bounded size, according to the IDL spec.

It currently looks like the implementation of bounded, dynamically sized
fields is incomplete, however. For example, the upper bound never gets
stored anywhere. I should implement `fixed_size` and `bounded_size`, but we
should also get around to exposing more of the rosidl API to the message
structure.

dirk-thomas:
I don't understand which "missing" information you are referring to.

The `rosidl_parser` API does provide the complete information of a message which is specified in the `.msg` file. And the introspection typesupport should make almost all of this available in code (including array sizes, upper bounds etc.: https://github.com/ros2/rosidl_dds/blob/81f873b13561eca5f8b5a72848446830658dcb87/rosidl_typesupport_introspection_cpp/include/rosidl_typesupport_introspection_cpp/message_introspection.hpp#L30-L45

jacquelinekay:
In 03fa56b I differentiate between "has_fixed_size" and `has_bounded_size`. In b6dbaac I implement `bounded_size`, which uses `sizeof` and metadata parsed from the `rosidl_parser` API to calculate the maximum size of bounded, dynamically sized messages.

Still in progress: use "enable_if" to make the `bounded_size` symbol available only if `has_bounded_size` is true.

dirk-thomas:
I don't understand how `sizeof` should be used for that. The size of a message with upper bounds for all fields can't be determined with it since we do use dynamic data structures like vectors.

jacquelinekay:
For a message `T` with a string field bounded by `n`, `bounded_size` adds `n * sizeof(char)` to the size of the message, since this will not be captured by `sizeof(T)`.

For a message `T` with a dynamic array of type `S` bounded by `n`, `bounded_size` adds `n * sizeof(S)` to the size of the message `T`.

oh, I just realized while writing this that the second case does not capture the case of a dynamic array of a message type that contains another dynamically sized field.

wjwwood:
If the "S" is not primitive then you need to call bounded size of that too.

dirk-thomas:
I might not see the big picture where this is heading to. Obviously you can't allocate `N` bytes and expect the message to be a POD and fit in it. What is the goal this "theoretical" number of `N` bytes should be used for?

wjwwood:
@dirk-thomas I pointed this out before and I believe it was you that suggested that it should be used to know how much memory to allocate for the static pool of memory, from which a C++ allocator could draw memory.

wjwwood:
I find it confusion that the `PrimitivesBounded.msg` message doesn't contain any "bounding" synatx, i.e. `[<=N]` or `string<=N`. And `Primitives.msg` does have a bounded string in it. Are we mixing the concept of bounded but dynamic and fixed sized (static)?

jacquelinekay:
@wjwwood Your confusion is understandable. The messages were originally lifted from `test_communication` then roughly adapted to what I wanted to test. I decided to recreate the suite of tested messages in fa6f0c5 with clearer names and more test case coverage. I also fixed a small bug revealed by expanding the test cases.

jacquelinekay:
in b522bcc I added the compile-time disabling of `bounded_size` for messages with unbounded sizes. Calling `msg_type::bounded_size::value` where `msg_type` is an unbounded message will cause a compilation error. The test has a commented out block that can be uncommented to see this behavior.

wjwwood:
@jacquelinekay to split bounded and fixed components of the pr up into multiple pr's.

jacquelinekay:
I'm going to put this on the backlog until we resolve the question of how to manage memory for dynamically sized, bounded data structures.

mikaelarguedas:
the messages traits have been implemented but `has_bounded_size` still doesn't exist. only `has_fixed_size` has been implemented so far. I'll remove all the unnecessary code from here and put it back in backlog with only the necessary changes

mikaelarguedas:
closed in favor of #229

PR #33, opened 2015-07-21
Implement has_fixed_size and test

jacquelinekay:
This pull request adds the ability to check if a message has fixed size at compile time. It adds a test for `has_fixed_size`, enables `rosidl_generator_cpp` to register itself as a generator before installation to be used in the test, and adds several messages to be used in the test. It also demonstrates in the test that `sizeof` is sufficient to calculate the fixed size of a messages.

### DIFF
--- a/rosidl_generator_cpp/CMakeLists.txt
+++ b/rosidl_generator_cpp/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 2.8.3)
 
-project(rosidl_generator_cpp NONE)
+project(rosidl_generator_cpp)
 
 find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_python REQUIRED)
@@ -13,10 +13,54 @@ ament_python_install_package(${PROJECT_NAME})
 if(AMENT_ENABLE_TESTING)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
+
+  set(message_files
+    "msg/BoundedArrayBounded.msg"
+    "msg/BoundedArrayStatic.msg"
+    "msg/BoundedArrayUnbounded.msg"
+
+    "msg/Empty.msg"
+
+    "msg/PrimitiveStaticArrays.msg"
+
+    "msg/PrimitivesBounded.msg"
+    "msg/PrimitivesStatic.msg"
+    "msg/PrimitivesUnbounded.msg"
+
+    "msg/StaticArrayBounded.msg"
+    "msg/StaticArrayStatic.msg"
+    "msg/StaticArrayUnbounded.msg"
+
+    "msg/UnboundedArrayBounded.msg"
+    "msg/UnboundedArrayStatic.msg"
+    "msg/UnboundedArrayUnbounded.msg"
+  )
+
+  include(cmake/register_cpp.cmake)
+  set(rosidl_generator_cpp_DIR "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+
+  rosidl_generator_cpp_extras(
+    "${CMAKE_CURRENT_SOURCE_DIR}/bin/rosidl_generator_cpp"
+    "${CMAKE_CURRENT_SOURCE_DIR}/rosidl_generator_cpp/__init__.py"
+    "${CMAKE_CURRENT_SOURCE_DIR}/resource"
+  )
+
+  rosidl_generate_interfaces(${PROJECT_NAME} ${message_files}
+    SKIP_INSTALL
+  )
+
+  if(NOT WIN32)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
+  endif()
+
+  add_executable(test_interfaces_cpp "test/test_interfaces.cpp")
+  add_dependencies(test_interfaces_cpp ${PROJECT_NAME} ${rosidl_generate_interfaces_TARGET})
+  # include the built files directly, instead of their install location
+  include_directories("${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}")
 endif()
 
 ament_package(
-  CONFIG_EXTRAS "rosidl_generator_cpp-extras.cmake.in"
+  CONFIG_EXTRAS "cmake/register_cpp.cmake" "rosidl_generator_cpp-extras.cmake.in"
 )
 
 install(
--- a/rosidl_generator_cpp/cmake/rosidl_generator_cpp_generate_interfaces.cmake
+++ b/rosidl_generator_cpp/cmake/rosidl_generator_cpp_generate_interfaces.cmake
@@ -26,11 +26,13 @@ foreach(_idl_file ${rosidl_generate_interfaces_IDL_FILES})
     list(APPEND _generated_msg_files
       "${_output_path}/${_parent_folder}/${_header_name}.hpp"
       "${_output_path}/${_parent_folder}/${_header_name}__struct.hpp"
+      "${_output_path}/${_parent_folder}/${_header_name}__traits.hpp"
     )
   elseif("${_parent_folder} " STREQUAL "srv ")
     list(APPEND _generated_srv_files
       "${_output_path}/${_parent_folder}/${_header_name}.hpp"
       "${_output_path}/${_parent_folder}/${_header_name}__struct.hpp"
+      "${_output_path}/${_parent_folder}/${_header_name}__traits.hpp"
     )
   else()
     message(FATAL_ERROR "Interface file with unknown parent folder: ${_idl_file}")
@@ -53,8 +55,10 @@ set(target_dependencies
   ${rosidl_generator_cpp_GENERATOR_FILES}
   "${rosidl_generator_cpp_TEMPLATE_DIR}/msg.hpp.template"
   "${rosidl_generator_cpp_TEMPLATE_DIR}/msg__struct.hpp.template"
+  "${rosidl_generator_cpp_TEMPLATE_DIR}/msg__traits.hpp.template"
   "${rosidl_generator_cpp_TEMPLATE_DIR}/srv.hpp.template"
   "${rosidl_generator_cpp_TEMPLATE_DIR}/srv__struct.hpp.template"
+  "${rosidl_generator_cpp_TEMPLATE_DIR}/srv__traits.hpp.template"
   ${rosidl_generate_interfaces_IDL_FILES}
   ${_dependency_files})
 foreach(dep ${target_dependencies})
@@ -83,11 +87,16 @@ add_custom_command(
   VERBATIM
 )
 
-add_custom_target(
+if (TARGET ${rosidl_generate_interfaces_TARGET}__cpp)
+  message(WARNING "Custom target ${rosidl_generate_interfaces_TARGET}__cpp already exists")
+else()
+  add_custom_target(
   ${rosidl_generate_interfaces_TARGET}__cpp
   DEPENDS
   ${_generated_msg_files} ${_generated_srv_files}
 )
+endif()
+
 add_dependencies(
   ${rosidl_generate_interfaces_TARGET}
   ${rosidl_generate_interfaces_TARGET}__cpp
--- a/rosidl_generator_cpp/package.xml
+++ b/rosidl_generator_cpp/package.xml
@@ -16,6 +16,9 @@
 
   <exec_depend>rosidl_parser</exec_depend>
 
+  <test_depend>ament_cmake_nose</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
+  <test_depend>rosidl_cmake</test_depend>
+  <test_depend>rosidl_generator_c</test_depend>
 </package>
--- a/rosidl_generator_cpp/resource/msg__traits.hpp.template
+++ b/rosidl_generator_cpp/resource/msg__traits.hpp.template
@@ -1,0 +1,69 @@
+// generated from rosidl_generator_cpp/resource/msg__traits.hpp.template
+
+@#######################################################################
+@# EmPy template for generating <msg>__traits.hpp files
+@#
+@# Context:
+@#  - spec (rosidl_parser.MessageSpecification)
+@#    Parsed specification of the .msg file
+@#  - subfolder (string)
+@#    The subfolder / subnamespace of the message
+@#    Either 'msg' or 'srv'
+@#  - get_header_filename_from_msg_name (function)
+@#######################################################################
+@
+@{
+from rosidl_generator_cpp import MSG_TYPE_TO_CPP
+
+cpp_namespace = '%s::%s::' % (spec.base_type.pkg_name, subfolder)
+}@
+#include <array>
+#include <stdint.h>
+#include <string>
+#include <vector>
+
+namespace rosidl_generator_traits
+{
+
+#ifndef __ROSIDL_GENERATOR_CPP_TRAITS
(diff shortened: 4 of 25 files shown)


rosidl_generator_cpp/CMakeLists.txt
```diff
@@ -13,10 +13,54 @@ ament_python_install_package(${PROJECT_NAME})
...
+  )
+
+  if(NOT WIN32)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall")
+  endif()
+
+  add_executable(test_interfaces_cpp "test/test_interfaces.cpp")
+  add_dependencies(test_interfaces_cpp ${PROJECT_NAME} ${rosidl_generate_interfaces_TARGET})
+  # include the built files directly, instead of their install location
+  include_directories("${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}" "~/code/ros2_ws/build/geometry_interfaces/rosidl_generator_cpp")
```
dirk-thomas: Using a local path won't work.

rosidl_generator_cpp/cmake/rosidl_generator_cpp_generate_interfaces.cmake
```diff
@@ -25,12 +25,14 @@ foreach(_idl_file ${rosidl_generate_interfaces_IDL_FILES})
   if("${_parent_folder} " STREQUAL "msg ")
     list(APPEND _generated_msg_files
       "${_output_path}/${_parent_folder}/${_header_name}.hpp"
+      "${_output_path}/${_parent_folder}/${_header_name}__traits.hpp"
```
dirk-thomas: Move `traits` one line lower after `struct`.

rosidl_generator_cpp/cmake/rosidl_generator_cpp_generate_interfaces.cmake
```diff
@@ -74,20 +78,30 @@ rosidl_write_generator_arguments(
   TARGET_DEPENDENCIES ${target_dependencies}
 )
 
+get_filename_component(generator_module_path ${rosidl_generator_cpp_GENERATOR_FILES} DIRECTORY)
```
dirk-thomas: `rosidl_generator_cpp_GENERATOR_FILES` can contain multiple files and should therefore not be interpreted as a single path.

rosidl_generator_cpp/package.xml
```diff
@@ -18,4 +18,7 @@
 
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
+  <test_depend>ament_cmake_nose</test_depend>
```
dirk-thomas: Please keep in alphabetical order.

rosidl_generator_cpp/resource/msg__traits.hpp.template
```diff
@@ -0,0 +1,79 @@
...
+            fixed = False
+            break
+        elif field.type.is_primitive_type():
+            fixed_template_string += "has_fixed_size<{}>::value && ".format(MSG_TYPE_TO_CPP[field.type.type])
+        else:
+            fixed_template_string += "has_fixed_size<{}::msg::{}>::value && ".format(field.type.pkg_name, field.type.type)
+    if not fixed:
+        fixed_template_string = "false"
+    else:
+        fixed_template_string = fixed_template_string[:len(fixed_template_string)-4]
```
wjwwood: Can I recommend instead collecting the fixed template strings in a list and then joining them in this case? Something like this:

```python
fixed_template_strings = []
fixed_template_strings.append("...".format(...))
...
fixed_template_string = ' && '.join(fixed_template_strings) if fixed_template_strings else 'true'
```

wjwwood:
+1 other than the comment.

rosidl_generator_cpp/CMakeLists.txt
```diff
@@ -13,10 +13,54 @@ ament_python_install_package(${PROJECT_NAME})
...
+    "${CMAKE_CURRENT_SOURCE_DIR}/rosidl_generator_cpp/__init__.py"
+    "${CMAKE_CURRENT_SOURCE_DIR}/resource"
+  )
+
+  rosidl_generate_interfaces(${PROJECT_NAME} ${message_files}
+    SKIP_INSTALL
+  )
+
+  if(NOT WIN32)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall")
```
dirk-thomas: Please also use `-Wextra`.

rosidl_generator_cpp/resource/msg__traits.hpp.template
```diff
@@ -0,0 +1,79 @@
...
+@
+@{
+from rosidl_generator_cpp import MSG_TYPE_TO_CPP
+
+cpp_namespace = '%s::%s::' % (spec.base_type.pkg_name, subfolder)
+
+cpp_fixed_types = ['bool', 'char', 'uint8_t', 'int8_t', 'uint16_t', 'int16_t', 'uint32_t', 'int32_t', 'uint64_t', 'int64_t', 'float', 'double']
+}@
+#include <stdint.h>
+#include <array>
```
dirk-thomas: Alphabetical order.

rosidl_generator_cpp/resource/msg__traits.hpp.template
```diff
@@ -0,0 +1,79 @@
...
+#include <string>
+#include <vector>
+
+#ifndef __ALL_MSG__BASIC_TEMPLATES
```
dirk-thomas: This should use a name more related to this generator. E.g. `__ROSIDL_GENERATOR_CPP_TRAIT_TEMPLATES`

rosidl_generator_cpp/resource/msg__traits.hpp.template
```diff
@@ -0,0 +1,79 @@
...
+        fixed_template_string = fixed_template_string[:len(fixed_template_string)-4]
+}@
+
+template<>
+struct has_fixed_size<@(cpp_namespace)@(spec.base_type.type)>
+{
+  static const bool value = @(fixed_template_string);
+};
+
+//////////////////////////////////////
```
dirk-thomas: Please remove this.

rosidl_generator_cpp/resource/msg__traits.hpp.template
```diff
@@ -0,0 +1,79 @@
...
+template<>
+struct has_fixed_size<@(fixed_type)>
+{
+  static const bool value = true;
+};
+@[end for]@
+
+#endif  /* __ALL_MSG__BASIC_TEMPLATES */
+
+#ifndef __@(spec.base_type.pkg_name)__@(subfolder)__@(get_header_filename_from_msg_name(spec.base_type.type))__TEMPLATES
```
dirk-thomas: Please use the same suffix as the filename: `traits` instead of `templates`

rosidl_generator_cpp/resource/msg__traits.hpp.template
```diff
@@ -0,0 +1,79 @@
...
+#include <stdint.h>
+#include <array>
+#include <string>
+#include <vector>
+
+#ifndef __ALL_MSG__BASIC_TEMPLATES
+#define __ALL_MSG__BASIC_TEMPLATES
+
+template<typename T>
+struct has_fixed_size {};
```
dirk-thomas: Should this (and all specializations) be defined in a namespace?

dirk-thomas:
Instead of having the "merge" commit can you rebase your change to master then it shouldn't need that separate commit.

rosidl_generator_cpp/resource/msg__traits.hpp.template
jacquelinekay: Yes, namespacing is a good idea. `rosidl_generator_cpp_traits`? Or maybe just `rosidl_cpp_traits`?

jacquelinekay:
All comments should now be addressed.

Expanding on William's suggestion, I simplified `has_fixed_size` so that the Python code is cleaner and there are no more specializations shared across all `__traits.hpp` files, just the base definition.

rosidl_generator_cpp/cmake/rosidl_generator_cpp_generate_interfaces.cmake
```diff
@@ -74,20 +78,34 @@ rosidl_write_generator_arguments(
   TARGET_DEPENDENCIES ${target_dependencies}
 )
 
+set(generator_module_paths "")
```
dirk-thomas: The variable is not used anyway.

rosidl_generator_cpp/resource/msg__traits.hpp.template
```diff
@@ -0,0 +1,70 @@
...
+
+#ifndef __@(spec.base_type.pkg_name)__@(subfolder)__@(get_header_filename_from_msg_name(spec.base_type.type))__traits__hpp__
+#define __@(spec.base_type.pkg_name)__@(subfolder)__@(get_header_filename_from_msg_name(spec.base_type.type))__traits__hpp__
+
+#include "@(spec.base_type.pkg_name)/@(subfolder)/@(get_header_filename_from_msg_name(spec.base_type.type))__struct.hpp"
+
+@{
+fixed_template_strings = []
+fixed = True
+fixed_template_string = ''
```
dirk-thomas: This definition is not necessary since the variable gets set later in all cases.

rosidl_generator_cpp/resource/msg__traits.hpp.template
dirk-thomas: This comment is still pending.

rosidl_generator_cpp/resource/msg__traits.hpp.template
dirk-thomas: This comment is still pending.

rosidl_generator_cpp/resource/srv__traits.hpp.template
```diff
@@ -0,0 +1,35 @@
...
+@#  - get_header_filename_from_msg_name (function)
+@#######################################################################
+@
+@{
+cpp_namespace = '%s::srv::' % (spec.pkg_name)
+}@
+
+#include "@(spec.pkg_name)/srv/@(get_header_filename_from_msg_name(spec.srv_name))__struct.hpp"
+
+#ifndef __ALL_MSG__BASIC_TEMPLATES
```
dirk-thomas: This should follow the naming in the msg equivalent.

rosidl_generator_cpp/resource/srv__traits.hpp.template
```diff
@@ -0,0 +1,35 @@
...
+#define __ALL_MSG__BASIC_TEMPLATES
+
+  template<typename T>
+  struct has_fixed_size { };
+
+#endif  /* __ALL_MSG__BASIC_TEMPLATES */
+
+#ifndef __@(spec.pkg_name)__srv__@(get_header_filename_from_msg_name(spec.srv_name))__TEMPLATES
```
dirk-thomas: This should follow the naming in the msg equivalent.

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
```diff
@@ -28,11 +28,13 @@ def generate_cpp(generator_arguments_file):
     template_dir = args['template_dir']
     mapping_msgs = {
         os.path.join(template_dir, 'msg.hpp.template'): '%s.hpp',
+        os.path.join(template_dir, 'msg__traits.hpp.template'): '%s__traits.hpp',
```
dirk-thomas: traits after struct. Same below.

rosidl_generator_cpp/resource/srv__traits.hpp.template
```diff
@@ -0,0 +1,35 @@
...
+@{
+cpp_namespace = '%s::srv::' % (spec.pkg_name)
+}@
+
+#include "@(spec.pkg_name)/srv/@(get_header_filename_from_msg_name(spec.srv_name))__struct.hpp"
+
+#ifndef __ALL_MSG__BASIC_TEMPLATES
+#define __ALL_MSG__BASIC_TEMPLATES
+
+  template<typename T>
```
dirk-thomas: Also namespaced.

dirk-thomas:
The build currently has compiler warnings.

dirk-thomas:
It is currently also installing all generated message files which I am not sure yet why... I am looking into that.

rosidl_generator_cpp/cmake/rosidl_generator_cpp_generate_interfaces.cmake
jacquelinekay: This is fixed as of 94828cf, where it is used instead of `generator_module_path` in the `PYTHONPATH`.

jacquelinekay:
fc3947a fixes compiler warnings in the test introduced by this PR

rosidl_generator_cpp/cmake/rosidl_generator_cpp_generate_interfaces.cmake
```diff
@@ -74,20 +78,34 @@ rosidl_write_generator_arguments(
...
+foreach (rosidl_generator_cpp_GENERATOR_FILE ${rosidl_generator_cpp_GENERATOR_FILES})
+  get_filename_component(generator_module_path ${rosidl_generator_cpp_GENERATOR_FILE} DIRECTORY)
+  get_filename_component(generator_module_path ${generator_module_path} DIRECTORY)
+  set(generator_module_paths "${generator_module_paths}:${generator_module_path}")
+endforeach()
+
+set(TMP_PYTHONPATH "$ENV{PYTHONPATH}${generator_module_paths}")
```
dirk-thomas: I would suggest to not fiddle with the PYTHONPATH especially since these paths are not correct when `generator files` contains multiple files from different locations.

Instead you can update the actually invoked Python script to work from source. Instead of `from rosidl_generator_cpp import generate_cpp` you can do the following:

    try:
        from rosidl_generator_cpp import generate_cpp
    except ImportError:
        rosidl_generator_cpp_root = os.path.dirname(os.path.dirname(__file__)))
        rosidl_generator_cpp_package = os.path.join(rosidl_generator_cpp_root, 'rosidl_generator_cpp')
        if not os.path.exists(rosidl_generator_cpp_package):
            raise
        sys.path.insert(0, os.path.abspath(rosidl_generator_cpp_root))
        from rosidl_generator_cpp import generate_cpp

dirk-thomas:
+1 (after squashing)

wjwwood:
+1 for squash and merge

dirk-thomas:
This fails to run (https://github.com/ros2/rosidl/pull/33/files#diff-6b733852654ce984020c970df67c2a22R10 since `os` was not imported.

I have fixed the error as well as a few tests due to style and copyright problem in: c2ef15c78b6b88a279afd24357ace4a487a29e6a

Please run the tests locally and CI job on Jenkins for future PRs.

dirk-thomas:
This also broke the windows CI build (or lets say: broke it early since it was failing before): http://ci.ros2.org/job/ros2_batch_ci_windows/16/consoleFull#console-section-48

Windows always prepends the current folder to sys.path and therefore can't import a package with the same name as the script. I used a different mechanism to load the Python package from the script in: 11739a1753b0d17985a63184ed9ff510a3d58802

wjwwood:
That's why we use `entry_points` on Windows. It would create a `.exe` for the script, avoiding the import issue.

dirk-thomas:
Making the script an entry point would be a good idea in general - currently we can't to that since from within CMake we can't use these setup.py features.

Anyway it wouldn't help in this case since the script is being invoked in the source space for the tests.

PR #34, opened 2015-07-22
fix SKIP_INSTALL to affect generated files

dirk-thomas:
Discovered in #33 that it was not working for generated files but only for the `.msg` / `.srv` files before.


jacquelinekay:
+1, thanks for fixing this!

wjwwood:
+1

PR #35, opened 2015-07-22
use recursive dependencies to determine interface dependencies

dirk-thomas:
Without this message package hierarchies like `A -> B -> C` fail to build in isolation because A will only find interfaces from B but not from C.


wjwwood:
+1

PR #37, opened 2015-08-11
support ROS 1 pointer type but deprecate it

dirk-thomas:
The renames `Ptr` to `RawPtr` to not break expectations.

It keeps `Ptr` as a `std::shared_ptr` (as it is in ROS 1) but deprecates its usage.

Connects to ros2/design#53

### DIFF
--- a/rosidl_generator_cpp/resource/msg__struct.hpp.template
+++ b/rosidl_generator_cpp/resource/msg__struct.hpp.template
@@ -39,7 +39,15 @@ for field in spec.fields:
         print(
             '#include <%s/msg/%s.hpp>' %
             (field.type.pkg_name, get_header_filename_from_msg_name(field.type.type)))
+deprecated_macro_name = \
+    '_'.join(['DEPRECATED', spec.base_type.pkg_name, subfolder, spec.base_type.type])
 }@
+
+#ifndef _WIN32
+# define @(deprecated_macro_name) __attribute__((deprecated))
+#else
+# define @(deprecated_macro_name) __declspec(deprecated)
+#endif
 
 namespace @(spec.base_type.pkg_name)
 {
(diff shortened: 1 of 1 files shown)


rosidl_generator_cpp/resource/msg__struct.hpp.template
```diff
@@ -144,6 +152,14 @@ for field in spec.fields:
   typedef std::weak_ptr<@(cpp_full_name)<ContainerAllocator> const>
     ConstWeakPtr;
 
+  // pointer types similar to ROS 1, use SharedPtr / ConstSharedPtr instead
+  typedef @(deprecated_macro_name)
+    std::shared_ptr<@(cpp_full_name)<ContainerAllocator>>
+    Ptr;
+  typedef @(deprecated_macro_name)
+    std::shared_ptr<@(cpp_full_name)<ContainerAllocator> const>
+    ConstPtr;
```
wjwwood: Looking at this:

https://gcc.gnu.org/onlinedocs/gcc-3.3/gcc/Type-Attributes.html

It looks like the deprecation attribute should go at the end of the typedef. Maybe this doesn't matter and has to be this way on Windows? Have you tried it?

wjwwood:
Other than the comment +1

rosidl_generator_cpp/resource/msg__struct.hpp.template
dirk-thomas: Putting it at the end doesn't work on Windows (http://ci.ros2.org/job/ros2_batch_ci_windows/146/console). On Linux / OS X it seems to work in both places.

Should I add it to the platform specific location (making the template and code a bit more complex) or just keep it where it is (http://ci.ros2.org/job/ros2_batch_ci_windows/147/console)?

wjwwood: I think leaving it where it is will be fine. I'd prefer the other location if not for Windows.

PR #38, opened 2015-08-13
add check that message package declares the interface files

dirk-thomas:
Instead of failing while building the specific type support package this will fail early if e.g. a ROS 1 message package is found instead.


esteve:
+1

tfoote:
+1

wjwwood:
+1

PR #41, opened 2015-10-08
add alias for unique pointer with deleter

jacquelinekay:
In order to properly delete unique pointers with custom allocators, we need to expose unique pointers templated on a Deleter. Open to feedback about the naming.

### DIFF
--- a/rosidl_generator_cpp/resource/msg__struct.hpp.template
+++ b/rosidl_generator_cpp/resource/msg__struct.hpp.template
@@ -142,10 +142,16 @@ for field in spec.fields:
   using ConstSharedPtr =
     std::shared_ptr<@(cpp_full_name)<ContainerAllocator> const>;
 
-  using UniquePtr =
-    std::unique_ptr<@(cpp_full_name)<ContainerAllocator>>;
-  using ConstUniquePtr =
-    std::unique_ptr<@(cpp_full_name)<ContainerAllocator> const>;
+  template<typename Deleter = std::default_delete<@(cpp_full_name)<ContainerAllocator>>>
+  using UniquePtrWithDeleter =
+    std::unique_ptr<@(cpp_full_name)<ContainerAllocator>, Deleter>;
+
+  using UniquePtr = UniquePtrWithDeleter<>;
+
+  template<typename Deleter = std::default_delete<@(cpp_full_name)<ContainerAllocator>>>
+  using ConstUniquePtrWithDeleter =
+    std::unique_ptr<@(cpp_full_name)<ContainerAllocator> const, Deleter>;
+  using ConstUniquePtr = ConstUniquePtrWithDeleter<>;
 
   using WeakPtr =
     std::weak_ptr<@(cpp_full_name)<ContainerAllocator>>;


dirk-thomas:
Instead of creating a second type with a second name could we just make the existing one a template but with the deleter having an optional value like `std::default_delete<T>`?

jacquelinekay:
I prefer the suggested approach, and implemented it as follows:

```
  template<typename Deleter = std::default_delete<@(cpp_full_name)<ContainerAllocator>>>
  using UniquePtr =
    std::unique_ptr<@(cpp_full_name)<ContainerAllocator>, Deleter>;

  template<typename Deleter = std::default_delete<@(cpp_full_name)<ContainerAllocator>>>
  using ConstUniquePtr =
    std::unique_ptr<@(cpp_full_name)<ContainerAllocator> const, Deleter>;
```

But that causes compile errors where UniquePtr is being used without a template:

```
/home/jackie/code/ros2_ws/src/ros2/demos/intra_process_demo/src/cyclic_pipeline/cyclic_pipeline.cpp:32:36: error: 
      use of class template 'std_msgs::msg::Int32::UniquePtr' requires template arguments
      [this](std_msgs::msg::Int32::UniquePtr & msg) {
                                   ^
/home/jackie/code/ros2_ws/install/include/std_msgs/msg/int32__struct.hpp:64:3: note: template is declared here
  using UniquePtr =
...
```

Even though the default deleter argument gets resolved, it expects the alias to be used with the bracket notation (`UniquePtr<>`).

dirk-thomas:
Right, since the template has then zero arguments it would not be great for using it. For the `std::unique_ptr` type that works nicely since it is either one or two arguments. Well, forget about the idea :wink: 

But one comment regarding the code snippet you posted: it uses `std_msgs::msg::Int32::UniquePtr & msg` as an argument. The reference must be removed to achieve the desired passing of ownership.

esteve:
An alternate approach is to have the template accept a deleter with `std::default_delete` as default, like @dirk-thomas suggested, but still add a `UniquePtrWithDefaultDeleter` aliased to `UniquePtr<>`:

```c++
using UniquePtrWithDefaultDeleter = UniquePtr<>;
```

but in the end it won't add much, so I think the original approach is much better and more expressive.

wjwwood:
I'd suggest the reverse so that the default case is simpler, e.g.:

```c++
#include <memory>

template<typename ContainerAllocator = std::allocator<void>>
struct A_ {
  template<typename Deleter = std::default_delete<A_<ContainerAllocator>>>
  using UniquePtrWithDeleter = std::unique_ptr<A_<ContainerAllocator>, Deleter>;
  using UniquePtr = UniquePtrWithDeleter<>;
};

typedef A_<> A;

int main(void) {
  A::UniquePtr a;
  return 0;
}
```

You could use `UniquePtrWithCustomDeleter` instead if that's more specific.

jacquelinekay:
I made the suggested changes and started CI:

http://ci.ros2.org/job/ros2_batch_ci_linux/420
http://ci.ros2.org/job/ros2_batch_ci_osx/331/
http://ci.ros2.org/job/ros2_batch_ci_windows/466/

dirk-thomas:
lgtm

esteve:
+1

PR #48, opened 2015-11-12
Fixed the include for [arrays.h] in C.

### LINKED ISSUE #47: Incorrect C string header included
https://github.com/ros2/rosidl/blob/master/rosidl_generator_c/include/rosidl_generator_c/arrays.h#L22
The include should be `string.h` instead of `string`.

ernestmc:
Fixes #47. Very simple fix: just added the C header instead of the C++ one. Otherwise compiling for c99 couldn't find the library.


wjwwood:
+1

PR #52, opened 2015-11-18
update generated C API

dirk-thomas:
This patch generates C messages with all currently supported types (especially bounded arrays as well as strings). Additionally it provides functions to operate on the array and string structs.

Improvements to be done later (both also missing from C++):
* option to preallocate bounded arrays
* expose default values

http://ci.ros2.org/job/ros2_batch_ci_linux/614/
http://ci.ros2.org/job/ros2_batch_ci_osx/506/
http://ci.ros2.org/job/ros2_batch_ci_windows/682/

Windows Connext doesn't build because of https://github.com/ros2/ros2/pull/154#issuecomment-157845377


dirk-thomas:
Based on the meeting I will work on the discussed changes for a while.

dirk-thomas:
After implementing all the changes (and getting it to compile on all three platforms) this is back in review.

I updated the description with the latest builds.

rosidl_generator_c/rosidl_generator_c/__init__.py
```diff
@@ -91,14 +102,63 @@ def msg_type_to_c(type_, name_, containing_msg_name='notset'):
...
+    if type_ == 'string':
+        return '"%s"' % escape_string(value)
+
+    assert False, "unknown primitive type '%s'" % type_
+
+
+def escape_string(s):
+    s = s.replace('\\', '\\\\')
+    s = s.replace('"', '\\"')
+    return s
```
wjwwood: I understand why the `"` needs to be escaped, but I don't understand why it needs to be double escaped as `\\"` rather than `\"`. Can you explain why you need that? I think it would be good to add this scenario to the test cases to make sure it's working as expected.

rosidl_generator_c/src/primitives_array_functions.c
```diff
@@ -0,0 +1,68 @@
...
+      data = malloc(sizeof(TYPE_NAME) * size); \
+      if (!data) { \
+        return false; \
+      } \
+    } \
+    array->data = data; \
+    array->size = size; \
+    array->capacity = size; \
+    return true; \
+  } \
```
wjwwood: Should you check to see if array is `NULL` in this function?

wjwwood: This applies to some of the other functions like the matching `_fini` function.

rosidl_generator_c/src/string_functions.c
```diff
@@ -0,0 +1,149 @@
...
+    return false;
+  }
+  // since n + 1 bytes are being allocated n can't be the maximum value
+  if (n == SIZE_MAX) {
+    return false;
+  }
+  char * data = malloc(n + 1);
+  if (!data) {
+    return false;
+  }
```
wjwwood: This function can return false for several reasons, but it isn't possible to get more information about why. Should this use `RMW_SET_ERROR_MSG` (or a clone/alias of it called `ROSIDL_SET_ERROR_MSG`) in order to convey additional information?

rosidl_generator_c/rosidl_generator_c/__init__.py
dirk-thomas: The first escaping happens in Python. The double backslash becomes a single backslash in the output and the generated code then contains `\"` in order to escape the quote within quotes.

wjwwood: Oh I see, yeah makes sense.

rosidl_generator_c/src/primitives_array_functions.c
dirk-thomas: Absolutely. I added it for all cases I found in 0694ccbeedec3eb5af33d483d1420a8e38b9ea8d

rosidl_generator_c/src/string_functions.c
dirk-thomas: I would rather no add complex error handling at this level since it would complicate the generated C message code significantly. Basically it comes down to two reasons:
* wrong function arguments
* unable to mallow
If the caller ensures that it passes valid arguments the only reason is memory allocation.

dirk-thomas:
Waiting for a +1 on this as well as the referenced PR.

wjwwood:
lgtm

PR #54, opened 2015-11-25
C services

dirk-thomas:
Generate C code for the request and response of services.

Currently no C code is being generated for the service itself. In C++ we do this: https://github.com/ros2/rosidl/blob/61db30d36b893415382a4e6ecca6fe05b0e16dce/rosidl_generator_cpp/resource/srv__struct.hpp.template#L24-L28 but for C I don't have an idea for something similar.

http://ci.ros2.org/job/ci_linux/638/


wjwwood:
The linked CI job was aborted, was there a new one? The code changes look good to me.

dirk-thomas:
The CI job was aborted after the build phase since it passed. It didn't make sense to run the tests since the generated code of this PR is not used anywhere yet.

PR #55, opened 2015-12-01
C typesupport

dirk-thomas:
Currently trying to make Windows happy...

### DIFF
--- a/rosidl_typesupport_introspection_c/include/rosidl_typesupport_introspection_c/message_introspection.h
+++ b/rosidl_typesupport_introspection_c/include/rosidl_typesupport_introspection_c/message_introspection.h
@@ -1,0 +1,51 @@
+// Copyright 2014-2015 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ROSIDL_TYPESUPPORT_INTROSPECTION_C__MESSAGE_INTROSPECTION_H_
+#define ROSIDL_TYPESUPPORT_INTROSPECTION_C__MESSAGE_INTROSPECTION_H_
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include "rosidl_generator_c/message_type_support.h"
+
+#include "rosidl_typesupport_introspection_c/visibility_control.h"
+
+typedef struct rosidl_typesupport_introspection_c__MessageMember
+{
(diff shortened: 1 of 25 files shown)


tfoote:
@wjwwood says that you're not exporting symbols; that's why you don't get a .LIB with your .DLL.

dirk-thomas:
The problem that it needs to be:
* in a single compilation unit in order to be able to compare the pointer instead of the string value
* visible in the the library in order to use the symbol in other code parts
* a compile-time constant in order to use it as in a static initializer list

I have not yet found a way to achieve all of this at the same time on Windows.

gerkey:
As we discovered earlier today, in C you just can't use static const variables in static initializers.  E.g., this code doesn't compile with gcc on Linux:

~~~
static const int a = 0;
static const int b = a;
~~~

...but it does compile with g++.

dirk-thomas:
This will be continued once ros2/rmw_implementation#9 has been completed.

dirk-thomas:
I have rebased and updated this - ready for review:
* http://ci.ros2.org/job/ci_linux/734/
* http://ci.ros2.org/job/ci_osx/610/
* http://ci.ros2.org/job/ci_windows_opensplice/803/
* http://ci.ros2.org/job/ci_windows_connext_static/64/
* http://ci.ros2.org/job/ci_windows_connext_dynamic/52/

Locally I got it to compile and link the following example: https://github.com/ros2/examples/compare/master...c_msgs but since it is a weird mixture of C and C++ it won't run.

I think this should be merged even if we don't have a working talker / listener example in C with rcl(c) yet.

rosidl_typesupport_introspection_c/include/rosidl_typesupport_introspection_c/message_introspection.h
```diff
@@ -0,0 +1,55 @@
...
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include "rosidl_generator_c/message_type_support.h"
+
+#include "rosidl_typesupport_introspection_c/visibility_control.h"
+
+typedef struct
+// ROSIDL_TYPESUPPORT_INTROSPECTION_C_PUBLIC
```
wjwwood: Should this be dropped?

rosidl_typesupport_introspection_c/include/rosidl_typesupport_introspection_c/message_introspection.h
```diff
@@ -0,0 +1,55 @@
...
+  uint32_t offset_;
+  const void * default_value_;
+  size_t (* size_function)(const void *);
+  const void * (*get_const_function)(const void *, size_t index);
+  void * (*get_function)(void *, size_t index);
+  bool (* resize_function)(void *, size_t size);
+} rosidl_typesupport_introspection_c__MessageMember;
+
+typedef struct
+// ROSIDL_TYPESUPPORT_INTROSPECTION_C_PUBLIC
```
wjwwood: Same here?

rosidl_typesupport_introspection_c/include/rosidl_typesupport_introspection_c/message_introspection.h
dirk-thomas: Yeah, I removed them for now. We can put them in when we need them down the road.

wjwwood:
lgtm

PR #57, opened 2015-12-18
add rosidl_default_runtime package

dirk-thomas:
Based on our discussion on the whiteboard this is the right way to fix ros2/rclcpp#181.

Connects to ros2/rclcpp#181

Without this the test failed: http://ci.ros2.org/job/ci_osx/656/
With this the test doesn't fail due to the not found library: http://ci.ros2.org/job/ci_osx/657/


tfoote:
+1 this passes in my local workspace on our osx test machine.

PR #58, opened 2015-12-18
Temporarily add extra dependencies to workaround linking issue on OS X

wjwwood:
We need a more complete solution, but for now we'll have to workaround linking issues in tests on OS X with these changes.

See: ros2/ros2#174


dirk-thomas:
If you happen to not have dependencies which are specified in the manifest there are simply being ignored.

gerkey:
+1

dirk-thomas:
Opened #59 to track removing this hack again.

PR #63, opened 2015-12-31
Preliminary Python generator

esteve:


esteve:
I can't get this (https://github.com/ros2/rosidl/blob/python-generator/rosidl_generator_py/CMakeLists.txt#L49) to work, so any help will be appreciated.

rosidl_generator_py/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+    "${CMAKE_CURRENT_SOURCE_DIR}/bin/rosidl_generator_py"
+    "${CMAKE_CURRENT_SOURCE_DIR}/rosidl_generator_py/__init__.py"
+    "${CMAKE_CURRENT_SOURCE_DIR}/resource"
+  )
+
+  rosidl_generate_interfaces(${PROJECT_NAME} ${message_files}
+    SKIP_INSTALL
+  )
+
+  ament_add_nose_test(test_interfaces_py "${CMAKE_CURRENT_BINARY_DIR}/test")
```
jacquelinekay: Doesn't the second argument here need to be a fully qualified path to a python script? e.g. `"${CMAKE_CURRENT_BINARY_DIR}/test_interfaces.py"`

esteve: I tried that too, but it won't work either. The reason is that I can't get the `test` directory to exist during the build phase (when `${CMAKE_CURRENT_BINARY_DIR}` points to `WORKSPACE/build_isolated/rosidl_generator_py`). I could use `configure_file()` to generate a file there, but I don't know if that's the correct approach.

jacquelinekay: Looking at our existing usages of `ament_add_nose_test`, it looks like `configure_file` is used to copy the file from the `test` folder directly into `CMAKE_CURRENT_BINARY_DIR`:

https://github.com/ros2/demos/blob/master/pendulum_control/CMakeLists.txt#L95
https://github.com/ros2/system_tests/blob/master/test_rclcpp/CMakeLists.txt#L90
https://github.com/ros2/examples/blob/master/rclcpp_examples/CMakeLists.txt#L103

I don't know what the *correct* method is, but this seems to be the most widely used one.

dirk-thomas: The macro has a docblock describing all arguments (https://github.com/ament/ament_cmake/blob/master/ament_cmake_nose/cmake/ament_add_nose_test.cmake#L18-L39). The second argument in this case can either be a file or a folder (either absolute or relative to `CMAKE_CURRENT_SOURCE_DIR` as any CMake command).

The questions are:
* where is the file `${CMAKE_CURRENT_BINARY_DIR}/test` supposed to come from?
* is `test/test_interfaces.py` not supposed to contains tests?

esteve: > where is the file `${CMAKE_CURRENT_BINARY_DIR}/test` supposed to come from?

From here https://github.com/ros2/rosidl/tree/python-generator/rosidl_generator_py/test

> is `test/test_interfaces.py` not supposed to contain the tests (without any templating)?

Yes, but since I couldn't get it to be available during the build phase, I moved on to other things and get back to adding tests once I resolved the aforementioned issue.

dirk-thomas: `${CMAKE_CURRENT_BINARY_DIR}` refers to the CMake build directory. But neither a template is being expanded using `configure_file` nor any file is copied to that location. If you want to use the file in the source path of the package you just need to specify the relative path to it ("test").

esteve: So, how can I copy the test directory to the build directory? Is calling `configure_file()` without expanding anything actually ok?

dirk-thomas: Why would you want to do that when the files are not changes at all? You can simply use the file from where it already is, no?

esteve: The test depends on files generated by `rosidl_generator_py` from the msg, and I'd rather not muck with `sys.path` to be able to import them from the tests.

dirk-thomas: I would highly recommend to set the environment accordingly rather then copying the test files.

esteve: How? Adding the build directory to `sys.path` in the test?

dirk-thomas: Setting the `PYTHONPATH` for the test. See ament/ament_cmake#52.

esteve: I've added `PYTHONPATH` to `APPEND_ENV`, however the empty test file still can't import the package if added via `ament_add_nose_test`. I load `test_interfaces.py` directly from the command line with the same `PYTHONPATH`, Python can find the the messages just fine. Any ideas?

dirk-thomas:
The error message of the failing test is:

    ImportError: No module named 'rosidl_generator_py.msg'

The important part is that it does **not** say:

    ImportError: No module named 'rosidl_generator_py'

So Python **does** find the Python package `rosidl_generator_py` but not the module `msg`.

Based on that I ran the test from a shell with the PYTHONPATH set but from two different locations:
* it fails when the cwd is the package source folder (same as what the CTest uses)
* it works when the cwd is something different (e.g. /tmp)

So the problem is obviously that Python finds the subfolder `rosidl_generator_py` in the package sources when trying to import `rosidl_generator_py.msg`. Since it is not possible (without ugly hacks) to mix Python code of a single package from two different locations and the test doesn't really need the Python code from the source space we need to make it only find the Python code from the build space.

To achieve that it needs to change the cwd when running nose. Since this was not yet possible I have implemented the option (ament/ament_cmake#54). With this you can pass a different path as the working directory to the nose test, e.g.:

    ament_add_nose_test(... WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/cmake")

rosidl_generator_py/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+    "${CMAKE_CURRENT_SOURCE_DIR}/rosidl_generator_py/__init__.py"
+    "${CMAKE_CURRENT_SOURCE_DIR}/resource"
+  )
+
+  rosidl_generate_interfaces(${PROJECT_NAME} ${message_files}
+    SKIP_INSTALL
+  )
+
+  ament_add_nose_test(test_interfaces_py "test/test_interfaces.py"
+    APPEND_ENV PYTHONPATH=${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py
```
dirk-thomas: You should quote the argument since it might contain spaces.

esteve:
@dirk-thomas that's awesome, thank you so much! I'll give it a try.

rosidl_generator_py/CMakeLists.txt
esteve: Thanks.

esteve:
A few things before I tag this for review that I'd appreciate some feedback for:

-  There's a bunch of cleanups pending, I've moved things around from the template to the code, back and forth.
- I ended up adding a few types (`Float64`, `Int8`, etc.) in `rosidl_generator_py_primitive_types` because of the limited types that Python support (no unsigned, no float size, etc.) and also so I could test this, because depending on `std_msgs` would create a circular dependency.
- There's no support for constraints (i.e. upperbound, number of elements, etc.). I don't know if this should happen when setting the `data` field in a type or delegate this to `rclc` when the types are used in `publish`, for example.

dirk-thomas:
* The Python message class should be self contained. It should store each field in a native Python type. The mapping from ROSIDL types to Python types needs to be carefully decided. E.g. `byte[]` should be represented as `builtins.bytes`.
  * I don't think that we should define custom types for these: rosidl_generator_py_primitive_types/__init__.py This only makes our data objects "specific to custom ROS stuff".
* Each member should ensure that all boundaries are satisfied according to the message definition. E.g. trying to assign a value `foo` to a field of type `string<=2` should result in an error.

esteve:
> * The Python message class should be self contained. It should store each field in a native Python type. > The mapping from ROSIDL types to Python types needs to be carefully decided. E.g. `byte[]` should be > represented as `builtins.bytes`.

I agree, but how should `int8` and `uint8` be mapped? Since the only representation for that we have now is in `std_msgs`, how can they be tested in `rosidl_generator_py` without depending on the former?

> * Each member should ensure that all boundaries are satisfied according to the message definition. E.g. > trying to assign a value `foo` to a field of type `string<=2` should result in an error.

Sounds good. I only thought that delegating this to `rclc` would just remove some redundancy, at the expense of only reporting such errors later in the game. In any case, I'll leave this for a followup PR so I can get this landed in master as soon as possible.

tfoote:
We should put together a article on type mappings covering the python datatypes like: generated c++ Interfaces

dirk-thomas:
> > * The Python message class should be self contained. It should store each field in a native Python type. > The mapping from ROSIDL types to Python types needs to be carefully decided. E.g. `byte[]` should be > represented as `builtins.bytes`.

> I agree, but how should `int8` and `uint8` be mapped? Since the only representation for that we have now is in `std_msgs`, how can they be tested in `rosidl_generator_py` without depending on the former?

`int8` and `uint8` have nothing to do with `std_msgs`. These names are only the identifier we are using in the `.msg` files. The message files in `std_msgs` are only for convenience to send messages of builtin types without requiring to create a message definition first.

> > * Each member should ensure that all boundaries are satisfied according to the message definition. E.g. > trying to assign a value `foo` to a field of type `string<=2` should result in an error.

> Sounds good. I only thought that delegating this to `rclc` would just remove some redundancy, at the expense of only reporting such errors later in the game. In any case, I'll leave this for a followup PR so I can get this landed in master as soon as possible.

The goal is that the Python data structures are usable independently usable which I would argue also includes the exact type and boundaries.

I will look into creating the article. I am sure I have the results from our previous discussions somewhere.

dirk-thomas:
Please see for the proposed mapping between ROSIDL and Python types ros2/design#76

esteve:
This is now ready for review, I'll add constraints in a followup branch.

rosidl_generator_py/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+    SKIP_INSTALL
+  )
+
+  ament_add_nose_test(test_interfaces_py "test/test_interfaces.py"
+    APPEND_ENV PYTHONPATH=${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py:${CMAKE_CURRENT_SOURCE_DIR}
+    WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py"
+  )
+  if(TARGET test_interfaces_py)
+    add_dependencies(test_interfaces_py ${PROJECT_NAME})
+  endif()
```
dirk-thomas: `test_interfaces_py` is not a target but a test. Therefore the condition is always false and using `add_dependencies` is not possible.

rosidl_generator_py/bin/rosidl_generator_py
```diff
@@ -0,0 +1,40 @@
...
+    from importlib.machinery import SourceFileLoader
+
+    loader = SourceFileLoader('rosidl_generator_py', rosidl_generator_py_module)
+    rosidl_generator_py = loader.load_module()
+    generate_py = rosidl_generator_py.generate_py
+
+
+def main(argv=sys.argv[1:]):
+    parser = argparse.ArgumentParser(
+        description='Generate the C++ ROS interfaces.',
```
dirk-thomas: `C++` - copy-n-paste.

rosidl_generator_py/bin/rosidl_generator_py
```diff
@@ -0,0 +1,40 @@
...
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument(
+        '--generator-arguments-file',
+        required=True,
+        help='The location of the file containing the generator arguments')
+    args = parser.parse_args(argv)
+
+    return generate_py(
+        args.generator_arguments_file,
+    )
```
dirk-thomas: This could be on a single line.

rosidl_generator_py/bin/rosidl_generator_py
```diff
@@ -0,0 +1,40 @@
...
+
+    loader = SourceFileLoader('rosidl_generator_py', rosidl_generator_py_module)
+    rosidl_generator_py = loader.load_module()
+    generate_py = rosidl_generator_py.generate_py
+
+
+def main(argv=sys.argv[1:]):
+    parser = argparse.ArgumentParser(
+        description='Generate the C++ ROS interfaces.',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
```
dirk-thomas: The `formatter_class` is useless since no argument has default values.

esteve: Good catch, I'll submit a PR to fix this in the C and C++ generators.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -0,0 +1,108 @@
...
+# limitations under the License.
+
+set(_output_path
+  "${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py/${PROJECT_NAME}")
+set(_generated_msg_files "")
+set(_generated_srv_files "")
+foreach(_idl_file ${rosidl_generate_interfaces_IDL_FILES})
+  get_filename_component(_parent_folder "${_idl_file}" DIRECTORY)
+  get_filename_component(_parent_folder "${_parent_folder}" NAME)
+  get_filename_component(_msg_name "${_idl_file}" NAME_WE)
```
dirk-thomas: This seems to use camel case for the filename. According to PEP 8 and our article we want to use underscore lowercase.

rosidl_generator_py/bin/rosidl_generator_py
esteve: Fixed.

rosidl_generator_py/bin/rosidl_generator_py
esteve: Fixed.

rosidl_generator_py/CMakeLists.txt
esteve: Fixed.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -0,0 +1,108 @@
...
+    normalize_path(_abs_idl_file "${_abs_idl_file}")
+    list(APPEND _dependency_files "${_abs_idl_file}")
+    list(APPEND _dependencies "${_pkg_name}:${_abs_idl_file}")
+  endforeach()
+endforeach()
+
+set(target_dependencies
+  "${rosidl_generator_py_BIN}"
+  ${rosidl_generator_py_GENERATOR_FILES}
+  "${rosidl_generator_py_TEMPLATE_DIR}/msg.py.template"
```
dirk-thomas: Should we call this template `_msg.py.template` in order to match the pattern it is being expanded to?

rosidl_generator_py/resource/msg.py.template
```diff
@@ -0,0 +1,27 @@
+# Generated by rosidl_generator_py
```
dirk-thomas: Please use similar comments at the top of the templates as in the other generators (exact template, copyright notice).

rosidl_generator_py/resource/msg.py.template
```diff
@@ -0,0 +1,27 @@
...
+@[end for]@
+@[else]@
+    __slots__ = []
+@[end if]@
+
+    def __eq__(self, other):
+        return (type(self) == type(other)
+@[for field in spec.fields]@
+            and self.@(field.name) == other.@(field.name)
+@[end for]@
```
dirk-thomas: Why is this necessary? How does it affect comparing subclasses?

dirk-thomas:
It currently lacks to generate the `__init__.py` files in the parent modules. And it will require logic to decide when the package-level module needs a generated `__init__.py` file or if the user code already provides one.

rosidl_generator_py/resource/msg.py.template
```diff
@@ -0,0 +1,27 @@
...
+class @(spec.base_type.type):
+@[if len(spec.fields) > 0]@
+    __slots__ = ['@("', '".join([field.name for field in spec.fields]))']
+    def __init__(self, **kwargs):
+@[for field in spec.fields]@
+@[if field.default_value]@
+        self.@(field.name) = kwargs.get(
+            '@(field.name)', @value_to_py(field.type, field.default_value))
+@[else]@
+        self.@(field.name) = kwargs.get('@(field.name)')
```
dirk-thomas: Doesn't this fail for all fields without a default value if the argument is not passed to the constructor?

esteve: `dict.get` will return `None` if it can't the key.

rosidl_generator_py/resource/msg.py.template
esteve: I ended up not using `isinstance` because I think it makes more sense to check types explicitly, and we shouldn't encourage users to inherit from generated classes.

rosidl_generator_py/resource/msg.py.template
```diff
@@ -0,0 +1,27 @@
+# Generated by rosidl_generator_py
+
+class @(spec.base_type.type):
+@[if len(spec.fields) > 0]@
+    __slots__ = ['@("', '".join([field.name for field in spec.fields]))']
```
dirk-thomas: Since a message might contain many fields should we generate the code like this:

```
__slots__ = [
    'field_1',
    'field_2',
]
```

This would also avoid needing special handling for message without fields.

rosidl_generator_py/rosidl_generator_py/__init__.py
```diff
@@ -0,0 +1,146 @@
...
+    open(os.path.join(args['output_dir'], '__init__.py'), 'w').close()
+    for module in modules:
+        with open(os.path.join(args['output_dir'], module, '__init__.py'), 'w') as f:
+            for type_ in modules[module]:
+                f.write('from %s._%s import %s\n' % (import_path, type_, type_))
+
+    return 0
+
+
+def convert_path_to_import(path):
```
dirk-thomas: Why is this necessary? The package knows the import pattern precisely (`<package_name>.msg`) and can just construct this.

rosidl_generator_py/rosidl_generator_py/__init__.py
```diff
@@ -0,0 +1,146 @@
...
+    template_dir = args['template_dir']
+    mapping_msgs = {
+        os.path.join(template_dir, 'msg.py.template'): '_%s.py',
+    }
+
+    mapping_srvs = {
+        os.path.join(template_dir, 'srv.py.template'): '_%s.py',
+    }
+    for template_file in mapping_msgs.keys():
+        assert os.path.exists(template_file)
```
dirk-thomas: Add an error message.

Same below.

rosidl_generator_py/rosidl_generator_py/__init__.py
```diff
@@ -0,0 +1,146 @@
...
+    for template_file in mapping_srvs.keys():
+        assert os.path.exists(template_file)
+
+    functions = {
+        'get_header_filename_from_msg_name': convert_camel_case_to_lower_case_underscore,
```
dirk-thomas: Not used anywhere.

rosidl_generator_py/rosidl_generator_py/__init__.py
```diff
@@ -0,0 +1,146 @@
...
+        'int32', 'uint32',
+        'int64', 'uint64',
+    ]:
+        return str(value)
+
+    if type_.type in ['float32', 'float64']:
+        return '%s' % value
+
+    if type_.type == 'string':
+        return '"%s"' % escape_string(value)
```
dirk-thomas: Can we use single quotes around the string here? And escape the single quotes inside.

rosidl_generator_py/test/test_interfaces.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
```
dirk-thomas: If this file is executable it won't be used by nose. If it is not it doesn't need a shebang line.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
esteve: Done.

rosidl_generator_py/test/test_interfaces.py
esteve: Done.

rosidl_generator_py/rosidl_generator_py/__init__.py
esteve: Done.

rosidl_generator_py/rosidl_generator_py/__init__.py
esteve: This uses the same style that's in the other generators, I find it more readable than escaping quotes inside.

dirk-thomas: C++ only support double quotes around strings but Python supports both types. Using single quotes would not result in more escaping:

```
    return "'%s'" % escape_string(value)

def escape_string(s):
    s = s.replace('\\', '\\\\')
    s = s.replace("'", "\\'")
    return s
```

rosidl_generator_py/resource/msg.py.template
dirk-thomas: Oh yeah, my bad. No idea which function I was thinking about.

rosidl_generator_py/resource/msg.py.template
dirk-thomas: I meant why is implementing a custom `__eq__` necessary?

Also I don't think we should discourage people from subclassing from generated classes. There are use cases where that makes sense. Also if we would discourage that it would need to be mentioned (and argued for) in the article since it is quite a restriction.

dirk-thomas:
The constants are currently not yet generated.

rosidl_generator_py/rosidl_generator_py/__init__.py
esteve: I see, thanks for the snippet. Done.

rosidl_generator_py/rosidl_generator_py/__init__.py
esteve: Done.

rosidl_generator_py/resource/msg.py.template
```diff
@@ -0,0 +1,27 @@
...
+    def __init__(self, **kwargs):
+@[for field in spec.fields]@
+@[if field.default_value]@
+        self.@(field.name) = kwargs.get(
+            '@(field.name)', @value_to_py(field.type, field.default_value))
+@[else]@
+        self.@(field.name) = kwargs.get('@(field.name)')
+@[end if]@
+@[end for]@
```
esteve: Done.

rosidl_generator_py/resource/msg.py.template
esteve: Done.

rosidl_generator_py/rosidl_generator_py/__init__.py
esteve: Done.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
esteve: Done.

rosidl_generator_py/resource/msg.py.template
esteve: Done.

esteve:
I've implemented read-only constants via metaclasses, but I don't know if I should move the metaclass to a common package (though that wouldn't make structures self-contained) or generate message-specific metaclasses (i.e. have a custom `__prepare__` method instead of a generic one):

https://github.com/ros2/rosidl/blob/python-generator/rosidl_generator_py/resource/_msg.py.template#L13

wjwwood:
What's the advantage to using a metaclass? I've not used that extensively before.

Why not just add a method like this to each message class:

```python
class @(name):
    __slots__ = [
        'field1',
        'field2',
        # ...
    ]

    def __setattr__(self, key, value):
        if key in self.__slots__:
            raise AttributeError("Message fields are read-only")
        super().__setattr__(key, value)
```

esteve:
@wjwwood the issue with that approach is that `__slots__` and `__setattr__` act on the instance, but not the class, e.g.

```python
class A:
  __slots__ = ['foo']

a = A()
#a.bar = 123 # fails
A.bar = 456 # works
```

esteve:
I've just squashed the commits.

wjwwood:
I see, in that case I'd say have a custom meta class for each message rather than a common one.

I find the current template a little confusing and I think that will be a detractor for users trying to look at the generated struct. For example a message like this:

```
string FOO="foo"
string BAR="bar"

string foo_or_bar
```

Would generate something like this:

```python
class ReadOnlyConstants(type):
    # ...

class FooOrBar(
    metaclass=ReadOnlyConstants,
    FOO="foo",
    BAR="bar",
):
    __slots__ = [
        'foo_or_bar',
    ]
    # ...
```

I think this might be easier to understand and just as concise so long as you're not going to have a common class for all messages (`ReadOnlyConstants`):

```python
class FooOrBar__Constants(type):
    @property
    def FOO(self):
        """Message contant 'FOO'"""
        return "foo"

    @property
    def BAR(self):
        """Message contant 'FOO'"""
        return self._BAR

class FooOrBar(metaclass=FooOrBar__Constants):
     __slots__ = [
         'foo_or_bar'
    ]
    # ...
```

I didn't test this extensively, but I think it should work and I think it is easier to understand.

wjwwood:
I updated the second example output.

esteve:
@wjwwood yeah, that looks better. Thanks.

wjwwood:
You could also use the `var = property(getter)` signature too, maybe even with a lambda:

```python
FOO = property(lambda: "foo")
```

I didn't try that though.

dirk-thomas:
Please make sure that all members and constants can also provide a custom docstring.

wjwwood:
It would be (the lambda needs to take the class as the first argument):

```python
FOO = property(lambda cls: "foo")
```

wjwwood:
I think you can document properties like this in Sphinx:

```python
FOO = property(lambda cls: "foo")
"""This is the 'FOO' constant, I'm not sure what documentation would go here..."""
```

I'm not sure how the MetaClass affects the visibility of that though.

wjwwood:
I did some experiments with this file:

```python
class FooOrBar__Constants(type):
    FOO = property(lambda cls: "foo", doc="The 'FOO' constant.")

    BAR = property(lambda cls: "bar", doc="The 'BAR' constant.")


class FooOrBar(metaclass=FooOrBar__Constants):
    __slots__ = [
        'FOO',
        'BAR',
        'foo_or_bar',
    ]

    def __init__(self, **kwargs):
        """Message fields can be initialized with keyword arguments."""
        self.foo_or_bar = kwargs.get('foo_or_bar', '')


print(FooOrBar.FOO)  # foo
help(FooOrBar.FOO)  # Does not print help for FOO.
help(FooOrBar__Constants.FOO)  # Does print help for FOO.
try:
    FooOrBar.FOO = 'bar'
    raise RuntimeError("Should not get here")
except AttributeError:
    pass
help(FooOrBar)
```

This is the output on my machine:

```
foo
No Python documentation found for 'foo'.
Use help() to get the interactive help utility.
Use help(str) for help on the str class.

Help on property:

    The 'FOO' constant.

Help on class FooOrBar in module __main__:

class FooOrBar(builtins.object)
 |  Methods defined here:
 |  
 |  __init__(self, **kwargs)
 |      Message fields can be initialized with keyword arguments.
 |  
 |  ----------------------------------------------------------------------
 |  Data descriptors defined here:
 |  
 |  foo_or_bar
 |  
 |  ----------------------------------------------------------------------
 |  Data and other attributes defined here:
 |  
 |  BAR = 'bar'
 |  
 |  FOO = 'foo'

```

That's the best mixture of documentation and readability I could get. Maybe others can improve on it.

esteve:
I made a few changes, it's based on @wjwwood 's suggestion, but with some variations:

- Use the `@property` decorator instead of the function version, I'll add constraints to the message fields, so I preferred to use methods instead for classes as well to maintain make the code more uniform: https://github.com/ros2/rosidl/pull/63/files#diff-d6b183785a53f8b68b433e842e9bc8bbR7

- Didn't add the class variable to `__slots__`, so the value of the class variable shows up when calling `help(instance)`

- Used `__prepare__` to prepopulate the class with the class variables.

wjwwood:
I had to add the class variables to the `__slots__` in order for them to be seen in the help() output. Doing help on either `FooOrBar` or an instance of that class produces the same result for me.

esteve:
@wjwwood you're right, I now remember why I didn't add it. If it's there, but is not assigned anywhere, `instance.CONSTANT` does not work. Not that I'd use it that way (I prefer Class.CONSTANT), but if you define a class variable normally, it's also accessible by instances. However, if I kept the constant in `__slots__`, but also in `__prepare__`, Python complains that both conflict.

wjwwood:
Ok, sounds reasonable to me.

Also why are we using `__slots__`? Is it for the performance? It might be the right thing to do, I'm just wondering why it isn't just properties without `__slots__`.

esteve:
@wjwwood it's both because of performance reasons and also to prevent from adding extra members to the object.

dirk-thomas:
`__slots__` are about saving memory: http://code.activestate.com/lists/python-list/531365/

wjwwood:
So long as it is about memory (performance). We could control the ability to add additional members with `__setattr__`. Also I'm not sure why it is necessary for us to do that.

This pr lgtm. Thanks for iterating with us @esteve.

rosidl_generator_py/resource/_msg.py.template
```diff
@@ -0,0 +1,66 @@
...
+@[end for]@
+
+    def __setattr__(cls, name, value):
+        raise AttributeError("Class variables are read-only")
+
+    @@classmethod
+    def __prepare__(cls, name, bases, **kwargs):
+        return {
+@[for constant in spec.constants]@
+            '@(constant.name)': @constant_value_to_py(constant.type, constant.value),
```
dirk-thomas: Is each constant and its value mentioned twice, once in this line and once in line 5-11?

rosidl_generator_py/resource/_msg.py.template
```diff
@@ -0,0 +1,66 @@
...
+    def @(field.name)(self, value):
+        self.__@(field.name) = value
+@[end for]@
+
+    def __eq__(self, other):
+        return (type(self) == type(other)
+@[for field in spec.fields]@
+            and self.@(field.name) == other.@(field.name)
+@[end for]@
+            )
```
dirk-thomas: What about this pending question: https://github.com/ros2/rosidl/pull/63#discussion_r49643680

dirk-thomas:
`help(Foo.BAR)` doesn't work yet.

rosidl_generator_py/resource/_msg.py.template
esteve: Yes, `__prepare__` (used for populating class variables) can only return a dictionary with names and values, but not with a docstring. So they are defined as properties too so they have a docstring as well.

esteve:
@dirk-thomas it'll never work, AFAIK. You can't obtain the docstring of a class variable with `help()`. The reason why it works with the metaclass is because it's a property.

rosidl_generator_py/resource/msg.py.template
esteve: It's not necessary, but I think it's convenient. However, if the C++ classes don't define a comparator either, I'm happy to remove it to keep both languages consistent.

dirk-thomas: I think it doesn't make any sense when the custom comparator is not any different then the default one.

esteve: I just removed the `__eq__` and `__ne__` methods.

esteve:
I've addressed all feedback. Going to merge this.

dirk-thomas:
The default values should also be available.

rosidl_generator_py/resource/_msg.py.template
```diff
@@ -0,0 +1,56 @@
...
+@[end for]@
+    ]
+
+@[if len(spec.fields) > 0]@
+    def __init__(self, **kwargs):
+        assert(all(['__' + key in self.__slots__ for key in kwargs.keys()])), \
+            "Invalid arguments passed to constructor: %r" % kwargs.keys()
+@[for field in spec.fields]@
+@[if field.default_value]@
+        self.__@(field.name) = kwargs.get(
```
dirk-thomas: The constructor doesn't need to know about the double underscore prefix. It can do the assignment without and rely on the property setter.

rosidl_generator_py/resource/_msg.py.template
```diff
@@ -0,0 +1,66 @@
...
+            '@(constant.name)': @constant_value_to_py(constant.type, constant.value),
+@[end for]@
+        }
+
+
+class @(spec.base_type.type)(metaclass=@(spec.base_type.type)__Constants):
+
+    __slots__ = [
+@[for field in spec.fields]@
+        '__@(field.name)',
```
dirk-thomas: Since two underscores are commonly used for private members should we use only a single underscore since these should be available to subclasses?

dirk-thomas:
Please don't merge PRs that fast (even if there is one "lgtm"). This PR has no CI jobs and breaks the build: http://ci.ros2.org/job/ci_windows/907/console#console-section-9

esteve:
@dirk-thomas I disagree to some extent, a broken build can be fixed or even the PR can be reverted, however a stagnant PR ends up being a huge burden and from, my experience, that demotivates people from contributing more often. We also agreed that a +1 was good enough and that we trusted the judgement of reviewers (@wjwwood's +1 and your suggestion about quotes). From my side, I concur I should have triggered a build before merging, I'll keep that in mind. In any case, thanks for having a look at this, I really appreciate it.

esteve:
And by the way, don't assume I didn't compile it locally. After being reprimanded so many times :smiley: I always build from scratch everything, including tests. However, I only ran the tests for the `rosidl_generator_py`, when I should have done it for all.

rosidl_generator_py/resource/_msg.py.template
esteve: Fixed in https://github.com/ros2/rosidl/commit/d118609b3cc2031c6abd96e62f82594e7e944045 (PR #68)

rosidl_generator_py/resource/_msg.py.template
esteve: Fixed in https://github.com/ros2/rosidl/commit/d118609b3cc2031c6abd96e62f82594e7e944045 (PR #68)

PR #65, opened 2016-01-15
Field constraints

esteve:
Connects to #67

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.template
+++ b/rosidl_generator_py/resource/_msg.py.template
@@ -67,5 +67,59 @@ class @(spec.base_type.type)(metaclass=@(spec.base_type.type)__Constants):
 
     @@@(field.name).setter
     def @(field.name)(self, value):
+@[if not field.type.is_primitive_type()]@
+        from @(field.type.pkg_name).msg import @(field.type.type)
+@[end if]@
+
+@[if field.type.is_array]@
+        from collections import Sequence
+        from collections import Set
+        from collections import UserList
+        from collections import UserString
+@[elif field.type.type == 'byte']@
+        from collections import ByteString
+@[elif field.type.type in ['char', 'string']]@
+        from collections import UserString
+@[end if]@
+
+        assert isinstance(value, type(None)) or \
+@[if field.type.is_array]@
+            ((isinstance(value, Sequence) or isinstance(value, Set) \
+            or isinstance(value, UserList)) \
+            and not isinstance(value, str) and not isinstance(value, UserString) \
+@[if field.type.array_size]@
+@[if field.type.is_upper_bound]@
+            and len(value) <= @(field.type.array_size) \
+@[else]@
+            and len(value) == @(field.type.array_size) \
+@[end if]@
+            and all([isinstance(v, @(get_python_type(field.type))) for v in value]))
+@[else]@
+            )
+@[end if]@
+@[elif field.type.string_upper_bound]@
+            ((isinstance(value, str) or isinstance(value, UserString)) \
+            and len(value) <= @(field.type.string_upper_bound))
+@[elif not field.type.is_primitive_type()]@
+            isinstance(value, @(field.type.type))
+@[elif field.type.type == 'byte']@
+            ((isinstance(value, bytes) or isinstance(value, ByteString)) \
+                and len(value) == 1)
+@[elif field.type.type == 'char']@
+            ((isinstance(value, str) or isinstance(value, UserString)) \
+            and len(value) == 1)
+@[elif field.type.type in [
+        'bool',
+        'float32', 'float64',
+        'int8', 'uint8',
+        'int16', 'uint16',
+        'int32', 'uint32',
+        'int64', 'uint64',
+        'string',
+    ]]@
+            isinstance(value, @(get_python_type(field.type)))
+@[else]@
+            False
+@[end if]@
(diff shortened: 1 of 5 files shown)


esteve:
CI:

http://ci.ros2.org/job/ci_linux/843/
http://ci.ros2.org/job/ci_osx/706/
http://ci.ros2.org/job/ci_windows/915/

esteve:
Question:

Should we be strict about checking types or should we use `isinstance`? Also, in the case of arrays, should we accept `tuples` and other `iterable` types (except strings) as well? This should go into the article too, but I prefer arguing here with actual code and then paste the conclusion in the article.

wjwwood:
> Should we be strict about checking types or should we use `isinstance`?

Do you mean should we allow subclasses? Maybe you could expand on the two options, as in what would be possible with one versus the other?

> Also, in the case of arrays, should we accept `tuple`s and other `iterable` types (except `string`s) as well?

I think `tuple`s would be ok, but `dict` doesn't really make sense for example.

esteve:
> Do you mean should we allow subclasses? Maybe you could expand on the two options, as in what would be possible with one versus the other?

Yes, I meant that. I'd prefer not supporting subclasses, but @dirk-thomas argued in https://github.com/ros2/rosidl/pull/63 that's a bit too restrictive for checking equality, so I don't know if that applies here too.

> I think tuples would be ok, but dict doesn't really make sense for example.

Yes. How about custom defined `iterable`s? I'd rather restrict it to `tuple`s and `list`s only, though.

esteve:
I've updated this PR to support `tuple`s for static arrays, but not for dynamic arrays.

wjwwood:
I don't see that there is any extra work that needs to be done to support subclassing, nor a good reason to disallow it. Can you provide some justification for avoiding it?

It be nice to be able to assign a tuple to any kind of "array" in the message, even if it just implicitly converts it to a list. It might be worth considering how `np.array` could be used as well, because sometimes converting on assignment might be unnecessarily expensive. So I might go back on my previous statement and agree that any iterable should be allowed so long as it contains the right types.

Also, this got me thinking that you'll need to check the types on publish too. For example, a reasonable error when someone does this would be nice:

```python
msg = Foo()
"""
# Foo.msg
float64[] floats
"""

msg.floats.append(1.3)
msg.floats.append(object())
msg.floats.append(None)
publisher.publish(msg)  # Should get a reasonable error here.
```

dirk-thomas:
Since Pythons philosophy is based on dock typing I think we should follow it. Anything which can 
"look" like the expected data should be allowed. I don't think we should restrict this in any way. As long as we can serialize / deserialize the data into the format specified in the rosidl it should be ok for the user to use that value.

esteve:
One way is, for tuples and lists (and similar objects), to just check that they implement the sequence protocol. Unfortunately, numpy's ndarray does not implement it. Any idea of how to generalize such logic or should we special-case for numpy?

esteve:
I just rebased to resolve the conflicts.

esteve:
I've replaced strict type checking with `isinstance`, checking the `Sequence` protocol instead for arrays. I've also replaced the other type checks (strings, bytes, etc.) with `isinstance`. I'm not checking for `numpy` arrays yet, though, as I'd rather start with the core language and add support for `numpy` later.

rosidl_generator_py/resource/_msg.py.template
```diff
@@ -53,4 +53,21 @@ class @(spec.base_type.type)(metaclass=@(spec.base_type.type)__Constants):
...
+        from collections import Sequence
+        from collections import Set
+        from collections import UserList
+        from collections import UserString
+@[end if]@
+@[if field.type.type == 'byte']@
+        from collections import ByteString
+@[end if]@
+@[if field.type.type in ['char', 'string']]@
+        from collections import UserString
```
dirk-thomas: Will this result in a duplicate import together with line 63?

rosidl_generator_py/resource/_msg.py.template
```diff
@@ -53,4 +53,21 @@ class @(spec.base_type.type)(metaclass=@(spec.base_type.type)__Constants):
...
+@[end if]@
+        @(check_constraints(field.type))
```
dirk-thomas: Instead of constructing this code in Python can this be done in the template instead? The tests should then not test the function but the expected behavior of the generated code.

esteve: I started moving the code into the Python counterpart, but yeah, over time it all looks very convoluted now. I'll move it back to the template.

rosidl_generator_py/resource/_msg.py.template
esteve: Good catch, fixing it now.

esteve:
CI:

http://ci.ros2.org/job/ci_linux/854/
http://ci.ros2.org/job/ci_osx/716/
http://ci.ros2.org/job/ci_windows/937/

dirk-thomas:
lgtm

esteve:
I've squashed the commits and put all the quotes replacements (from double to single) in a separate commit.

CI:

http://ci.ros2.org/job/ci_linux/855/
http://ci.ros2.org/job/ci_osx/717/
http://ci.ros2.org/job/ci_windows/938/

esteve:
I'm going to merge this. I rewrote the history of the commits, but nothing was added/removed.

Issue #67, opened 2016-01-15
Add Python generator


esteve:
- [x] Preliminary Python generator https://github.com/ros2/rosidl/pull/63
- [x] Constraints https://github.com/ros2/rosidl/pull/65
- [x] Defaults https://github.com/ros2/rosidl/pull/69
- [x] Support nested messages https://github.com/ros2/rosidl/pull/113 https://github.com/ros2/rosidl/pull/119
- [ ] Services

esteve:
This is just a ticket to aggregate all the Python related PRs.

mikaelarguedas:
@ros2/team I think the related PRs are ready for a first round of review. Every feedback is welcome. Timers and guard_conditions will be submitted in a separate PR.

mikaelarguedas:
NOTE: the new functions are not documented yet, the goal of submitting this for review is to get a feature / API approval before documenting it

PR #70, opened 2016-01-22
Install generated Python files to a path that can be imported

esteve:
This fixes the `rosidl_generator_py` to install the generated Python files into an appropriate location.

Connects to #67

### DIFF
--- a/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
@@ -35,6 +35,22 @@ foreach(_idl_file ${rosidl_generate_interfaces_IDL_FILES})
   endif()
 endforeach()
 
+if(NOT "${_generated_msg_files} " STREQUAL " ")
+  list(GET _generated_msg_files 0 _msg_file)
+  get_filename_component(_parent_folder "${_msg_file}" DIRECTORY)
+  list(APPEND _generated_msg_files
+    "${_parent_folder}/__init__.py"
+  )
+endif()
+
+if(NOT "${_generated_srv_files} " STREQUAL " ")
+  list(GET _generated_srv_files 0 _srv_file)
+  get_filename_component(_parent_folder "${_srv_file}" DIRECTORY)
+  list(APPEND _generated_srv_files
+    "${_parent_folder}/__init__.py"
+  )
+endif()
+
 set(_dependency_files "")
 set(_dependencies "")
 foreach(_pkg_name ${rosidl_generate_interfaces_DEPENDENCY_PACKAGE_NAMES})
@@ -70,6 +86,8 @@ rosidl_write_generator_arguments(
   TARGET_DEPENDENCIES ${target_dependencies}
 )
 
+file(MAKE_DIRECTORY "${_output_path}")
+file(WRITE "${_output_path}/__init__.py" "")
 add_custom_command(
   OUTPUT ${_generated_msg_files} ${_generated_srv_files}
   COMMAND ${PYTHON_EXECUTABLE} ${rosidl_generator_py_BIN}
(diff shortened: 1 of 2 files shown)


esteve:
CI:

http://ci.ros2.org/job/ci_linux/856/
http://ci.ros2.org/job/ci_osx/718/
http://ci.ros2.org/job/ci_windows/939/

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -35,6 +35,30 @@ foreach(_idl_file ${rosidl_generate_interfaces_IDL_FILES})
...
 
+if(NOT "${_generated_msg_files} " STREQUAL " ")
+  list(GET _generated_msg_files 0 _msg_file)
+  get_filename_component(_parent_folder "${_msg_file}" DIRECTORY)
+  list(APPEND _generated_msg_files
+    "${_parent_folder}/__init__.py"
+  )
+  get_filename_component(_parent_folder "${_parent_folder}" DIRECTORY)
+  list(APPEND _generated_msg_files
+    "${_parent_folder}/__init__.py"
```
dirk-thomas: This adds two different `__init__.py` files but the install rule only has one destination. I don't think the files will be installed in the location you expect them to?

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -96,15 +120,27 @@ add_dependencies(
 
 if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
   if(NOT "${_generated_msg_files} " STREQUAL " ")
+    _ament_cmake_python_get_python_install_dir()
```
dirk-thomas: Calling a private function is not a good idea. It can change any time.

esteve: I know. The comment below explains that I couldn't get `ament_python_install_package` to work properly, so I put this up for review to ask for feedback. I'd rather use `ament_python_install_package` instead of relying on this private function.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
esteve: I've cleaned that up.

wjwwood:
What's the status of this pr? Are we waiting to resolve the issue related to `ament_python_install_package`?

esteve:
@wjwwood yes, I can't get it to work with `ament_python_install_package`. Any ideas @dirk-thomas ?

dirk-thomas:
I need to look into the CMake code to understand what the problem is and what the options. But that has to wait until after the build farm work.

But as far as I can see the PR does not set the `PYTHON_PATH` anywhere. Therefore the generated code (while having `__init__.py` files) is still not importable or am I missing something?

dirk-thomas:
Maybe that would happen automatically when the appropriate CMake function would be called instead of the private function. So probably not necessary to look into until that has happened.

esteve:
@dirk-thomas `PYTHONPATH` is already there:

https://github.com/ros2/rosidl/blob/install-generated-python/rosidl_generator_py/CMakeLists.txt#L52

Does `install()` implicitly depend on the path passed to it? If so, that's maybe why `ament_python_install_package` doesn't work because it only depends on the parent directory where the generated files are meant to be, whereas the current approach depends on the generated files themselves (`DIRECTORY` vs `FILES` in `install()`)

esteve:
@dirk-thomas when `ament_python_install_package` is used the destination directory exists, but it doesn't contain the generated files. When using `install()` with `FILES`, the directory contents look fine.

dirk-thomas:
The line you refer to only sets the `PYTHONPATH` for the test. I was referring to be able to import the Python code after sourcing the setup file from the install space.

esteve:
@dirk-thomas you're right, but in any case, this fails with `ament_python_install_package` because `__init__.py` does not exist in `PACKAGE_DIR`

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
esteve: Have you figured out a way to do this inside `.ament_python_install_package`? I've been digging, but couldn't find a solution.

In the worst case, we can just add a TODO and try to resolve it later, given that this PR is blocking the Python work.

esteve:
@dirk-thomas @ernestmc @jacquelinekay @tfoote @wjwwood any thoughts about this PR? It's blocking the Python development.

PS: I forgot what's the GitHub alias to ping the entire team, does anyone remember?

dirk-thomas:
Fixed PYTHON_PATH in f46ec4304bd0c437e4aae7eb00f9b805aa95598e.

esteve:
@dirk-thomas thanks! I'll squash the commits and trigger a CI.

esteve:
CI:

http://ci.ros2.org/job/ci_linux/907/
http://ci.ros2.org/job/ci_osx/759/
http://ci.ros2.org/job/ci_windows/987/

esteve:
CI passes, this should be okay to merge.

dirk-thomas:
+1

PR #71, opened 2016-01-25
rename message_type_support{_struct}.h

wjwwood:
The motivation for this change is that previously the `.hpp` message/service type support headers would include this as `#include <rosidl_generator_c/message_type_support.h>` for the `rosidl_message_type_support_t` structure. The problem is that this collides with the pattern of having a generic declaration, e.g. `get_type_support<MsgType>()` or `ROSIDL_GET_MSG_TYPE_SUPPORT(MsgPkg, Msg)()`, in a header called `message_type_support_decl.hpp` and then have each middleware impl provide its own `message_type_support.hpp`. Now that I'm writing the C versions of those middleware implementations, they'd like to use the same pattern without the `pp` suffix, which is when the collision comes.

I considered naming it `message_type_support_decl.h`, but since it only contains a structure now and it also included by the C++ versions of the headers I thought this was more appropriate.

There will be several other pull requests to this effect as it touches several packages. This is also the first thing I plan to change in a series of small changes in order to open a pr for the OpenSplice C type support pr.


rosidl_generator_c/include/rosidl_generator_c/message_type_support_struct.h
```diff
@@ -28,19 +28,8 @@ typedef struct ROSIDL_PUBLIC_TYPE rosidl_message_type_support_t
...
- * function for a specific message type. The library of the message package
- * which defines a given message will provide the symbol to which this macro
- * expands.
- */
-#define ROSIDL_GET_TYPE_SUPPORT(MsgPkgName, MsgSubfolder, MsgName) \
-  rosidl_get_message_type_support__ ## MsgPkgName ## __ ## MsgSubfolder ## __ ## MsgName()
-
-#define ROSIDL_GET_MESSAGE_TYPE_SUPPORT(MsgPkgName, MsgName) \
-  ROSIDL_GET_TYPE_SUPPORT(MsgPkgName, msg, MsgName)
-
```
wjwwood: Also, because of how these macros need to work, we cannot declare them here and put the implementation specific version of them elsewhere (as we did with template functions in C++).

rosidl_typesupport_introspection_c/include/rosidl_typesupport_introspection_c/impl/rosidl_generator_c/message_type_support.h
```diff
@@ -0,0 +1,43 @@
...
+ * the OpenSplice shared library symbols.
+ */
+#define ROSIDL_GET_MSG_TYPE_SUPPORT(PkgName, MsgName) \
+  ROSIDL_GET_TYPE_SUPPORT(PkgName, msg, MsgName)
+
+#define ROSIDL_GET_TYPE_SUPPORT(PkgName, MsgSubfolder, MsgName) \
+  ROSIDL_GET_TYPE_SUPPORT_FUNCTION(PkgName, MsgSubfolder, MsgName)()
+
+#define ROSIDL_GET_TYPE_SUPPORT_FUNCTION(PkgName, MsgSubfolder, MsgName) \
+  rosidl_typesupport_introspection_c_get_message__ ## PkgName ## __ ## MsgSubfolder ## __ ## MsgName
```
wjwwood: These functions still need to be implemented, but they were not implemented before (when they were called `rosidl_get_message_type_support__ ## MsgPkgName ## __ ## MsgSubfolder ## __ ## MsgName()`). I think we should implement them in a follow up pull request which addresses the other things missing to allow these introspection based type supports to support complex types like arrays and strings.

rosidl_typesupport_introspection_c/include/rosidl_typesupport_introspection_c/impl/rosidl_generator_c/message_type_support.h
```diff
@@ -0,0 +1,43 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
```
dirk-thomas: Shouldn't this be `2015-2016` since part of the code came is coming from a previously existing file?

rosidl_typesupport_introspection_c/include/rosidl_typesupport_introspection_c/impl/rosidl_generator_c/message_type_support.h
```diff
@@ -0,0 +1,43 @@
...
+#define ROSIDL_GET_MSG_TYPE_SUPPORT(PkgName, MsgName) \
+  ROSIDL_GET_TYPE_SUPPORT(PkgName, msg, MsgName)
+
+#define ROSIDL_GET_TYPE_SUPPORT(PkgName, MsgSubfolder, MsgName) \
+  ROSIDL_GET_TYPE_SUPPORT_FUNCTION(PkgName, MsgSubfolder, MsgName)()
+
+#define ROSIDL_GET_TYPE_SUPPORT_FUNCTION(PkgName, MsgSubfolder, MsgName) \
+  rosidl_typesupport_introspection_c_get_message__ ## PkgName ## __ ## MsgSubfolder ## __ ## MsgName
+
+#define USING_ROSIDL_TYPESUPPORT_INTROSPECTION_C 1
```
dirk-thomas: Can you please comment on what this is needed for.

wjwwood: Ah, this is new and based on how I implement these functions in `rmw_opensplice`. Basically the C file which implements these functions will static assert that this is set. The reason being that each type support package will provide this exact header (in the same namespace) and it would potentially catch an error where the C file for this header would accidentally include the wrong version of this header.

Since this doesn't have the C implementation yet, I can remove it if you like.

rosidl_typesupport_introspection_c/include/rosidl_typesupport_introspection_c/impl/rosidl_generator_c/message_type_support.h
wjwwood: Yes.

wjwwood: Oh wait, no I don't think this file ever existed as it is here before. I mean it was similar to the header in `rosidl_generator_c` in that it implements the `ROSIDL_*` macros, but each type support package will have a similar header, so it's not a clear moved from here to there situation.

wjwwood:
CI:

- http://ci.ros2.org/job/ci_linux/858/
- http://ci.ros2.org/job/ci_osx/720/
- http://ci.ros2.org/job/ci_windows/942/

wjwwood:
I had some local changes I forgot to push. Pushed in 580c523. New CI:

- http://ci.ros2.org/job/ci_linux/859/
- http://ci.ros2.org/job/ci_osx/721/
- http://ci.ros2.org/job/ci_windows/943/

wjwwood:
Forgot one more line, CI again:

- http://ci.ros2.org/job/ci_linux/860/
- http://ci.ros2.org/job/ci_osx/722/
- http://ci.ros2.org/job/ci_windows/944/

wjwwood:
I forgot to include changes to how the local `rosidl_generator_c` tests include their headers (26f2da5). CI again:

- http://ci.ros2.org/job/ci_linux/861/
- http://ci.ros2.org/job/ci_osx/723/
- http://ci.ros2.org/job/ci_windows/945/

wjwwood:
Had to fix a bug in the `rmw_connext` patches. However, moving further I realize that this cannot be merged until I add a `rosidl_generator_c/message_type_support.h` header into `rosidl_typesupport_connext_cpp`'s `include/rosidl_typesupport_connext_cpp/impl` folder.

I'll have to add an unimplemented one for now (as I have done with `rosidl_typesupport_introspection_cpp`) until I (or someone) has time to implement the Connext type support as I have done in OpenSplice's type support packages.

jacquelinekay:
@wjwwood I could look into implementing the Opensplice typesupport.

rosidl_typesupport_introspection_c/include/rosidl_typesupport_introspection_c/impl/rosidl_generator_c/message_type_support.h
jacquelinekay: So, to clarify, the C++ equivalent of of this file would be the same, except it would define `USING_ROSIDL_TYPESUPPORT_INTROSPECTION_C` as 0?

jacquelinekay:
Ok, after talking to William yesterday, it sounds like the process will be: add the placeholder for `rosidl_generator_c/message_type_support.h` for Connext (so that it actually builds), merge these changes, then filling in the implementations in the next set of pull requests.

rosidl_typesupport_introspection_c/include/rosidl_typesupport_introspection_c/impl/rosidl_generator_c/message_type_support.h
dirk-thomas: if the current set of changes doesn't need this yet then I would suggest we keep it out for now. With the next set of changes it might be more obvious how this is being used and why this is necessary.

wjwwood: @jacquelinekay no that's not it. This file is implemented by exactly one C file. It is not included by any CPP specific headers or files. In that single C file I use the existence of this define (the value doesn't matter) to check to make sure this header was included and that some other header was not, because there is a version of this header for each type support implementation.

wjwwood:
Ok, I had to do a few more things (open a few more PR's), but I finally have passing CI:

- http://ci.ros2.org/job/ci_linux/872
- http://ci.ros2.org/job/ci_osx/730
- http://ci.ros2.org/job/ci_windows/954
 - There were some test failures, but they are those flaky parameter server hanging problems I think.

I'd like to merge this tomorrow if there is no additional feedback.

jacquelinekay:
When I build these branches locally I still get a compile error about `rosidl_generator_c/message_type_support.h` missing in `rcl`. When I build with `--isolated` it works, however (which is why it builds on Jenkins).

rosidl_typesupport_introspection_c/CMakeLists.txt
```diff
@@ -11,7 +11,10 @@ find_package(ament_cmake_python REQUIRED)
 
 ament_export_dependencies(rosidl_cmake)
 ament_export_dependencies(rosidl_generator_c)
-ament_export_include_directories(include)
+# The reason the impl folder is exported is that it contains the implementation
+# for the get_*_type_support_handle functions and defines the opensplice
```
jacquelinekay: Is this comment correct? It seems to be specific to the Opensplice typesupport package.

jacquelinekay:
When I build without `--isolated`, I get these CMake warnings:

```
CMake Warning at CMakeLists.txt:19 (find_package):
  By not providing "Findrosidl_typesupport_opensplice_c.cmake" in
  CMAKE_MODULE_PATH this project has asked CMake to find a package
  configuration file provided by "rosidl_typesupport_opensplice_c", but CMake
  did not find one.

  Could not find a package configuration file provided by
  "rosidl_typesupport_opensplice_c" with any of the following names:

    rosidl_typesupport_opensplice_cConfig.cmake
    rosidl_typesupport_opensplice_c-config.cmake

  Add the installation prefix of "rosidl_typesupport_opensplice_c" to
  CMAKE_PREFIX_PATH or set "rosidl_typesupport_opensplice_c_DIR" to a
  directory containing one of the above files.  If
  "rosidl_typesupport_opensplice_c" provides a separate development package
  or SDK, be sure it has been installed.


-- Found rosidl_typesupport_opensplice_cpp: 0.0.0 (/home/jackie/code/ros2_ws/install/share/rosidl_typesupport_opensplice_cpp/cmake)
-- Found OpenSplice: /usr/lib/libcmxml.so;/usr/lib/libcommonserv.so;/usr/lib/libdcpsgapi.so;/usr/lib/libdcpssac.so;/usr/lib/libddsconfparser.so;/usr/lib/libddsconf.so;/usr/lib/libddsdatabase.so;/usr/lib/libddsi2.so;/usr/lib/libddskernel.so;/usr/lib/libddsosnet.so;/usr/lib/libddsos.so;/usr/lib/libddsserialization.so;/usr/lib/libddsuser.so;/usr/lib/libddsutil.so;/usr/lib/libdurability.so;/usr/lib/libspliced.so;/usr/lib/libdcpsisocpp.so;/usr/lib/libdcpssacpp.so  
-- Found PrismTech OpenSplice: /usr/share/opensplice/cmake
-- Found OpenSplice: /usr/share/opensplice/cmake/../../../include/opensplice;/usr/share/opensplice/cmake/../../../include/opensplice/sys;/usr/share/opensplice/cmake/../../../include/opensplice/dcps/C++/SACPP  
CMake Warning at CMakeLists.txt:23 (message):
  Could not find ROSIDL TypeSupport for OpenSplice
  (rosidl_typesupport_opensplice_cpp or rosidl_typesupport_opensplice_c) -
  skipping 'rmw_opensplice_cpp'
```

I'm debugging this right now since @wjwwood is taking today off. If I find a fix, I'll commit it and re-run CI. This issue is what is currently blocking my +1.

jacquelinekay:
non-isolated build is fixed now. (We should come up with a name for that, especially if isolated is going to become the default at some point)

+1 to these changes

wjwwood:
@jacquelinekay did you clear your workspace before building without `--isolated`?

jacquelinekay:
no, but I nuked my install folder and did a --force-cmake-config. Adding a missing `ament_package()` to the CMakeLists.txt of rosidl_typesupport_opensplice_c fixed the issue though.

wjwwood:
With @jacquelinekay's fix everything still works for me locally as well as on the farm. So I'm going to merge these and start the next round of changes.

PR #72, opened 2016-01-27
Rename identifiers

esteve:
This just uses a different value for the identifiers of each of the introspection implementations.


esteve:
CI:

http://ci.ros2.org/job/ci_linux/867/
http://ci.ros2.org/job/ci_osx/727/
http://ci.ros2.org/job/ci_windows/950/

dirk-thomas:
+1

PR #73, opened 2016-01-29
expose type support for rclpy

esteve:
This PR adds the machinery for `rclpy` to be able to obtain the type support for a given Python class, and to convert from and to ROS C messages.

Depends on #70 

Connects to #67

### DIFF
--- a/python_cmake_module/CMakeLists.txt
+++ b/python_cmake_module/CMakeLists.txt
@@ -1,0 +1,17 @@
+cmake_minimum_required(VERSION 2.8.3)
+
+project(python_cmake_module NONE)
+
+find_package(ament_cmake REQUIRED)
+
+if(AMENT_ENABLE_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
+endif()
+
+ament_package(
+  CONFIG_EXTRAS "python_cmake_module-extras.cmake"
+)
+
+install(DIRECTORY cmake
+  DESTINATION share/${PROJECT_NAME})
--- a/python_cmake_module/package.xml
+++ b/python_cmake_module/package.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>python_cmake_module</name>
+  <version>0.0.0</version>
+  <description>Provide CMake module with extra functionality for Python.</description>
+
+  <maintainer email="esteve@osrfoundation.org">Esteve Fernandez</maintainer>
+  <license>Apache License 2.0</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <buildtool_export_depend>python3-dev</buildtool_export_depend>
+
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
+
+  <export>
+    <build_type>ament_cmake</build_type>
+  </export>
+</package>
--- a/rosidl_generator_py/CMakeLists.txt
+++ b/rosidl_generator_py/CMakeLists.txt
@@ -3,13 +3,17 @@ cmake_minimum_required(VERSION 2.8.3)
 project(rosidl_generator_py)
 
 find_package(ament_cmake REQUIRED)
-find_package(ament_cmake_python REQUIRED)
 
 ament_export_dependencies(rosidl_cmake)
 
 ament_python_install_package(${PROJECT_NAME})
 
 if(AMENT_ENABLE_TESTING)
+  find_package(ament_cmake_nose REQUIRED)
+
+  find_package(rosidl_cmake REQUIRED)
+  find_package(rosidl_generator_c REQUIRED)
+
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
 
@@ -48,8 +52,14 @@ if(AMENT_ENABLE_TESTING)
     SKIP_INSTALL
   )
 
+  set(_append_library_dirs "")
+  if(WIN32)
+    set(_append_library_dirs "${CMAKE_CURRENT_BINARY_DIR}/$<CONFIGURATION>")
+  endif()
+
   ament_add_nose_test(test_interfaces_py "test/test_interfaces.py"
-    APPEND_ENV PYTHONPATH=${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py:${CMAKE_CURRENT_SOURCE_DIR}
+    APPEND_ENV PYTHONPATH="${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py:${CMAKE_CURRENT_SOURCE_DIR}"
+    APPEND_LIBRARY_DIRS "${_append_library_dirs}"
     WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py"
   )
 endif()
--- a/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
@@ -12,20 +12,71 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+find_package(rosidl_generator_c REQUIRED)
+find_package(rmw_implementation_cmake REQUIRED)
+find_package(rmw REQUIRED)
+
+# NOTE(esteve): required for CMake-2.8 in Ubuntu 14.04
+set(Python_ADDITIONAL_VERSIONS 3.4)
+find_package(PythonInterp 3.4 REQUIRED)
+
+find_package(python_cmake_module REQUIRED)
+find_package(PythonExtra MODULE)
+
+# TODO(esteve): force opensplice and connext C type supports only, uncomment
+# the following line when all typesupport implementations are ported to C
+#get_rmw_typesupport(_typesupport_impls ${rmw_implementation})
+set(_typesupport_impls "")
+foreach(_extension IN LISTS AMENT_EXTENSIONS_rosidl_generate_interfaces)
+  string(REPLACE ":" ";" _extension_list "${_extension}")
+  list(LENGTH _extension_list _length)
+  if(NOT _length EQUAL 2)
+    message(FATAL_ERROR "ament_execute_extensions(${extension_point}) "
+      "registered extension '${_extension}' can not be split into package "
+      "name and cmake filename")
+  endif()
+  list(GET _extension_list 0 _pkg_name)
+  list(GET _extension_list 1 _cmake_filename)
+  if("${_pkg_name} " STREQUAL "rosidl_typesupport_opensplice_c ")
+    list(APPEND _typesupport_impls "rosidl_typesupport_opensplice_c")
+  endif()
+  if("${_pkg_name} " STREQUAL "rosidl_typesupport_connext_c ")
+    list(APPEND _typesupport_impls "rosidl_typesupport_connext_c")
+  endif()
+endforeach()
+
 set(_output_path
   "${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py/${PROJECT_NAME}")
-set(_generated_msg_files "")
+set(_generated_msg_py_files "")
+set(_generated_msg_c_files "")
+set(_generated_msg_c_common_files "")
+set(_generated_msg_c_ts_files "")
 set(_generated_srv_files "")
 foreach(_idl_file ${rosidl_generate_interfaces_IDL_FILES})
   get_filename_component(_parent_folder "${_idl_file}" DIRECTORY)
   get_filename_component(_parent_folder "${_parent_folder}" NAME)
-  get_filename_component(_msg_name "${_idl_file}" NAME_WE)
-  string_camel_case_to_lower_case_underscore("${_msg_name}" _module_name)
+  get_filename_component(_msg_name1 "${_idl_file}" NAME_WE)
+  string_camel_case_to_lower_case_underscore("${_msg_name1}" _module_name)
 
   if("${_parent_folder} " STREQUAL "msg ")
-    list(APPEND _generated_msg_files
+    list(APPEND _generated_msg_py_files
       "${_output_path}/${_parent_folder}/_${_module_name}.py"
     )
+    list(APPEND _generated_msg_c_files
+      "${_output_path}/${_parent_folder}/_${_module_name}_s.c"
+    )
+    list(APPEND _generated_msg_c_common_files
+      "${_output_path}/${_parent_folder}/_${_module_name}_s.c"
+    )
+    foreach(_typesupport_impl ${_typesupport_impls})
+      list(APPEND _generated_msg_c_files
+        "${_output_path}/${_parent_folder}/_${_module_name}_s.ep.${_typesupport_impl}.c"
+      )
+      list(APPEND _generated_msg_c_ts_files
+        "${_output_path}/${_parent_folder}/_${_module_name}_s.ep.${_typesupport_impl}.c"
+      )
+      list(APPEND _type_support_by_generated_msg_c_files "${_typesupport_impl}")
+    endforeach()
   elseif("${_parent_folder} " STREQUAL "srv ")
     list(APPEND _generated_srv_files
       "${_output_path}/${_parent_folder}/_${_module_name}.py"
@@ -35,10 +86,10 @@ foreach(_idl_file ${rosidl_generate_interfaces_IDL_FILES})
   endif()
 endforeach()
 
-if(NOT "${_generated_msg_files} " STREQUAL " ")
-  list(GET _generated_msg_files 0 _msg_file)
+if(NOT "${_generated_msg_py_files} " STREQUAL " ")
+  list(GET _generated_msg_py_files 0 _msg_file)
   get_filename_component(_parent_folder "${_msg_file}" DIRECTORY)
-  list(APPEND _generated_msg_files
+  list(APPEND _generated_msg_py_files
     "${_parent_folder}/__init__.py"
   )
 endif()
@@ -65,6 +116,8 @@ endforeach()
 set(target_dependencies
   "${rosidl_generator_py_BIN}"
   ${rosidl_generator_py_GENERATOR_FILES}
+  "${rosidl_generator_py_TEMPLATE_DIR}/_msg_support.c.template"
+  "${rosidl_generator_py_TEMPLATE_DIR}/_msg_support.entry_point.c.template"
   "${rosidl_generator_py_TEMPLATE_DIR}/_msg.py.template"
   "${rosidl_generator_py_TEMPLATE_DIR}/_srv.py.template"
   ${rosidl_generate_interfaces_IDL_FILES}
@@ -88,33 +141,135 @@ rosidl_write_generator_arguments(
 
 file(MAKE_DIRECTORY "${_output_path}")
 file(WRITE "${_output_path}/__init__.py" "")
+
+set(_generated_extension_files "")
+set(_extension_dependencies "")
+set(_target_suffix "__py")
+
 add_custom_command(
-  OUTPUT ${_generated_msg_files} ${_generated_srv_files}
+  # OUTPUT ${_generated_msg_py_files} ${_generated_msg_c__${_typesupport_impl}_files} ${_generated_srv_files}
+  OUTPUT ${_generated_msg_py_files} ${_generated_msg_c_files} ${_generated_srv_files}
   COMMAND ${PYTHON_EXECUTABLE} ${rosidl_generator_py_BIN}
   --generator-arguments-file "${generator_arguments_file}"
+  --typesupport-impl "${_typesupport_impl}"
+  --typesupport-impls "${_typesupport_impls}"
   DEPENDS ${target_dependencies}
   COMMENT "Generating Python code for ROS interfaces"
   VERBATIM
 )
 
-if(TARGET ${rosidl_generate_interfaces_TARGET}__py)
-  message(WARNING "Custom target ${rosidl_generate_interfaces_TARGET}__py already exists")
+macro(set_properties _build_type)
+  set_target_properties(${_msg_name}${_pyext_suffix} PROPERTIES
+    COMPILE_FLAGS "${_extension_compile_flags}"
+    PREFIX ""
+    LIBRARY_OUTPUT_DIRECTORY${_build_type} "${_output_path}/${_parent_folder}"
+    RUNTIME_OUTPUT_DIRECTORY${_build_type} "${_output_path}/${_parent_folder}"
+    OUTPUT_NAME "${_base_msg_name}__${_typesupport_impl}${PythonExtra_EXTENSION_SUFFIX}"
+    SUFFIX "${PythonExtra_EXTENSION_EXTENSION}")
+endmacro()
+
+foreach(_generated_msg_c_ts_file ${_generated_msg_c_ts_files})
+  get_filename_component(_full_folder "${_generated_msg_c_ts_file}" DIRECTORY)
+  get_filename_component(_parent_folder "${_full_folder}" NAME)
+  get_filename_component(_base_msg_name "${_generated_msg_c_ts_file}" NAME_WE)
+  get_filename_component(_full_extension_msg_name "${_generated_msg_c_ts_file}" EXT)
+
+  set(_msg_name "${_base_msg_name}${_full_extension_msg_name}")
+
+  # foreach(_typesupport_impl ${_typesupport_impls})
+    list(FIND _generated_msg_c_ts_files ${_generated_msg_c_ts_file} _file_index)
+    list(GET _type_support_by_generated_msg_c_files ${_file_index} _typesupport_impl)
+    find_package(${_typesupport_impl} REQUIRED)
+    set(_generated_msg_c_common_file "${_full_folder}/${_base_msg_name}.c")
+
+    set(_pyext_suffix "__pyext")
+
+    # TODO(esteve): Change the following code so that each file is compiled independently
+    add_library(${_msg_name}${_pyext_suffix} SHARED
+      "${_generated_msg_c_ts_file}"
+      "${_generated_msg_c_common_file}"
+    )
+
+    set_properties("")
+    if(WIN32)
+      set_properties("_DEBUG")
+      set_properties("_MINSIZEREL")
+      set_properties("_RELEASE")
+      set_properties("_RELWITHDEBINFO")
+    endif()
+
+    add_dependencies(
+      ${_msg_name}${_pyext_suffix}
+      ${rosidl_generate_interfaces_TARGET}__rosidl_generator_c
+    )
+
+    set(_extension_compile_flags "")
+    if(NOT WIN32)
+      set(_extension_compile_flags "-Wall -Wextra")
+    endif()
+
+    target_link_libraries(
+      ${_msg_name}${_pyext_suffix}
+      ${PythonExtra_LIBRARIES}
+      ${PROJECT_NAME}__${_typesupport_impl}
+    )
+
+    list(APPEND _generated_extension_files
+      "$<TARGET_FILE:${_msg_name}${_pyext_suffix}>"
+    )
+
+    target_include_directories(${_msg_name}${_pyext_suffix}
+      PUBLIC
+      ${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_c
+      ${PythonExtra_INCLUDE_DIRS}
+    )
+
+    foreach(_pkg_name ${rosidl_generate_interfaces_DEPENDENCY_PACKAGE_NAMES})
+      ament_target_dependencies(
+        ${_msg_name}${_pyext_suffix}
+        ${_pkg_name}
+      )
+    endforeach()
+    ament_target_dependencies(${_msg_name}${_pyext_suffix}
+      "rosidl_generator_c"
+      "${_typesupport_impl}"
+    )
+
+    list(APPEND _extension_dependencies ${_msg_name}${_pyext_suffix})
+
+    ament_target_dependencies(${_msg_name}${_pyext_suffix}
+      ${_typesupport_impl}
+    )
+    add_dependencies(${_msg_name}${_pyext_suffix}
+      ${rosidl_generate_interfaces_TARGET}__${_typesupport_impl}
+    )
+
+  # endforeach()
+  ament_target_dependencies(${_msg_name}${_pyext_suffix}
+    "rosidl_generator_c"
+    "rosidl_generator_py"
+    "${PROJECT_NAME}__rosidl_generator_c"
+  )
+endforeach()
+
+if(TARGET ${rosidl_generate_interfaces_TARGET}${_target_suffix})
+  message(WARNING "Custom target ${rosidl_generate_interfaces_TARGET}${_target_suffix} already exists")
 else()
   add_custom_target(
-    ${rosidl_generate_interfaces_TARGET}__py
+    ${rosidl_generate_interfaces_TARGET}${_target_suffix}
     DEPENDS
-    ${_generated_msg_files} ${_generated_srv_files}
+    ${_generated_msg_py_files}
+    ${_generated_msg_c_files}
+    ${_generated_srv_files}
+    ${_generated_extension_files}
+    ${_extension_dependencies}
+    ${rosidl_generate_interfaces_TARGET}__rosidl_generator_c
   )
 endif()
 
-add_dependencies(
-  ${rosidl_generate_interfaces_TARGET}
-  ${rosidl_generate_interfaces_TARGET}__py
-)
-
 if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
-  if(NOT "${_generated_msg_files} " STREQUAL " ")
-    list(GET _generated_msg_files 0 _msg_file)
+  if(NOT "${_generated_msg_py_files} " STREQUAL " ")
+    list(GET _generated_msg_py_files 0 _msg_file)
     get_filename_component(_msg_package_dir "${_msg_file}" DIRECTORY)
     get_filename_component(_msg_package_dir "${_msg_package_dir}" DIRECTORY)
     ament_python_install_package("${PROJECT_NAME}" PACKAGE_DIR "${_msg_package_dir}")
--- a/rosidl_generator_py/package.xml
+++ b/rosidl_generator_py/package.xml
@@ -9,14 +9,27 @@
   <buildtool_depend>ament_cmake</buildtool_depend>
 
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
+  <buildtool_export_depend>python_cmake_module</buildtool_export_depend>
   <buildtool_export_depend>rosidl_cmake</buildtool_export_depend>
 
+  <exec_depend>rmw_implementation</exec_depend>
+  <exec_depend>rmw_implementation_cmake</exec_depend>
+  <exec_depend>rosidl_generator_c</exec_depend>
   <exec_depend>rosidl_parser</exec_depend>
 
   <test_depend>ament_cmake_nose</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
+  <test_depend>python_cmake_module</test_depend>
+  <test_depend>rmw_implementation</test_depend>
+  <test_depend>rmw_implementation_cmake</test_depend>
+  <test_depend>rosidl_generator_c</test_depend>
+  <test_depend>rosidl_parser</test_depend>
   <test_depend>rosidl_cmake</test_depend>
+  <!-- Duplicated from rosidl_default_generator in order to avoid a circular dependency. -->
+  <test_depend>rosidl_typesupport_connext_c</test_depend>
+  <test_depend>rosidl_typesupport_introspection_c</test_depend>
+  <test_depend>rosidl_typesupport_opensplice_c</test_depend>
 
   <export>
     <build_type>ament_cmake</build_type>
--- a/rosidl_generator_py/resource/_msg.py.template
+++ b/rosidl_generator_py/resource/_msg.py.template
@@ -1,9 +1,35 @@
 # generated from rosidl_generator_py/resource/_msg.py.template
 # generated code does not contain a copyright notice
 
+import logging
+import traceback
+
+_convert_from_py = None
+_convert_to_py = None
+_type_support = None
+__type_support_importable = False
+try:
+    import rclpy
+    from rosidl_generator_py import import_type_support
+    __type_support_importable = True
+except ImportError:
+    logger = logging.getLogger('rosidl_generator_py.@(spec.base_type.type)')
+    logger.debug('Failed to import needed modules for type support:\n' + traceback.format_exc())
+
+if __type_support_importable:
+    rclpy_implementation = rclpy._rclpy.rclpy_get_implementation_identifier()
+    module = import_type_support('@(package_name)', '@(subfolder)', '@(module_name)', rclpy_implementation)
+    _convert_from_py = module.convert_from_py
+    _convert_to_py = module.convert_to_py
+    _type_support = module.type_support
+
 
 class Metaclass(type):
     """Metaclass of message '@(spec.base_type.type)'."""
+
+    _CONVERT_FROM_PY = _convert_from_py
+    _CONVERT_TO_PY = _convert_to_py
+    _TYPE_SUPPORT = _type_support
 
     __constants = {
 @[for constant in spec.constants]@
@@ -72,7 +98,23 @@ class @(spec.base_type.type)(metaclass=Metaclass):
         self.@(field.name) = kwargs.get(
             '@(field.name)', @(spec.base_type.type).@(field.name.upper())__DEFAULT)
 @[else]@
-        self.@(field.name) = kwargs.get('@(field.name)')
+@[if not field.type.is_primitive_type()]@
+        from @(field.type.pkg_name).msg import @(field.type.type)
+@[end if]@
+@[if field.type.type == 'byte']@
+        self.@(field.name) = kwargs.get('@(field.name)', b'0')
+@[elif field.type.type == 'char']@
+        self.@(field.name) = kwargs.get('@(field.name)', '\0')
+@[elif field.type.array_size]
+        self.@(field.name) = kwargs.get(
+            '@(field.name)',
+            tuple([@(get_python_type(field.type))() for x in range(@(field.type.array_size))])
+        )
+@[elif field.type.is_array]
+        self.@(field.name) = kwargs.get('@(field.name)', list())
+@[else]@
+        self.@(field.name) = kwargs.get('@(field.name)', @(get_python_type(field.type))())
+@[end if]@
 @[end if]@
 @[end for]@
 @[end if]@
--- a/rosidl_generator_py/resource/_msg_support.c.template
+++ b/rosidl_generator_py/resource/_msg_support.c.template
@@ -1,0 +1,160 @@
+#include <Python.h>
+
+#include <@(spec.base_type.pkg_name)/@(subfolder)/@(module_name)__struct.h>
+#include <@(spec.base_type.pkg_name)/@(subfolder)/@(module_name)__functions.h>
+
+@{
+have_not_included_primitive_arrays = True
+have_not_included_string = True
+}@
+@[for field in spec.fields]@
+@[if field.type.is_array and have_not_included_primitive_arrays]@
+@{have_not_included_primitive_arrays = False}@
+#include <rosidl_generator_c/primitives_array.h>
+#include <rosidl_generator_c/primitives_array_functions.h>
+
+@[end if]@
+@[if field.type.type == 'string' and have_not_included_string]@
+@{have_not_included_string = False}@
+#include <rosidl_generator_c/string.h>
+#include <rosidl_generator_c/string_functions.h>
+
+@[end if]@
+@[end for]@
+
+@{
+msg_typename = '%s__%s__%s' % (spec.base_type.pkg_name, subfolder, spec.base_type.type)
+}@
+
+/* TODO(esteve): implement conversion function */
+void * @(spec.base_type.pkg_name)_@(module_name)__convert_from_py(PyObject *_pymsg)
+{
+  @(msg_typename) * ros_message = @(msg_typename)__create();
+  (void)ros_message;
+@{
+full_classname = "@(spec.base_type.pkg_name).@(subfolder)._@(module_name).@(spec.base_type.type)"
+}@
+  char full_classname_dest[@(len(full_classname) + 1)];
+
+  /* TODO(esteve): assume 1BYTE encoding for unicode strings. */
+  char * class_name = (char *)PyUnicode_1BYTE_DATA(
+    PyObject_GetAttrString(PyObject_GetAttrString(_pymsg, "__class__"), "__name__"));
+  char * module_name = (char *)PyUnicode_1BYTE_DATA(
+    PyObject_GetAttrString(PyObject_GetAttrString(_pymsg, "__class__"), "__module__"));
+
+  snprintf(full_classname_dest, @(len(full_classname) + 1), "%s.%s", class_name, module_name);
+
+  assert(strncmp(
+    "@(spec.base_type.pkg_name).@(subfolder)._@(module_name).@(spec.base_type.type)",
+    full_classname_dest, @(len(full_classname))));
+
+@[for field in spec.fields]@
+  PyObject * py@(field.name) = PyObject_GetAttrString(_pymsg, "@(field.name)");
+  (void)py@(field.name);
+@[if not field.type.is_primitive_type()]@
+// TODO(esteve): add support for nested types
+@[end if]@
+
+@[if field.type.is_array]@
+  assert(PySequence_Check(py@(field.name)));
+@[elif field.type.type in ['byte', 'char']]@
+  assert(PyBytes_Check(py@(field.name)));
+  ros_message->@(field.name) = PyBytes_AS_STRING(py@(field.name))[0];
+@[elif field.type.type in ['string']]@
+  assert(PyUnicode_Check(py@(field.name)));
+  rosidl_generator_c__String__assign(
+    &ros_message->@(field.name), (char *)PyUnicode_1BYTE_DATA(py@(field.name)));
+@[elif field.type.type =='bool']@
+  assert(PyBool_Check(py@(field.name)));
+  ros_message->@(field.name) = (py@(field.name) == Py_True);
+@[elif field.type.type in ['float32', 'float64']]@
+  assert(PyFloat_Check(py@(field.name)));
+  ros_message->@(field.name) = PyFloat_AS_DOUBLE(py@(field.name));
+@[elif field.type.type in [
+        'int8',
+        'int16',
+        'int32',
+    ]]@
+  assert(PyLong_Check(py@(field.name)));
+  ros_message->@(field.name) = PyLong_AsLong(py@(field.name));
+@[elif field.type.type in [
+        'uint8',
+        'uint16',
+        'uint32',
+    ]]@
+  assert(PyLong_Check(py@(field.name)));
+  ros_message->@(field.name) = PyLong_AsUnsignedLong(py@(field.name));
+@[elif field.type.type == 'int64']@
+  assert(PyLong_Check(py@(field.name)));
+  ros_message->@(field.name) = PyLong_AsLongLong(py@(field.name));
+@[elif field.type.type == 'uint64']@
+  assert(PyLong_Check(py@(field.name)));
+  ros_message->@(field.name) = PyLong_AsUnsignedLongLong(py@(field.name));
+@[else]@
+  assert(false);
+@[end if]@
+@[end for]@
+  assert(ros_message != NULL);
+  return ros_message;
+}
+
+/* TODO(esteve): implement conversion function */
+PyObject * @(spec.base_type.pkg_name)_@(module_name)__convert_to_py(void * raw_ros_message)
+{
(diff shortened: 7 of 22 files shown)


esteve:
CI:

http://ci.ros2.org/job/ci_linux/873/
http://ci.ros2.org/job/ci_osx/731/
http://ci.ros2.org/job/ci_windows/955/

esteve:
I forgot to build it locally with `--isolated`

esteve:
I'll fix the dependencies and trigger the CI again.

jacquelinekay:
I'm having trouble building this locally with or without isolated.

esteve:
@jacquelinekay could you post what's the error output of your build? Thanks!

jacquelinekay:
```
[  9%] make[2]: *** [CMakeFiles/_float64_support__pyext.dir/rosidl_generator_py/rosidl_generator_py/msg/_byte_support.c.o] Error 1
[ 10%] /home/jackie/code/ros2_ws/build_isolated/rosidl_generator_py/rosidl_generator_py/rosidl_generator_py/msg/_char_support.c:3:53: fatal error: rosidl_generator_c/message_type_support.h: No such file or directory
 #include <rosidl_generator_c/message_type_support.h>
```

esteve:
@jacquelinekay yeah, that's the error I'm seeing with `--isolated`, but not without it.

esteve:
@jacquelinekay this now compiles locally with `--isolated`, however it has an issue with not linking against the type support libraries properly.

jacquelinekay:
yep, it builds for me now! I don't see a linking issue with a non-isolated build. I haven't tried building it with rclpy though.

rosidl_generator_py/resource/_msg_support.c.template
```diff
@@ -0,0 +1,78 @@
+#include <Python.h>
+
+#include <rosidl_generator_c/message_type_support.h>
+#include <@(spec.base_type.pkg_name)/@(subfolder)/@(module_name).h>
+
+static PyMethodDef @(spec.base_type.pkg_name)_@(module_name)__methods[] = {
+  {NULL, NULL, 0, NULL} /* sentinel */
```
jacquelinekay: Put 2 spaces between code and inline comments.

rosidl_generator_py/resource/_msg_support.c.template
```diff
@@ -0,0 +1,78 @@
...
+PyMODINIT_FUNC
+PyInit__@(module_name)_support(void)
+{
+  PyObject *m = NULL;
+  m = PyModule_Create(&@(spec.base_type.pkg_name)_@(module_name)__module);
+  if (m != NULL)
+  {
+    PyObject *pyconverter = NULL;
+    pyconverter = PyCapsule_New((void *)&@(spec.base_type.pkg_name)_@(module_name)__converter, NULL, NULL);
+    if(pyconverter != NULL)
```
jacquelinekay: Space between if and parentheses.

rosidl_generator_py/package.xml
```diff
@@ -12,6 +12,10 @@
   <buildtool_export_depend>rosidl_cmake</buildtool_export_depend>
 
   <exec_depend>rosidl_parser</exec_depend>
+  <build_depend>rosidl_generator_c</build_depend>
```
jacquelinekay: super nitpicky, but in most package.xml files the `exec_depends` and `build_depends` are grouped together, can you do that here?

jacquelinekay:
This language converter and the `Py_XDECREF` business is pretty fancy looking. I've never seen it before so I can't claim to know how it all works. I assume once you add the converter functions this will be testable?

rosidl_generator_py/resource/_msg_support.c.template
esteve: Fixed.

rosidl_generator_py/resource/_msg_support.c.template
esteve: Fixed.

rosidl_generator_py/package.xml
esteve: Fixed.

esteve:
@jacquelinekay I should have added a few more comments, but this PR is still in progress. This is a quick summary of how it all works,: `Py_INCREF` and `Py_(X)DECREF` is Python's way of implementing reference counting, `PyCapsule_New` wraps native C pointers and returns a reference, so no need to increment the count.

The linking issue I have is that the types generated from the `rosidl_generator_py` tests need to be linked against the C type support libraries. This worked when I wasn't building with `--isolated` because CMake would pick up the libraries from `rosidl_generator_c`, but that was wrong because they should be independent. This is what happens when importing the types:

```shell
esteve@osrf-esteve:~/Projects/ros2/ros2_ws/build_isolated/rosidl_generator_py$ python3 -c 'import _bool_support'
  File "<string>", line 1, in <module>
ImportError: /home/esteve/Projects/ros2/ros2_ws/build_isolated/rosidl_generator_py/_bool_support.cpython-34m.so: undefined symbol: rosidl_get_message_type_support__rosidl_generator_py__msg__Bool
```

Playing with `LD_PRELOAD` works around the issue, but that's not a solution.

esteve:
@jacquelinekay I've cleaned up the CMake files, the Python native extensions for interfaces are built and linked properly (e.g. `std_interfaces`), but the the test fixtures can't be linked properly. That is, this works:

```shell
esteve@osrf-esteve:~/Projects/ros2/ros2_ws$ src/ament/ament_tools/scripts/ament.py build --isolated --end-with std_msgs
...
esteve@osrf-esteve:~/Projects/ros2/ros2_ws$ ldd ./install_isolated/builtin_interfaces/lib/python3.4/site-packages/builtin_interfaces/msg/_time_support.cpython-34m.so
        linux-vdso.so.1 =>  (0x00007ffd2e998000)
        libpython3.4m.so.1.0 => /usr/lib/x86_64-linux-gnu/libpython3.4m.so.1.0 (0x00007f2e72987000)
        libbuiltin_interfaces__rosidl_typesupport_introspection_c.so => /home/esteve/Projects/ros2/ros2_ws/build_isolated/builtin_interfaces/libbuiltin_interfaces__rosidl_typesupport_introspection_c.so (0x00007f2e72785000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2e723c0000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f2e721a2000)
        libexpat.so.1 => /lib/x86_64-linux-gnu/libexpat.so.1 (0x00007f2e71f78000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f2e71d5f000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f2e71b5b000)
        libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x00007f2e71958000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f2e71652000)
        librosidl_typesupport_introspection_c.so => /home/esteve/Projects/ros2/ros2_ws/install_isolated/rosidl_typesupport_introspection_c/lib/librosidl_typesupport_introspection_c.so (0x00007f2e71450000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f2e731bf000)
```

However, this doesn't:

```shell
esteve@osrf-esteve:~/Projects/ros2/ros2_ws$ src/ament/ament_tools/scripts/ament.py build --build-tests --isolated --end-with std_msgs
...
Linking C shared library _float64_support.cpython-34m.so
/usr/bin/ld: cannot find -lrosidl_generator_py__rosidl_typesupport_introspection_c
collect2: error: ld returned 1 exit status
```

@dirk-thomas any idea how to proceed? I definitely still need to learn more about `ament`

jacquelinekay:
maybe you need to `ament_export_dependencies(rosidl_typesupport_introspection_c)` in `rosidl_generator_py_generate_interfaces.cmake`? or add `rosidl_typesupport_introspection_c` to line 162 of that file (https://github.com/ros2/rosidl/pull/73/files#diff-5b25e427fc24855bbe29fe86a0960284R162)?

esteve:
After https://github.com/ros2/rosidl/pull/71 got merged, this no longer builds, I'm working on a fix.

jacquelinekay:
I did a little work to get this to compile since @esteve was out this afternoon. It's restored to the last working state (link error on rosidl_typesupport_introspection_c). @wjwwood, I tried to follow the pattern in the middleware typesupport packages, but I'm not really sure that's appropriate here. Want to check it out?

5d0793e

jacquelinekay:
Maybe I did this wrong and I'm supposed to use rosidl_typesupport_introspection_c instead, I have no clue.

jacquelinekay:
I did fix the linking error though, in d8532ff.

wjwwood:
I don't know. For example, I'm not sure why this type support package needs to implement the `ROSIDL_GET_TYPE_SUPPORT_FUNCTION` macro or what the contents should be for the type support structure returned from the resulting `rosidl_generator_py_get_message__<pkg name>__msg__<msg name>` function.

I will need to understand what the pieces are and why they are there a little better before I can say one way or the other...

esteve:
@jacquelinekay thanks for taking a look at this! I'm reviewing your changes.

esteve:
@jacquelinekay the approach I took with this PR is that the extensions built for each type link against the type support libraries, instead of implementing new type support machinery.

Here's a brief summary of how this worked before the refactor (sorry I didn't add more comments in the source code):

- Given a `Message.msg` file in the `my_package` package:
  - Python structures are generated from the `.msg` file (`my_package/msg/_message.py`)
  - An accompanying Python extension (`my_package/msg/_message_support.cpython-34m.so`) is generated and compiled. Linked against the corresponding type support library (e.g. `libmy_package_interfaces__rosidl_typesupport_introspection_c.so` for `rosidl_typesupport_introspection`)
  - Type support and a function for converting from Python to C types are wrapped as `PyCapsule`s https://github.com/ros2/rosidl/blob/rclpy-conversion-functions/rosidl_generator_py/resource
https://github.com/ros2/rosidl/blob/96be3579413eb77ea73a597ff52818413f636ec0/rosidl_generator_py/resource/_msg_support.c.template#L49 and https://github.com/ros2/rosidl/blob/96be3579413eb77ea73a597ff52818413f636ec0/rosidl_generator_py/resource/_msg_support.c.template#L23
  - The `PyCapsule`s are added to the Python structure's metaclass (i.e. `_TYPE_SUPPORT` and `_CONVERTER`) https://github.com/ros2/rosidl/blob/96be3579413eb77ea73a597ff52818413f636ec0/rosidl_generator_py/resource/_msg.py.template#L9 and https://github.com/ros2/rosidl/blob/96be3579413eb77ea73a597ff52818413f636ec0/rosidl_generator_py/resource/_msg.py.template#L11

Then in `rclpy`:
- When a Python structure is passed to an `rclpy` function (e.g. `create_publisher` https://github.com/ros2/rclpy/blob/initial/rclpy/_rclpy.c#L38):
  - The `_rclpy` extension retrieves the type support `PyCapsule` from the Python structure (via `Message.__class__,_TYPE_SUPPORT`) https://github.com/ros2/rclpy/blob/initial/rclpy/_rclpy.c#L53
  - The `PyCapsule` is converted back to the proper `rosidl_message_type_support_t *` structure via `PyCapsule_Pointer` https://github.com/ros2/rclpy/blob/initial/rclpy/_rclpy.c#L55
  - Now that we have the "real" type support structure, we can pass it to `rcl_publisher_init` https://github.com/ros2/rclpy/blob/initial/rclpy/_rclpy.c#L60

BRRRRRRRAAAWRWRRRMRMRM!!!

jacquelinekay:
OK, so the problem that the recent rosidl_generator_c change introduced is that messages generated by `rosidl_generator_c` expect a header, `rosidl_generator_c/message_type_support.h`, to be provided by the underlying typesupport package. I went down the OTHER rabbit hole by provided generator_py specific typesupport machinery.

Perhaps the solution is for `rosidl_generator_py` to provide the `rosidl_generator_c/message_type_support.h` header, which simply delegates to the underlying type support by including a header somehow?

jacquelinekay:
Ok, I've pushed changes from the pair programming session we did (actually, trio session at some point since William joined us).

We're going to put the test fixtures in a separate package so that we can depend on rmw_implementation.

We made some changes to build a separate library target for each message and typesupport implementation pairing, but it could use a little polishing (e.g. right now the typesupport implementation list is hardcoded).

I've added hooks for a `typesupport_impl` argument to the Python generator, and I added an example of using it in the C template. I'm not sure about where exactly to add it for the Python module to use it correctly, however, so I'll leave the rest of that up to @esteve.

rosidl_typesupport_opensplice_c library does not get generated for interface packages, even on the `add_opensplice_typesupport_c` branch of rmw_opensplice. I or @wjwwood can look more into this issue.

Also, I rebased, so you may need to do a `git reset --hard origin/rclpy-conversion-functions` to avoid merge conflicts.

esteve:
@jacquelinekay thanks!

esteve:
I've rebased against master and I'm going to remove the extra files that were added during the refactor.

esteve:
I've moved the tests to its own separate package as we agreed. The mechanism for picking a type support Python extension is a bit brittle, just looks up the first available one.

A weird thing is that the testsuite does not run when run from `ament`, it's reported as a "BAD COMMAND", but when run manually using the same incantation that `ament` prints, it passes:

```shell
esteve@osrf-esteve:~/Projects/ros2/ros2_ws$ /usr/bin/python3 "-u" "/home/esteve/Projects/ros2/ros2_ws/install_isolated/ament_cmake_test/share/ament_cmake_test/cmake/run_test.py" "/home/esteve/Projects/ros2/ros2_ws/build_isolated/rosidl_generator_py_tests/test_results/rosidl_generator_py_tests/test_interfaces_py.xunit.xml" "--output-file" "/home/esteve/Projects/ros2/ros2_ws/build_isolated/rosidl_generator_py_tests/ament_cmake_nose/test_interfaces_py.txt" "--append-env" "PYTHONPATH=/home/esteve/Projects/ros2/ros2_ws/build_isolated/rosidl_generator_py_tests/rosidl_generator_py:/home/esteve/Projects/ros2/ros2_ws/src/ros2/rosidl/rosidl_generator_py_tests" "--command" "/usr/bin/nosetests3" "/home/esteve/Projects/ros2/ros2_ws/src/ros2/rosidl/rosidl_generator_py_tests/test/test_interfaces.py" "--with-xunit" "--xunit-file=/home/esteve/Projects/ros2/ros2_ws/build_isolated/rosidl_generator_py_tests/test_results/rosidl_generator_py_tests/test_interfaces_py.xunit.xml"
-- run_test.py: extra environment variables to append:
 - PYTHONPATH=/home/esteve/Projects/ros2/ros2_ws/build_isolated/rosidl_generator_py_tests/rosidl_generator_py:/home/esteve/Projects/ros2/ros2_ws/src/ros2/rosidl/rosidl_generator_py_tests
-- run_test.py: invoking following command in '/home/esteve/Projects/ros2/ros2_ws':
 - /usr/bin/nosetests3 /home/esteve/Projects/ros2/ros2_ws/src/ros2/rosidl/rosidl_generator_py_tests/test/test_interfaces.py --with-xunit --xunit-file=/home/esteve/Projects/ros2/ros2_ws/build_isolated/rosidl_generator_py_tests/test_results/rosidl_generator_py_tests/test_interfaces_py.xunit.xml
......
----------------------------------------------------------------------
Ran 6 tests in 0.001s

OK
-- run_test.py: return code 0
-- run_test.py: verify result file '/home/esteve/Projects/ros2/ros2_ws/build_isolated/rosidl_generator_py_tests/test_results/rosidl_generator_py_tests/test_interfaces_py.xunit.xml'
```

dirk-thomas:
Can you please explain why the tests should be in a separate package in this case? What is the use case which doesn't work within the same package?

esteve:
@dirk-thomas it's because the Python generator also depends on the artifacts of the C generator, which proved to be very difficult to get working (we couldn't), so we all three (@jacquelinekay @wjwwood and me) decided to put the testsuite in a separate package.

wjwwood:
I think it was a complication of the fact that this is the first type support package that is middleware agnostic but also needs to generate a shared library for each middleware implementation available. I'm not sure it is impossible to put them in the same package, but I think that might be much easier.

dirk-thomas:
Having separate packages results in extra effort later (during packaging). Therefore I suggest this should be ticketed to be merged later again.

Also all test-only packages start with `test_`. It would be good if the new test package would follow that pattern.

wjwwood:
+1 for the `test_` prefix.

dirk-thomas:
I might be missing something but currently the CMake code of the test package seems to not do anything special. Simply triggering message generation and then running a single Python nosetest. I don't understand which part of that doesn't work within the same package.

esteve:
I just renamed `rosidl_generator_py_tests` to `test_rosidl_generator_py`.

@dirk-thomas we couldn't find a way for the test extensions to be linked against the corresponding type support library. The rest of the message packages (e.g. `std_msgs`, `example_interfaces`) were fine, the problem was only in the test messages if they were in the same package as the Python generator.

dirk-thomas:
But that step happens within the type support package - and not in the new test package. If it doesn't work correctly within the package that only means something is incomplete / wrong with in the CMake code.

Other generators are able to use their own message generation within the same package and have some extra CMake code to setup everything accordingly: e.g. https://github.com/ros2/rosidl/blob/99d6e1fda2ecc0d721f0af197ef5515b3ae44777/rosidl_generator_cpp/CMakeLists.txt#L43-L54

Based on your feedback I am convinced that the problem can be resolved to perform the tests within the same package.

esteve:
@dirk-thomas I'm not saying it's impossible, I'm just saying we couldn't figure out how to do it :smile: In this case, it's not only the Python code that's generated, but also the C code (through `rosidl_generator_c`) and the libraries that it depends on. The current Python generator that's in master uses the approach you point out https://github.com/ros2/rosidl/blob/master/rosidl_generator_py/CMakeLists.txt#L39 but things got more complicated when the Python extensions were added.

dirk-thomas:
I have created #80 to merge the content of the new test package back into `rosidl_generator_py` and it seems to pass the tests. Is this what you wanted to achieve or am I missing something?

esteve:
@dirk-thomas that worked, thanks! That's what CMakeLists.txt looked originally, maybe the changes we made to the CMake files just fixed the issue as a side effect. Anyway, thanks for looking into this :+1:

esteve:
CI after merging in the tests:

http://ci.ros2.org/job/ci_linux/901/
http://ci.ros2.org/job/ci_osx/753/
http://ci.ros2.org/job/ci_windows/978/

esteve:
@dirk-thomas I had to revert the reintegration of the test package locally (I haven't pushed it) so I could keep working on `rclpy`. This branch can no longer build because of this error (same in the CI job):

```
  /home/rosbuild/ci_scripts/ws/install/rosidl_generator_py/share/rosidl_generator_py/cmake/rosidl_generator_pyConfig.cmake:30 (include)
  /home/rosbuild/ci_scripts/ws/install/rosidl_default_generators/share/rosidl_default_generators/cmake/ament_cmake_export_dependencies-extras.cmake:21 (find_package)
  /home/rosbuild/ci_scripts/ws/install/rosidl_default_generators/share/rosidl_default_generators/cmake/rosidl_default_generatorsConfig.cmake:30 (include)
  CMakeLists.txt:10 (find_package)


CMake Error at /home/rosbuild/ci_scripts/ws/install/rosidl_generator_py/share/rosidl_generator_py/cmake/ament_cmake_export_libraries-extras.cmake:42 (message):
  Package 'rosidl_generator_py' exports the library
  'rosidl_generator_py__rosidl_typesupport_introspection_c' which couldn't be
  found
  /home/rosbuild/ci_scripts/ws/install/rosidl_generator_py/share/rosidl_generator_py/cmake/rosidl_generator_pyConfig.cmake:30 (include)
  /home/rosbuild/ci_scripts/ws/install/rosidl_default_generators/share/rosidl_default_generators/cmake/ament_cmake_export_dependencies-extras.cmake:21 (find_package)
  /home/rosbuild/ci_scripts/ws/install/rosidl_default_generators/share/rosidl_default_generators/cmake/rosidl_default_generatorsConfig.cmake:30 (include)
  CMakeLists.txt:10 (find_package)
```

esteve:
CI after spinning off the test into a separate package:

http://ci.ros2.org/job/ci_linux/905/
http://ci.ros2.org/job/ci_osx/757/
http://ci.ros2.org/job/ci_windows/985/

esteve:
This is the CI with the tests integrated in the same package:

http://ci.ros2.org/job/ci_linux/901/

esteve:
python3-dev needs to be installed on the Windows and OSX slaves, working on a fix now.

esteve:
CI for OSX that should fix the python3-dev conflict:

http://ci.ros2.org/job/ci_osx/762/

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -12,9 +12,175 @@
...
+      python${_python_version}m
+      python${_python_version}u
+      python${_python_version}
+      PATHS
+      ${_library_paths}
+      NO_SYSTEM_ENVIRONMENT_PATH
+    )
+  endif()
+
+  set(PYTHON_LIBRARIES "${PYTHON_LIBRARY}")
```
dirk-thomas: This block should print a similar message as the non-OSX branch and also state where the information was coming from.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -12,9 +12,175 @@
...
+      COMMAND
+      "${PYTHON_CONFIG_EXECUTABLE}"
+      "--ldflags"
+      OUTPUT_VARIABLE _output
+      RESULT_VARIABLE _result
+      OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+    if(NOT _result EQUAL 0)
+      message(FATAL_ERROR
+        "execute_process(${PYTHON_CONFIG_EXECUTABLE} --libs returned "
```
dirk-thomas: The error message doesn't match the actual command.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -12,9 +12,175 @@
...
+      COMMAND
+      "${PYTHON_CONFIG_EXECUTABLE}"
+      "--includes"
+      OUTPUT_VARIABLE _output
+      RESULT_VARIABLE _result
+      OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+    if(NOT _result EQUAL 0)
+      message(FATAL_ERROR
+        "execute_process(${PYTHON_CONFIG_EXECUTABLE} --includes returned "
```
dirk-thomas: The closing parenthesis of `execute_process(`  is missing.

Same below

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
esteve: Fixed.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -12,9 +12,175 @@
...
+      RESULT_VARIABLE _result
+      OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+    if(NOT _result EQUAL 0)
+      message(FATAL_ERROR
+        "execute_process(${PYTHON_CONFIG_EXECUTABLE} --libs returned "
+        "error code ${_result}")
+    endif()
+
+    string(REPLACE " " ";" _output_list ${_output})
```
dirk-thomas: You might want to quote the input string.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
esteve: Fixed.

dirk-thomas:
Please don't just split the packages again. The previous case had a pretty simple fix. Very likely the same is true now again. So please keep investigating it and keep the tests in a single package.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
esteve: I added two messages informing about the Python libraries and directories.

esteve:
@dirk-thomas yeah, I'd rather not split them, but I couldn't find a solution. Let's have a look at it together.

esteve:
@dirk-thomas thanks for the fix.

dirk-thomas:
Fixed the exporting of the typesupport introspection c when SKIP_INSTALL is used: 60bee78600f460f09ba63062817cf151deb2d7a5

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
esteve: Done.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
dirk-thomas: The first message is only displayed once while the second one is shown always.

esteve: Fixed.

esteve:
CI:

http://ci.ros2.org/job/ci_linux/910/
http://ci.ros2.org/job/ci_osx/769/
http://ci.ros2.org/job/ci_windows/998/

esteve:
New CI job for OSX:

http://ci.ros2.org/job/ci_osx/772/

esteve:
OSX finally passes, looking into Windows now.

esteve:
I finally got Windows passing locally, new CI jobs:

http://ci.ros2.org/job/ci_linux/938/
http://ci.ros2.org/job/ci_osx/787/
http://ci.ros2.org/job/ci_windows/1032/

esteve:
Both OSX and Windows slaves went offline, triggering again:

http://ci.ros2.org/job/ci_osx/791/
http://ci.ros2.org/job/ci_windows/1036/

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
```diff
@@ -145,7 +145,10 @@ macro(rosidl_generate_interfaces target)
   set(rosidl_generate_interfaces_IDL_FILES ${_idl_files})
   set(rosidl_generate_interfaces_DEPENDENCY_PACKAGE_NAMES ${_recursive_dependencies})
   set(rosidl_generate_interfaces_SKIP_INSTALL ${_ARG_SKIP_INSTALL})
-  ament_execute_extensions("rosidl_generate_interfaces")
+  if(_ARG_TRACE)
+    set(_trace_args "TRACE")
+  endif()
+  ament_execute_extensions("rosidl_generate_interfaces" ${_trace_args})
```
dirk-thomas: The function `ament_execute_extensions` does not accept any other arguments. So the changes from this file should be removed.

rosidl_generator_py/CMakeLists.txt
```diff
@@ -46,10 +47,12 @@ if(AMENT_ENABLE_TESTING)
 
   rosidl_generate_interfaces(${PROJECT_NAME} ${message_files}
     SKIP_INSTALL
+    TRACE
   )
 
   ament_add_nose_test(test_interfaces_py "test/test_interfaces.py"
-    APPEND_ENV PYTHONPATH=${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py:${CMAKE_CURRENT_SOURCE_DIR}
+    APPEND_ENV PYTHONPATH="${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py:${CMAKE_CURRENT_SOURCE_DIR}"
+    APPEND_LIBRARY_DIRS "${CMAKE_CURRENT_BINARY_DIR}/Release"
```
dirk-thomas: What is this library dir needed for? If it is only necessary for Windows it should also only be added for that platform? And if it points to the folder on Windows which contains the compiled target this needs to handle the currently selected configuration rather then hardcoded `Release`.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -12,20 +12,223 @@
...
+find_package(rosidl_generator_c REQUIRED)
+find_package(rmw_implementation_cmake REQUIRED)
+find_package(rmw REQUIRED)
+
+# NOTE(esteve): required for CMake-2.8 in Ubuntu 14.04
+set(Python_ADDITIONAL_VERSIONS 3.4)
+find_package(PythonInterp 3.4 REQUIRED)
+
+find_program(PYTHON_CONFIG_EXECUTABLE NAMES "python3-config")
+if(NOT DEFINED PYTHON_CONFIG_EXECUTABLE)
```
dirk-thomas: This check is wrong since the variable is always defined even when being set to `<VAR>-NOTFOUND`.

rosidl_generator_py/package.xml
```diff
@@ -10,13 +10,27 @@
 
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
   <buildtool_export_depend>rosidl_cmake</buildtool_export_depend>
+  <buildtool_export_depend>rosidl_generator_c</buildtool_export_depend>
 
+  <build_depend>rosidl_generator_c</build_depend>
+  <!-- export??? -->
```
dirk-thomas: What about this comment / question?

rosidl_generator_py/resource/_msg.py.template
```diff
@@ -1,10 +1,30 @@
...
+if rclpy_implementation == 'opensplice_static':
+    from @(spec.base_type.pkg_name).@(subfolder)._@(module_name)_support__rosidl_typesupport_opensplice_c import convert_from_py
+    from @(spec.base_type.pkg_name).@(subfolder)._@(module_name)_support__rosidl_typesupport_opensplice_c import convert_to_py
+    from @(spec.base_type.pkg_name).@(subfolder)._@(module_name)_support__rosidl_typesupport_opensplice_c import type_support
+elif rclpy_implementation == 'connext_static':
+    from @(spec.base_type.pkg_name).@(subfolder)._@(module_name)_support__rosidl_typesupport_connext_c import convert_from_py
+    from @(spec.base_type.pkg_name).@(subfolder)._@(module_name)_support__rosidl_typesupport_connext_c import convert_to_py
+    from @(spec.base_type.pkg_name).@(subfolder)._@(module_name)_support__rosidl_typesupport_connext_c import type_support
+else:
+    raise RuntimeError('Unknown implementation: {0}'.format(rclpy_implementation))
```
dirk-thomas: This block means that the generate Python code for messages can not be uses as-is as standalone data type since it depends on a type support being present.

rosidl_generator_py/resource/_msg_support.c.template
```diff
@@ -0,0 +1,248 @@
...
+        }
+      }
+    } else
+    {
+      Py_XDECREF(m);
+      m = NULL;
+    }
+  }
+  return m;
+}
```
dirk-thomas: The file doesn't follow the code style.

rosidl_generator_py/rosidl_generator_py/__init__.py
```diff
@@ -65,14 +72,20 @@ def generate_py(generator_arguments_file):
...
-            generated_file = os.path.join(
-                args['output_dir'], subfolder, generated_filename % module_name)
-            expand_template(
-                template_file, data, generated_file,
-                minimum_timestamp=latest_target_timestamp)
+        for template_file, generated_filenames in mapping.items():
+            for generated_filename in generated_filenames:
+                data = {
+                    'module_name': module_name, 'package_name': args['package_name'],
+                    'spec': spec, 'subfolder': subfolder,
```
dirk-thomas: Please wrap the lines after each comma since it makes it very hard to see what keys are actually in the dict otherwise.

rosidl_generator_py/resource/_msg.py.template
```diff
@@ -1,10 +1,30 @@
...
+
 
 class Metaclass(type):
     """Metaclass of message '@(spec.base_type.type)'."""
 
+    _CONVERT_FROM_PY = convert_from_py
+
+    _CONVERT_TO_PY = convert_to_py
+
+    _TYPE_SUPPORT = type_support
```
dirk-thomas: Since it is not obvious what these variables are used for it would be good to add a comment.

rosidl_generator_py/resource/_msg_support.c.template
```diff
@@ -0,0 +1,248 @@
...
+@{have_not_included_string = False}@
+#include <rosidl_generator_c/string.h>
+#include <rosidl_generator_c/string_functions.h>
+
+@[end if]@
+@[end for]@
+
+@{
+msg_typename = '%s__%s__%s' % (spec.base_type.pkg_name, subfolder, spec.base_type.type)
+array_typename = '%s__Array' % msg_typename
```
dirk-thomas: This variable seems to be not used anywhere?

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
wjwwood: 4a01ed4

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
wjwwood: 132d143

rosidl_generator_py/package.xml
wjwwood: In my opinion it doesn't need to export `rmw`. I did, however, clean up the dependencies 5fcb215.

wjwwood:
@esteve I refactored the way the generated python code imports rclpy and then made it optional to import the correct type support based on that.

This allows the tests to run again in this package, but I stopped short of reenabling them because there is still a test failure. I think we might also need changes in rclpy to notice when messages are imported incorrectly (basically `Metaclass._CONVERT*` and `Metaclass._TYPE_SUPPORT` are all `None`).

esteve:
@wjwwood thanks!

rosidl_generator_py/resource/_msg_support.c.template
esteve: Removed.

wjwwood:
Tests are re-enabled and you can now import the generated Python code without `rcl` or `rosidl_generator_py` on your `PYTHON_PATH`.

esteve:
CI jobs with the latest changes in all the branches:

http://ci.ros2.org/job/ci_linux/986/
http://ci.ros2.org/job/ci_osx/818/
http://ci.ros2.org/job/ci_windows/1069/

esteve:
rclpy test passing on all CI jobs:

http://ci.ros2.org/job/ci_linux/993/
http://ci.ros2.org/job/ci_linux/993/testReport/%28root%29/test_talker_listener/

http://ci.ros2.org/job/ci_osx/822/
http://ci.ros2.org/job/ci_osx/822/testReport/%28root%29/test_talker_listener/

http://ci.ros2.org/job/ci_windows/1072/
http://ci.ros2.org/job/ci_windows/1072/testReport/%28root%29/test_talker_listener/test_rclpy_talker_listener/

dirk-thomas:
The tests look good now. With the extra time until the release we should aim to also address the warning across all platforms.

rosidl_generator_py/rosidl_generator_py/import_type_support_impl.py
```diff
@@ -0,0 +1,58 @@
...
+
+    :param pkg_name str: name of the package which contains the rosidl
+    :param subfolder str: name of the rosidl containing folder, e.g. `msg`, `srv`, etc.
+    :param rosidl_name str: name of the rosidl
+    :param rmw_implementation str: name of the rmw implementation
+    :returns: the type support Python module for this specific rosidl and rmw implementation pair
+    """
+    if rmw_implementation not in type_support_map.keys():
+        raise UnsupportedTypeSupport(
+            "No supported type support for '{0}'".format(rmw_implementation)
```
dirk-thomas: Instead of embedding the rmw impl. into the string the exception could store the rmw implementation separately and still provide the same string representation.

rosidl_generator_py/rosidl_generator_py/import_type_support_impl.py
```diff
@@ -0,0 +1,58 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import importlib
+
+type_support_map = {
+    'opensplice_static': 'rosidl_typesupport_opensplice_c',
+    'connext_static': 'rosidl_typesupport_connext_c',
```
dirk-thomas: If this map is supposed to grow order alphabetically.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -0,0 +1,207 @@
...
+        type_support_impl_by_filename.keys(),
+    }
+
+    mapping_srvs = {
+        os.path.join(template_dir, '_srv.py.template'): ['_%s.py'],
+    }
+
+    for template_file in mapping_msgs.keys():
+        assert os.path.exists(template_file), \
+            'Messages template file %s not found' % template_file
```
dirk-thomas: Quotes around the template file.

Same below.

rosidl_generator_py/resource/_msg_support.entry_point.c.template
```diff
@@ -0,0 +1,95 @@
...
+        }
+      }
+    } else
+    {
+      Py_XDECREF(m);
+      m = NULL;
+    }
+  }
+  return m;
+}
```
dirk-thomas: Style.

rosidl_generator_py/package.xml
```diff
@@ -10,13 +10,28 @@
 
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
   <buildtool_export_depend>rosidl_cmake</buildtool_export_depend>
+  <buildtool_export_depend>rosidl_generator_c</buildtool_export_depend>
 
+  <build_depend>rmw</build_depend>
+  <build_depend>rmw_implementation</build_depend>
+  <build_depend>rosidl_generator_c</build_depend>
```
dirk-thomas: Should build dependencies be listed before any run dependencies like export dependencies?

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -88,33 +300,159 @@ rosidl_write_generator_arguments(
...
+
+  # foreach(_typesupport_impl ${_typesupport_impls})
+    list(FIND _generated_msg_c_ts_files ${_generated_msg_c_ts_file} _file_index)
+    list(GET _type_support_by_generated_msg_c_files ${_file_index} _typesupport_impl)
+    find_package(${_typesupport_impl} REQUIRED)
+    set(_generated_msg_c_common_file "${_full_folder}/${_base_msg_name}.c")
+
+    set(_pyext_suffix "__pyext")
+
+# Change the following code so that each file is compiled independently
```
dirk-thomas: Should this be a `TODO`?

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -88,33 +300,159 @@ rosidl_write_generator_arguments(
...
 file(WRITE "${_output_path}/__init__.py" "")
+
+set(_generated_extension_files "")
+set(_extension_dependencies "")
+set(_target_suffix "__py")
+
+if(WIN32)
+  set(PYTHON_EXTENSION_EXTENSION ".pyd")
+else()
+# Also use .so for OSX, not dylib
```
dirk-thomas: Comment should have same indentation as block.

rosidl_generator_py/CMakeLists.txt
```diff
@@ -13,6 +15,7 @@ if(AMENT_ENABLE_TESTING)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
 
+  # TODO(esteve): disable tests for now until we resolve the rclpy circular dependency
```
dirk-thomas: Aren't the tests enabled right now? Is this comment obsolete?

rosidl_generator_py/package.xml
esteve: I followed the same order as other packages, e.g. https://github.com/ros2/rmw_connext/blob/master/rmw_connext_cpp/package.xml Let me know what should be the correct order.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
esteve: Added.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
esteve: Fixed.

rosidl_generator_py/CMakeLists.txt
esteve: Removed.

rosidl_generator_py/rosidl_generator_py/import_type_support_impl.py
esteve: Done.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
esteve: Added.

rosidl_generator_py/package.xml
dirk-thomas: When reading a manifest I find it most reasonable to first see all build dependencies, then all run dependencies and at the end the test dependencies. Within build and run its probably good to list the tool dependencies first.

rosidl_generator_py/rosidl_generator_py/import_type_support_impl.py
```diff
@@ -0,0 +1,65 @@
...
+
+    :param pkg_name str: name of the package which contains the rosidl
+    :param subfolder str: name of the rosidl containing folder, e.g. `msg`, `srv`, etc.
+    :param rosidl_name str: name of the rosidl
+    :param rmw_implementation str: name of the rmw implementation
+    :returns: the type support Python module for this specific rosidl and rmw implementation pair
+    """
+    if rmw_implementation not in type_support_map.keys():
+        raise UnsupportedTypeSupport(
+            "No supported type support for '{0}'".format(rmw_implementation),
```
dirk-thomas: I would suggest that the exception class should contain the message and also do the format operation.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -0,0 +1,207 @@
...
+        type_support_impl_by_filename.keys(),
+    }
+
+    mapping_srvs = {
+        os.path.join(template_dir, '_srv.py.template'): ['_%s.py'],
+    }
+
+    for template_file in mapping_msgs.keys():
+        assert os.path.exists(template_file), \
+            'Messages template file "%s" not found' % template_file
```
dirk-thomas: Other code uses single quotes for the user visible part and then double quotes around the string.

rosidl_generator_py/rosidl_generator_py/import_type_support_impl.py
esteve: Done.

rosidl_generator_py/rosidl_generator_py/import_type_support_impl.py
esteve: Done.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
esteve: I ended up copying the same error as in the c and c++ generators.

rosidl_generator_py/package.xml
esteve: I've moved the build dependencies to the top, though I've always seen the buildtool_depend first and then the other dependencies. Does this look ok? https://github.com/ros2/rosidl/commit/b9cf5eb36dab4075951f7f4a8bc3575db9079dc7

rosidl_generator_py/CMakeLists.txt
esteve: Done.

rosidl_generator_py/package.xml
dirk-thomas: The change doesn't look like whats described above. I suggested to have:
* build before run before test
* withing these groups tool before non-tools

Example list:
* buildtool
* build
* buildtool_export
* build_export
* exec
* test

This order basically resembled the workflow:
* install build tools for build platform
* install build dep for target platform
* export build tools for build platform
* export build dep for target platform
* export run dep for target platform
* install test dep if necesary

rosidl_generator_py/resource/_msg_support.entry_point.c.template
esteve: Done.

esteve:
CI:

http://ci.ros2.org/job/ci_linux/1003/
http://ci.ros2.org/job/ci_osx/834/
http://ci.ros2.org/job/ci_windows/1083/

python_cmake_module/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+cmake_minimum_required(VERSION 2.8.3)
+
+project(python_cmake_module)
+
+find_package(ament_cmake REQUIRED)
+
+list(INSERT CMAKE_MODULE_PATH 0 "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
+find_package(PythonExtra MODULE)
```
dirk-thomas: Why are those two lines needed?

python_cmake_module/package.xml
```diff
@@ -0,0 +1,20 @@
...
+  <name>python_cmake_module</name>                                                                       
+  <version>0.0.0</version>                                                                                   
+  <description>Provide CMake module with extra functionality for Python.</description>
+
+  <maintainer email="esteve@osrfoundation.org">Esteve Fernandez</maintainer>
+  <license>Apache License 2.0</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <buildtool_export_depend>ament_cmake</buildtool_export_depend>
```
dirk-thomas: Why is this needed?

rosidl_generator_py/package.xml
```diff
@@ -6,17 +6,37 @@
   <maintainer email="esteve@osrfoundation.org">Esteve Fernandez</maintainer>
   <license>Apache License 2.0</license>
 
+  <build_depend>python_cmake_module</build_depend>
```
dirk-thomas: It doesn't seem to be used for building this package?

rosidl_generator_py/CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ project(rosidl_generator_py)
 
 find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_python REQUIRED)
+find_package(rosidl_cmake REQUIRED)
+find_package(rosidl_generator_c REQUIRED)
```
dirk-thomas: Why is this needed at build time of this package? Maybe only for the tests?

rosidl_generator_py/package.xml
```diff
@@ -6,17 +6,37 @@
...
+  <build_depend>rmw</build_depend>
+  <build_depend>rmw_implementation</build_depend>
+  <build_depend>rosidl_generator_c</build_depend>
```
dirk-thomas: Same for all of these.

python_cmake_module/CMakeLists.txt
esteve: Isn't this needed so it can be found by CMake? `opensplice_cmake_module` does something similar.

rosidl_generator_py/package.xml
```diff
@@ -6,17 +6,37 @@
...
+
+  <build_export_depend>python_cmake_module</build_export_depend>
```
dirk-thomas: Since it is also a `buildtool_depend` and targeting the build platform this is not needed.

rosidl_generator_py/package.xml
```diff
@@ -6,17 +6,37 @@
...
+  <build_export_depend>rmw_implementation</build_export_depend>
+
   <buildtool_depend>ament_cmake</buildtool_depend>
+  <buildtool_depend>python_cmake_module</buildtool_depend>
 
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
   <buildtool_export_depend>rosidl_cmake</buildtool_export_depend>
+  <buildtool_export_depend>rosidl_generator_c</buildtool_export_depend>
 
+  <exec_depend>python_cmake_module</exec_depend>
```
dirk-thomas: Since it is also a `buildtool_depend` and targeting the build platform this is not needed.

python_cmake_module/package.xml
esteve: Removed.

rosidl_generator_py/package.xml
esteve: Removed.

python_cmake_module/CMakeLists.txt
dirk-thomas: But this CMake file seem to not use the result anywhere to why find it in the first place?

rosidl_generator_py/CMakeLists.txt
esteve: Yes, for the tests. Should I move these inside the AMENT_ENABLE_TESTING block?

dirk-thomas: Yes, test dependencies are only available when testing is enabled therefore can only be used within the conditional block. Then these are also no build dependencies anymore.

python_cmake_module/CMakeLists.txt
esteve: I thought it was needed so other packages would find `python_cmake_module`. Removed.

dirk-thomas: No, that is what the `python_cmake_module-extras.cmake` file is for. That gets included when this package is `find_package`d.

rosidl_generator_py/CMakeLists.txt
esteve: Understood. Just moved the lines and removed them from the build_dependencies.

esteve:
Updated CI:

http://ci.ros2.org/job/ci_linux/1005/
http://ci.ros2.org/job/ci_osx/836/

python_cmake_module/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+cmake_minimum_required(VERSION 2.8.3)
+
+project(python_cmake_module)
```
dirk-thomas: Since it does't require a compiler this can be `project(... NONE)`.

python_cmake_module/package.xml
```diff
@@ -0,0 +1,18 @@
...
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
+
+  <export>
+    <build_type>ament_cmake</build_type>
+  </export>
+</package>
```
dirk-thomas: This package needs to buildtool export depend on whatever rosdep key provides the necessary Python headers.

python_cmake_module/CMakeLists.txt
esteve: I had to remove the `NONE` or the `PythonLibs` package wouldn't work.

dirk-thomas:
The dependencies specified in the manifest are still very different from what is being used in CMake for the Python generator package. I will make a PR for them since it is difficult to iterate on this diff line by line.

esteve:
@dirk-thomas thanks!

dirk-thomas:
@esteve Should we abort http://ci.ros2.org/job/ci_windows/1083/ since it will be superseded anyway soon?

esteve:
@dirk-thomas yeah, good idea, I just aborted it.

dirk-thomas:
http://ci.ros2.org/job/ci_linux/1008/
http://ci.ros2.org/job/ci_osx/840/
http://ci.ros2.org/job/ci_windows/1085/

esteve:
OSX:

http://ci.ros2.org/job/ci_osx/

esteve:
Linux and Windows with the rclpy repo enabled:

http://ci.ros2.org/job/ci_linux/1009/
http://ci.ros2.org/job/ci_windows/1085/

rosidl_generator_py/resource/_msg_support.c.template
```diff
@@ -0,0 +1,160 @@
...
+  char * class_name = (char *)PyUnicode_1BYTE_DATA(
+    PyObject_GetAttrString(PyObject_GetAttrString(_pymsg, "__class__"), "__name__"));
+  char * module_name = (char *)PyUnicode_1BYTE_DATA(
+    PyObject_GetAttrString(PyObject_GetAttrString(_pymsg, "__class__"), "__module__"));
+
+  snprintf(full_classname_dest, @(len(full_classname) + 1), "%s.%s", class_name, module_name);
+
+  assert(strncmp(
+    "@(spec.base_type.pkg_name).@(subfolder)._@(module_name).@(spec.base_type.type)",
+    full_classname_dest, @(len(full_classname))));
```
dirk-thomas: I assume this is trying to check that the type name matches. But `strncmp` will return `0` if the strings are equal and `assert` wants `true`?

Since the check only looks at `len` (and not at `len +1`) it would also fail if `full_classname_dest` is just a prefix of the constructed string.

dirk-thomas: I created a PR to fix the broken assertion: #159

PR #74, opened 2016-01-30
Workaround constants

esteve:
This PR adds a work around a bug initializing string constants.

This also fixes headers not being included if the message didn't define any fields.

### DIFF
--- a/rosidl_generator_c/CMakeLists.txt
+++ b/rosidl_generator_c/CMakeLists.txt
@@ -59,6 +59,7 @@ if(AMENT_ENABLE_TESTING)
     "msg/Telegram1.msg"
     "msg/Telegram2.msg"
     "msg/Wire.msg"
+    "msg/Constants.msg"
   )
 
   include(cmake/register_c.cmake)
--- a/rosidl_generator_c/resource/msg__struct.h.template
+++ b/rosidl_generator_c/resource/msg__struct.h.template
@@ -55,16 +55,6 @@ for field in spec.fields:
             [])
         field_names.append(field.name)
 }@
-@[if includes]@
-// include message dependencies
-@[for header_file, field_names in includes.items()]@
-@[for field_name in field_names]@
-// @(field_name)
-@[end for]@
-#include "@(header_file)"
-@[end for]@
-
-@[end if]@
 @
 @#######################################################################
 @# constants defined in the message
@@ -87,6 +77,16 @@ for constant in spec.constants:
             primitive_value_to_c(constant.type, constant.value),
         ))
 }@
+@[if includes]@
+// include message dependencies
+@[for header_file, field_names in includes.items()]@
+@[for field_name in field_names]@
+// @(field_name)
+@[end for]@
+#include "@(header_file)"
+@[end for]@
+
+@[end if]@
 @[if constants]@
 // constants defined in the message
 @[for constant_type, constant_name, key, value in constants]@
@@ -96,6 +96,11 @@ enum
 {
   @(key) = @(value)
 };
+@[elif key.split()[0] == 'rosidl_generator_c__String']@
+@{
+name = key.split()[1]
+}@
+static const char * const @(name) = @(value);
 @[else]@
 static const @(key) = @(value);
 @[end if]@
--- a/rosidl_generator_c/msg/Constants.msg
+++ b/rosidl_generator_c/msg/Constants.msg
@@ -1,0 +1,3 @@
+int32 X=123
+int32 Y=-123
+string FOO=foo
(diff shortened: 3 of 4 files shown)


esteve:
CI:

http://ci.ros2.org/job/ci_linux/874/
http://ci.ros2.org/job/ci_osx/732/
http://ci.ros2.org/job/ci_windows/956/

jacquelinekay:
It looks to me that even though an IDL file is added for the Constants message, it never gets tested.

esteve:
@jacquelinekay yeah it's not tested, this just ensures that it can be parsed and compiled, without this change string constants fail to compile. Also messages that don't define fields fail to have the proper includes added, so they fail to compile too.

@ernestmc are you still working on adding tests to the C generator? I can work on adding tests for constants or if you prefer to take over, that's ok too.

jacquelinekay:
Can you do `static const char * @(name) = @(value) to assign the string constant on line 109 (https://github.com/ros2/rosidl/pull/74/files#diff-d01267ca97e5f8e7bb2f3faeef43018fR109)?

Results in generated code like this:

static const char * rosidl_generator_c__msg__Constants__FOO = "foo";

jacquelinekay:
Don't you need to add `msg/Constants.msg` to the CMakeLists.txt in order for it to get parsed?

jacquelinekay:
ok, I implemented the thing I suggested in 3a5fe380e63997bde54b2f31edb349b75587a935, feel free to check it out if you think it's a good idea

esteve:
@jacquelinekay good catch, I had this tested with #73 , but didn't update `rosidl_generator_c`, just updated this PR to add `msg/Constants.msg`

Using `const char *` would defeat the goals of `rosidl_generator_c__String`. https://github.com/ros2/rosidl/commit/3a5fe380e63997bde54b2f31edb349b75587a935 still generates `static const rosidl_generator_c__String rosidl_generator_c__msg__Constants__FOO = "foo";`, which is invalid syntax.

esteve:
CI with `Constants.msg` added:

http://ci.ros2.org/job/ci_linux/875/
http://ci.ros2.org/job/ci_osx/733/
http://ci.ros2.org/job/ci_windows/957/

esteve:
The diff in https://github.com/ros2/rosidl/commit/3a5fe380e63997bde54b2f31edb349b75587a935 doesn't show all the changes and I thought that it didn't change the type for string constants.

This diff shows all the changes https://github.com/ros2/rosidl/compare/workaround-constants...workaround-constants-rawstring

jacquelinekay:
Thank you... my Github-fu is still in training

esteve:
@jacquelinekay no worries, my fault for only looking at the diff and not at the whole branch.

@wjwwood do you have an opinion on whether constants should use `rosidl_generator_c_String` or `const char *`?

wjwwood:
My gut feeling would be to have `const char *` as the constants, since I don't think we can support non-null terminated string constants in the ROS .msg format.

jacquelinekay:
Branch has conflicts. I'm ready to approve this once those are resolved and CI is re-run.

esteve:
@jacquelinekay yeah, I'm fixing them, forgot to comment here.

esteve:
CI:

http://ci.ros2.org/job/ci_linux/881/
http://ci.ros2.org/job/ci_osx/738/
http://ci.ros2.org/job/ci_windows/965/

esteve:
Triggering CI again to check the fix for a warning didn't break anything:

http://ci.ros2.org/job/ci_linux/883/
http://ci.ros2.org/job/ci_osx/739/
http://ci.ros2.org/job/ci_windows/966/

esteve:
The warning is gone:

http://ci.ros2.org/job/ci_linux/883/warnings17Result/fixed/

jacquelinekay:
+1

rosidl_generator_c/CMakeLists.txt
```diff
@@ -59,6 +59,7 @@ if(AMENT_ENABLE_TESTING)
     "msg/Telegram1.msg"
     "msg/Telegram2.msg"
     "msg/Wire.msg"
+    "msg/Constants.msg"
```
dirk-thomas: Out of order.

rosidl_generator_c/resource/msg__struct.h.template
```diff
@@ -96,6 +96,11 @@ enum
 {
   @(key) = @(value)
 };
+@[elif key.split()[0] == 'rosidl_generator_c__String']@
```
dirk-thomas: Instead of comparing the C type this should check the ROSIDL type.

dirk-thomas:
@esteve Can you please address the post-merge comments. Thanks.

rosidl_generator_c/CMakeLists.txt
mikaelarguedas: this has been adressed in #222

rosidl_generator_c/resource/msg__struct.h.template
mikaelarguedas: @dirk-thomas Tentatively adressed in #227

PR #79, opened 2016-02-02
Added connext C typesupport

jacquelinekay:
Connects to ros2/rmw_connext#122

### DIFF
--- a/rosidl_generator_c/CMakeLists.txt
+++ b/rosidl_generator_c/CMakeLists.txt
@@ -16,7 +16,7 @@ add_library(${PROJECT_NAME} SHARED
 )
 if(NOT WIN32)
   set_target_properties(${PROJECT_NAME} PROPERTIES
-    COMPILE_FLAGS "-std=c11 -Wall -Wextra")
+    COMPILE_FLAGS "-Wall -Wextra")
 endif()
 if(WIN32)
   target_compile_definitions(${PROJECT_NAME}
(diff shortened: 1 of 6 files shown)


rosidl_generator_c/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ add_library(${PROJECT_NAME} SHARED
 )
 if(NOT WIN32)
   set_target_properties(${PROJECT_NAME} PROPERTIES
-    COMPILE_FLAGS "-std=c11 -Wall -Wextra")
+    COMPILE_FLAGS "-Wall -Wextra")
```
dirk-thomas: Why not `-std=c11`?

wjwwood: We don't use C11 anywhere. The only thing we use is `stdatomics.h` in `rcl`, but `-std=c11` doesn't affect it at all.

dirk-thomas: I think we should definitely make use of the new features of C11. Therefore I would propose to keep this compile flag *in* in every package containing C code.

wjwwood: I disagree, we should add it when we need it. It has received a lot less adoption than C++11 and so it isn't as low risk to make a hard dependency on it.

rosidl_typesupport_introspection_c/resource/msg__introspection_type_support.h.template
```diff
@@ -36,7 +36,7 @@ extern "C"
 
 ROSIDL_GENERATOR_C_PUBLIC_@(spec.base_type.pkg_name)
 const rosidl_message_type_support_t *
-rosidl_get_message_type_support__@(spec.base_type.pkg_name)__msg__@(spec.base_type.type)();
+rosidl_typesupport_introspection_c_get_message__@(spec.base_type.pkg_name)__msg__@(spec.msg_name)();
```
jacquelinekay: @wjwwood: I think this change might be why the rcl tests for C typesupport pass for FastRTPS on my branches.

PR #80, opened 2016-02-04
reintegrate separate test package into rosidl_generator_py

dirk-thomas:
This updates #73 and merges the separate test package back into `rosidl_generator_py`. 

@esteve I am not sure if I am missing something but this passes the tests for me?


esteve:
This seems to work for me too, happy to know it was just a matter of reverting CMakeLists.txt to the original state, this was driving us nuts :smile:

PR #81, opened 2016-02-09
update generated Python code

dirk-thomas:
The patch updates the generated Python code as follows:

* meta class
  * no need to prefix the metaclass with the message name (`MsgName__Constants`) since the module already provides a namespace
  * since the metaclass provides not only constants but also default values I renamed it to just `Metaclass`
  * properties of the metaclass are instance methods - not class methods -  and should therefore use `self` as the argument
  * no need for `__setattr__`
  * duplicated default values, not scalable when default values can be complex type, therefore move the values from the `__prepare__` function and the properties into an internal class variable
  * add comments to `__prepare__` function to describe why this is necessary
* message class
  * docblock should mention constants (in the future with their description)
  * space after `assert`, otherwise is looks like a function call
* general
  * fix PEP8 and PEP257 errors in the generated code

--

* http://ci.ros2.org/job/ci_linux/914/
* http://ci.ros2.org/job/ci_osx/775/
* http://ci.ros2.org/job/ci_windows/1007/

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.template
+++ b/rosidl_generator_py/resource/_msg.py.template
@@ -1,32 +1,24 @@
 # generated from rosidl_generator_py/resource/_msg.py.template
 # generated code does not contain a copyright notice
 
-class @(spec.base_type.type)__Constants(type):
+
+class Metaclass(type):
+    """Metaclass of message '@(spec.base_type.type)'."""
+
+    __constants = {
 @[for constant in spec.constants]@
-
-    @@property
-    def @(constant.name)(cls):
-        """Message constant '@(constant.name)'"""
-        return @constant_value_to_py(constant.type, constant.value)
+        '@(constant.name)': @constant_value_to_py(constant.type, constant.value),
 @[end for]@
-@[for field in spec.fields]@
-@[if field.default_value]@
-
-    @@property
-    def @(field.name.upper())__DEFAULT(cls):
-        """Default value for message field '@(field.name)'"""
-        return @value_to_py(field.type, field.default_value, array_as_tuple=True)
-@[end if]@
-@[end for]@
-
-    def __setattr__(cls, name, value):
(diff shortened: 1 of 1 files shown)


rosidl_generator_py/resource/_msg.py.template
```diff
@@ -1,32 +1,24 @@
...
 
-    @@property
-    def @(constant.name)(cls):
-        """Message constant '@(constant.name)'"""
-        return @constant_value_to_py(constant.type, constant.value)
-@[end for]@
-@[for field in spec.fields]@
-@[if field.default_value]@
+class Metaclass(type):
+    """MetaClass of message '@(spec.base_type.type)'."""
```
wjwwood: nitpik - Metaclass is a single word (computer science only) and so it doesn't need camel case here: https://en.wikipedia.org/wiki/Metaclass

dirk-thomas: I missed the comment when updating the class name. Fixed.

rosidl_generator_py/resource/_msg.py.template
```diff
@@ -47,7 +65,7 @@ class @(spec.base_type.type)(metaclass=@(spec.base_type.type)__Constants):
 
 @[if len(spec.fields) > 0]@
     def __init__(self, **kwargs):
-        assert(all(['_' + key in self.__slots__ for key in kwargs.keys()])), \
+        assert (all(['_' + key in self.__slots__ for key in kwargs.keys()])), \
```
wjwwood: Are the outside parenthesis (the ones around the call to `all()`) needed?

dirk-thomas: I don't think so. I removed them.

wjwwood:
lgtm, but I'll defer to @esteve.

esteve:
+1, thanks for the cleanups!

PR #82, opened 2016-02-17
update manifests and CMake

dirk-thomas:
http://ci.ros2.org/job/ci_osx/840/


wjwwood:
@dirk-thomas this doesn't merge automatically FYI.

dirk-thomas:
I think the OS X build made it far enough. @esteve I would propose to merge this into the other PR now and then trigger new jobs for #73?

esteve:
@dirk-thomas sounds great.

PR #83, opened 2016-02-17
Avoid importing empy

esteve:
This avoids importing the generator code, which is not necessary at run time. On Windows, if empy is not installed, rosidl_cmake, which is imported by generate_py_impl


rosidl_generator_py/rosidl_generator_py/__init__.py
```diff
@@ -12,10 +12,14 @@
...
 
-from .generate_py_impl import generate_py
-from .import_type_support_impl import import_type_support
+__all__ = ['import_type_support']
+
+try:
+    from .generate_py_impl import generate_py
+    __all__.append('generate_py')
+except ImportError:
+    logger = logging.getLogger('rosidl_generator_py')
```
dirk-thomas: This line fails since `logging` is not imported.

esteve: Fixed.

dirk-thomas: Please try to run the code in the exception block to confirm that it works as expected.

rosidl_generator_py/rosidl_generator_py/__init__.py
```diff
@@ -12,10 +12,17 @@
...
+__all__ = ['import_type_support']
 
-__all__ = [
-    'generate_py',
-    'import_type_support',
-]
+try:
+    from .generate_py_impl import generate_py
+    __all__.append('generate_py')
+except ImportError:
```
esteve: You mean that traceback is not imported? Fixed in https://github.com/ros2/rosidl/commit/ff356c12ef1ce0d82fb9ef283d1407c1a3a30aba

rosidl_generator_py/rosidl_generator_py/__init__.py
esteve: You mean that traceback is not imported? Fixed in ff356c1

rosidl_generator_py/rosidl_generator_py/__init__.py
```diff
@@ -12,10 +12,17 @@
...
+    logger = logging.getLogger('rosidl_generator_py')
+    logger.debug(
+        'Failed to import modules for generating Python structures:\n' + traceback.format_exc())
+
+from .import_type_support_impl import import_type_support
```
dirk-thomas: Please move this line back above line 18 to be close to where it gets exported.

esteve: Done.

dirk-thomas: I meant *above* since the common pattern is:

    import ...
   __all__ = [...]

esteve: Done.

dirk-thomas:
While the patch lgtm I am not sure this is the right approach. https://github.com/ros2/rosidl/blob/master/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py does import `rosidl_cmake` and `rosidl_parser` which makes both of them a exec dependency. Therefore both of those should be added to the manifest. Does that already solve the problem and makes this patch obsolete?

esteve:
@dirk-thomas I'll check that, thanks for the suggestion.

esteve:
@dirk-thomas I've added those two dependencies to the package.xml files in the share directory for rosidl_generator_py, rclpy_examples and rclpy, but it still fails. Do I need to trigger a new packaging job?

esteve:
@dirk-thomas I mean that I modified the package.xml files that are the installed share directory, after extracting the zip file.

dirk-thomas:
Changing the XML won't affect any of the already generated files, e.g. the CMake extra files. Does that only happen for the packaging job or can you reproduce and confirm the problem with a normal from source build?

esteve:
I'll go ahead and build locally, and remove empy before running the examples.

esteve:
I've added the extra dependencies in rclpy, rclpy_examples and rosidl_generator_py, rebuilt, removed python3-empy and the examples still don't work because the message classes can't import the type support (which happens because they can't import generate_impl_py properly):

```
  File "/home/esteve/Projects/ros2/ros2_ws/install_isolated/rclpy_examples/bin/listener_py", line 9, in <module>
    load_entry_point('rclpy-examples==0.0.0', 'console_scripts', 'listener_py')()
  File "/home/esteve/Projects/ros2/ros2_ws/install_isolated/rclpy_examples/lib/python3.4/site-packages/listener_py.py", line 34, in main
    sub = node.create_subscription(String, 'chatter', chatter_callback, qos_profile_default)
  File "/home/esteve/Projects/ros2/ros2_ws/install_isolated/rclpy/lib/python3.4/site-packages/rclpy/node.py", line 33, in create_subscription
    subscription_handle = rclpy._rclpy.rclpy_create_subscription(self.handle, msg_type, topic)
RuntimeError: Failed to create subscriptions: type_support argument is null, at /home/esteve/Projects/ros2/ros2_ws/src/ros2/rcl/rcl/src/rcl/subscription.c:49
```

dirk-thomas:
Do you have a stacktrace with the actual failing import of `empy`? The above only shows the problem later when calling `rclpy`.

Anyway I think the root problem is that the `rosidl_generator_py` package has dependencies it needs to perform the actual code generation step. But it is also used as the runtime dependency for the generated Python messages. I think the only clean solution is to put the code necessary at runtime into a package separate from the code generator.

esteve:
```
esteve@osrf-esteve:~/Projects/ros2/ros2_ws$ python3 -c 'from rosidl_generator_py import import_type_support'
  File "<string>", line 1, in <module>                                                                       
  File "/home/esteve/Projects/ros2/ros2_ws/install_isolated/rosidl_generator_py/lib/python3.4/site-packages/rosidl_generator_py/__init__.py", line 15, in <module>                                                        
    from .generate_py_impl import generate_py                                                                
  File "/home/esteve/Projects/ros2/ros2_ws/install_isolated/rosidl_generator_py/lib/python3.4/site-packages/rosidl_generator_py/generate_py_impl.py", line 18, in <module>                                                
    from rosidl_cmake import convert_camel_case_to_lower_case_underscore                                     
  File "/home/esteve/Projects/ros2/ros2_ws/install_isolated/rosidl_cmake/lib/python3.4/site-packages/rosidl_cmake/__init__.py", line 15, in <module>                                                                      
    import em                                                                                                
ImportError: No module named 'em'
```

But when I install python3-empy:

```
esteve@osrf-esteve:~/Projects/ros2/ros2_ws$ python3 -c 'from rosidl_generator_py import import_type_support' 
```

wjwwood:
I guess we could put the `import_type_support` module in a different Python package. Should `empy` just be installed? It is a run dependency of the generator script also exported by this package. So as long as this package is installed (which it must be for messages to import from it) then em should probably be installed.

The other-other option is to put this code back into each generated Python file which uses it, or generate it once for each message package and import it from there.

dirk-thomas:
Actually all other generators also have a run dependency on `empy` so just getting rid of it in this package by splitting it is probably not worth it.

Maybe the easiest way is to remove `__all__` and let code import the specific module directly. Then running Python messages should not even try to import `empy`.

esteve:
Sounds good. Does any of you have spare time to look into that? I'm trying to fix that C sources are installed into the Python packages, I'm about to get it done, but don't know if it may require more work.

esteve:
I've removed the __all__ logic and now the examples work if python3-empy is not installed.

esteve:
Merging this, since examples run fine locally.

PR #84, opened 2016-02-17
Fix CMAKE_CXX_FLAGS usage

tfoote:
rosidl_generator_py was not building due to missing --std=c++11

```
[ 13%] /usr/bin/c++   -Drosidl_generator_py__rosidl_typesupport_introspection_cpp_EXPORTS -fPIC -I/root/ros2_ws/build/rosidl_generator_py/rosidl_generator_cpp -I/root/ros2_ws/install/include -I/root/ros2_ws/install/include/rosidl_typesupport_introspection_cpp/impl    -o CMakeFiles/rosidl_generator_py__rosidl_typesupport_introspection_cpp.dir/rosidl_typesupport_introspection_cpp/rosidl_generator_py/msg/constants__type_support.cpp.o -c /root/ros2_ws/build/rosidl_generator_py/rosidl_typesupport_introspection_cpp/rosidl_generator_py/msg/constants__type_support.cpp
Built target rosidl_generator_py__rosidl_typesupport_introspection_c
In file included from /usr/include/c++/4.8/array:35:0,
                 from /root/ros2_ws/build/rosidl_generator_py/rosidl_generator_cpp/rosidl_generator_py/msg/empty__struct.hpp:7,
                 from /root/ros2_ws/build/rosidl_generator_py/rosidl_typesupport_introspection_cpp/rosidl_generator_py/msg/empty__type_support.cpp:14:
/usr/include/c++/4.8/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
 #error This file requires compiler and library support for the \
  ^
In file included from /usr/include/c++/4.8/array:35:0,
                 from /root/ros2_ws/build/rosidl_generator_py/rosidl_generator_cpp/rosidl_generator_py/msg/bool__struct.hpp:7,
                 from /root/ros2_ws/build/rosidl_generator_py/rosidl_typesupport_introspection_cpp/rosidl_generator_py/msg/bool__type_support.cpp:14:
/usr/include/c++/4.8/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
 #error This file requires compiler and library support for the \
  ^
```

And when I was comparing to `rosidl_generator_cpp` I noticed the duplicate code.

### DIFF
--- a/rosidl_generator_cpp/CMakeLists.txt
+++ b/rosidl_generator_cpp/CMakeLists.txt
@@ -2,10 +2,6 @@ cmake_minimum_required(VERSION 2.8.3)
 
 project(rosidl_generator_cpp)
 
-if(NOT WIN32)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
-endif()
-
 find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_python REQUIRED)
 
--- a/rosidl_typesupport_introspection_cpp/cmake/rosidl_typesupport_introspection_cpp_generate_interfaces.cmake
+++ b/rosidl_typesupport_introspection_cpp/cmake/rosidl_typesupport_introspection_cpp_generate_interfaces.cmake
@@ -85,6 +85,10 @@ if(WIN32)
   target_compile_definitions(${rosidl_generate_interfaces_TARGET}${_target_suffix}
     PRIVATE "ROSIDL_TYPESUPPORT_INTROSPECTION_CPP_BUILDING_DLL")
 endif()
+if(NOT WIN32)
+  set_target_properties(${rosidl_generate_interfaces_TARGET}${_target_suffix}
+    PROPERTIES COMPILE_FLAGS "-std=c++11 -Wall -Wextra")
+endif()
 target_include_directories(${rosidl_generate_interfaces_TARGET}${_target_suffix}
   PUBLIC
   ${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_cpp


esteve:
+1

wjwwood:
I don't think this is the right fix. I'll explain in a follow up comment.

rosidl_generator_py/CMakeLists.txt
```diff
@@ -2,6 +2,10 @@ cmake_minimum_required(VERSION 2.8.3)
 
 project(rosidl_generator_py)
 
+if(NOT WIN32)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
+endif()
```
dirk-thomas: Why would this be necessary? The package itself doesn't build any C++ code and each message generator needs to take care of its compiler flags on its own.

rosidl_generator_cpp/CMakeLists.txt
```diff
@@ -53,10 +53,6 @@ if(AMENT_ENABLE_TESTING)
     SKIP_INSTALL
   )
 
-  if(NOT WIN32)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
-  endif()
```
dirk-thomas: This is necessary to build the test with C++11 and all warning enabled. It could be moved down to the test target though.

tfoote: I removed this because it's already in lines 5-7, it could be moved below

jacquelinekay:
Connext C++ typesupport already does this: https://github.com/ros2/rmw_connext/blob/master/rosidl_typesupport_connext_cpp/cmake/rosidl_typesupport_connext_cpp_generate_interfaces.cmake#L163

I agree with William, it makes sense that the C++ typesupport package should bring in the relevant C++11 flags, not the Python generator.

tfoote:
Updated PR with exported flags and switched deduplication to be inside test block. Local testing in progress.

rosidl_typesupport_introspection_cpp/cmake/rosidl_typesupport_introspection_cpp_generate_interfaces.cmake
```diff
@@ -85,6 +85,10 @@ if(WIN32)
   target_compile_definitions(${rosidl_generate_interfaces_TARGET}${_target_suffix}
     PRIVATE "ROSIDL_TYPESUPPORT_INTROSPECTION_CPP_BUILDING_DLL")
 endif()
+if(NOT WIN32)
+  set_target_properties(${rosidl_generate_interfaces_TARGET}${_target_suffix}
+    PROPERTIES COMPILE_FLAGS "-std=c++11")
```
dirk-thomas: Please also keep ` -Wall -Wextra`.

tfoote: Can do, should I add that to the other instance?

wjwwood:
+1

rosidl_typesupport_introspection_cpp/cmake/rosidl_typesupport_introspection_cpp_generate_interfaces.cmake
dirk-thomas: We claim to build everything with max warning level. So if it is missing somewhere I think it should be added.

tfoote:
One CI Passed and two local tests also pass with this now: http://ci.ros2.org/job/ci_linux/1014/

Issue #87, opened 2016-02-17
Generated C source code files from rosidl_generator_py are installed


esteve:
The install directory should only contain .py files and native extensions

dirk-thomas:
While this is certainly not good I would suggest to focus on the blockers for now. This is a nice-to-have.

PR #88, opened 2016-02-17
Use install to install Python extension and generated code

esteve:
Connects to #87

### DIFF
--- a/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
@@ -142,6 +142,29 @@ rosidl_write_generator_arguments(
 file(MAKE_DIRECTORY "${_output_path}")
 file(WRITE "${_output_path}/__init__.py" "")
 
+if(NOT "${_generated_msg_py_files} " STREQUAL " ")
+  list(GET _generated_msg_py_files 0 _msg_file)
+  get_filename_component(_msg_package_dir1 "${_msg_file}" DIRECTORY)
+  get_filename_component(_msg_package_dir2 "${_msg_package_dir1}" NAME)
+endif()
+
+if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
+  ament_python_install_module("${_output_path}/__init__.py"
+    DESTINATION "${PROJECT_NAME}"
+  )
+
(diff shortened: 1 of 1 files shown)


rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -142,6 +142,26 @@ rosidl_write_generator_arguments(
...
+  ament_python_install_module("${_output_path}/__init__.py"
+    DESTINATION "${PROJECT_NAME}"
+  )
+
+  ament_python_install_module("${_output_path}/__init__.py"
+    DESTINATION "${PROJECT_NAME}/${_msg_package_dir2}"
+  )
+
+  install(FILES ${_generated_msg_py_files}
+    DESTINATION "${PYTHON_INSTALL_DIR}/${PROJECT_NAME}/${_msg_package_dir2}"
```
dirk-thomas: This should use the Python install functions to also benefit from future improvements like compiling the Python code.

esteve: Yes, but unfortunately when I replaced this with `ament_python_install_module`, the code that checks if the file to be installed exists fails. Could this be because that doesn't run at install time?

wjwwood: Arguably lines 32-38 in `ament_python_install_module.cmake` (https://github.com/ament/ament_cmake/blob/master/ament_cmake_python/cmake/ament_python_install_module.cmake#L32-L38) are too aggressive and should be removed. The `INSTALL` command should be able to figure out if the files exist at the appropriate time, i.e. at the point of install.

Also it either needs to be able to take more than one module file at a time or you'll have to iterate over the files and call `ament_python_install_module` N times.

I think we should skip that for now and make an issue to track the problem (as well as a TODO) and put it on the list for next cycle since it's more than just changing the call in this file.

wjwwood: I made an issue https://github.com/ros2/rosidl/issues/89, @esteve can you add a TODO here? And then I think we should merge it and move on.

esteve: TODO added.

wjwwood:
Thanks, +1.

PR #91, opened 2016-02-18
Fix import logic

### LINKED ISSUE #90: Fix rosidl_generator_py imports
http://ci.ros2.org/job/ci_linux/1018/

esteve:
Although this is not the best solution, we need to move forward.

Fixes #90

### DIFF
--- a/rosidl_generator_py/bin/rosidl_generator_py
+++ b/rosidl_generator_py/bin/rosidl_generator_py
@@ -5,7 +5,7 @@ import os
 import sys
 
 try:
-    from rosidl_generator_py.generate_py_impl import generate_py
+    from rosidl_generator_py import generate_py
 except ImportError:
     # modifying sys.path and importing the Python package with the same
     # name as this script does not work on Windows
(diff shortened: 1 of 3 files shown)


wjwwood:
For context without this the import no longer works:

http://ci.ros2.org/job/ci_linux/1018/console
```
00:54:04 Traceback (most recent call last):
00:54:04   File "/home/rosbuild/ci_scripts/ws/src/ros2/rosidl/rosidl_generator_py/bin/rosidl_generator_py", line 8, in <module>
00:54:04     from rosidl_generator_py.generate_py_impl import generate_py
00:54:04 ImportError: No module named 'rosidl_generator_py'
00:54:04 
00:54:04 During handling of the above exception, another exception occurred:
00:54:04 
00:54:04 Traceback (most recent call last):
00:54:04   File "/home/rosbuild/ci_scripts/ws/src/ros2/rosidl/rosidl_generator_py/bin/rosidl_generator_py", line 21, in <module>
00:54:04     generate_py = rosidl_generator_py.generate_py
00:54:04 AttributeError: 'module' object has no attribute 'generate_py'
```

wjwwood:
We're going to fix it after the alpha (https://github.com/ros2/rosidl/issues/92). +1

jacquelinekay:
Not the solution I was expecting, but yeah, we need to move on.

PR #93, opened 2016-02-19
change DESTINATION argument name of ament_python_install_module()

dirk-thomas:
Depends on ament/ament_cmake#58


dirk-thomas:
With the referenced PR being approved I will go ahead and also merge this one.

PR #94, opened 2016-02-19
Share enums for fields with rosidl_typesupport_introspection_c

jacquelinekay:
As discussed in ros2/rmw_connext#133

These enum types should be shared between introspection_c and introspection_cpp to simplify supporting both C and C++ typesupport in an rmw implementation.

Currently, sharing these enums works because the fields happen to have the same underlying int values.

However, if the definitions are reordered in some way (the int values are changed) in one package, code paths shared for C and C++ introspection will stop work.

This PR creates a dependency of introspection_cpp on introspection_c so that it can directly map its field definitions to introspection_c's field enums.

### DIFF
--- a/rosidl_typesupport_introspection_cpp/CMakeLists.txt
+++ b/rosidl_typesupport_introspection_cpp/CMakeLists.txt
@@ -8,10 +8,13 @@ endif()
 
 find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_python REQUIRED)
+find_package(rosidl_typesupport_introspection_c REQUIRED)
 
 ament_export_dependencies(rosidl_cmake)
 ament_export_dependencies(rosidl_generator_c)
 ament_export_dependencies(rosidl_generator_cpp)
+ament_export_dependencies(rosidl_typesupport_introspection_c)
+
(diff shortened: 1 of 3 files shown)


rosidl_typesupport_introspection_cpp/cmake/rosidl_typesupport_introspection_cpp_generate_interfaces.cmake
```diff
@@ -99,7 +99,8 @@ foreach(_pkg_name ${rosidl_generate_interfaces_DEPENDENCY_PACKAGE_NAMES})
     ${_pkg_name})
 endforeach()
 ament_target_dependencies(${rosidl_generate_interfaces_TARGET}${_target_suffix}
-  "rosidl_typesupport_introspection_cpp")
+  "rosidl_typesupport_introspection_cpp"
+  "rosidl_typesupport_introspection_c")
```
dirk-thomas: This should not be necessary since `rosidl_typesupport_introspection_cpp` will already pull in all its dependencies. No need to specify transitive ones explicitly.

tfoote:
+1

PR #95, opened 2016-02-19
Fail in rosidl_generator_py if no typesupport implementations found

### LINKED ISSUE #99: build fails when only building FastRTPS
http://ci.ros2.org/job/ci_windows/1092/

@esteve Can you please take a look at this since it seems to be related to the Python message generator.

jacquelinekay:
Fixes #99
this is possibly a fix for python messages appearing to fail to build in ros2/common_interfaces#11

### DIFF
--- a/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
@@ -26,24 +26,12 @@ find_package(PythonExtra MODULE)
 # TODO(esteve): force opensplice and connext C type supports only, uncomment
 # the following line when all typesupport implementations are ported to C
 #get_rmw_typesupport(_typesupport_impls ${rmw_implementation})
-set(_typesupport_impls "")
-foreach(_extension IN LISTS AMENT_EXTENSIONS_rosidl_generate_interfaces)
-  string(REPLACE ":" ";" _extension_list "${_extension}")
-  list(LENGTH _extension_list _length)
-  if(NOT _length EQUAL 2)
-    message(FATAL_ERROR "ament_execute_extensions(${extension_point}) "
-      "registered extension '${_extension}' can not be split into package "
-      "name and cmake filename")
-  endif()
-  list(GET _extension_list 0 _pkg_name)
-  list(GET _extension_list 1 _cmake_filename)
-  if("${_pkg_name} " STREQUAL "rosidl_typesupport_opensplice_c ")
-    list(APPEND _typesupport_impls "rosidl_typesupport_opensplice_c")
-  endif()
-  if("${_pkg_name} " STREQUAL "rosidl_typesupport_connext_c ")
-    list(APPEND _typesupport_impls "rosidl_typesupport_connext_c")
-  endif()
-endforeach()
+rosidl_generator_py_get_typesupports(_typesupport_impls)
+
+if("${_typesupport_impls} " STREQUAL " ")
+  message(WARNING "No valid typesupport for Python generator. Python messages will not be generated.")
+  return()
+endif()
 
 set(_output_path
   "${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py/${PROJECT_NAME}")
--- a/rosidl_generator_py/CMakeLists.txt
+++ b/rosidl_generator_py/CMakeLists.txt
@@ -40,32 +40,43 @@ if(AMENT_ENABLE_TESTING)
   )
 
   include(cmake/register_py.cmake)
+  include(cmake/rosidl_generator_py_get_typesupports.cmake)
+
   set(rosidl_generator_py_DIR "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
+  # Need to call extras before get_typesupports, to register the extension
   rosidl_generator_py_extras(
     "${CMAKE_CURRENT_SOURCE_DIR}/bin/rosidl_generator_py"
     "${CMAKE_CURRENT_SOURCE_DIR}/rosidl_generator_py/__init__.py"
     "${CMAKE_CURRENT_SOURCE_DIR}/resource"
   )
 
-  rosidl_generate_interfaces(${PROJECT_NAME} ${message_files}
-    SKIP_INSTALL
-  )
+  # If no message files were generated, skip the test.
+  rosidl_generator_py_get_typesupports(_typesupport_impls)
+  if("${_typesupport_impls} " STREQUAL " ")
+    message(WARNING "No valid typesupport found for Python generator, skipping tests.")
+  else()
+    rosidl_generate_interfaces(${PROJECT_NAME} ${message_files}
+      SKIP_INSTALL
+    )
 
-  set(_append_library_dirs "")
-  if(WIN32)
-    set(_append_library_dirs "${CMAKE_CURRENT_BINARY_DIR}/$<CONFIGURATION>")
+    set(_append_library_dirs "")
+    if(WIN32)
+      set(_append_library_dirs "${CMAKE_CURRENT_BINARY_DIR}/$<CONFIGURATION>")
+    endif()
+
+    ament_add_nose_test(test_interfaces_py "test/test_interfaces.py"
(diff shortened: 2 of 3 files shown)


rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -45,6 +45,10 @@ foreach(_extension IN LISTS AMENT_EXTENSIONS_rosidl_generate_interfaces)
   endif()
 endforeach()
 
+if (_typesupport_impls STREQUAL "" )
```
esteve: Shouldn't this be like this? My CMake-fu is pretty bad.

`if ("${_typesupport_impls} " STREQUAL " ")`

dirk-thomas:
In order to build e.g. only with FastRTPS various packages need to be manually ignores. I don't think this is a viable option.

Instead the generator could simply warn about not being able to generate Python messages and do nothing.

jacquelinekay:
Alright, I'll change it to a warning and `return()` (is that valid CMake?).

This is what currently happens on master when only FastRTPS is enabled (same error message as ros2/common_interfaces#11):

http://ci.ros2.org/job/ci_linux/1027/console

This is only FastRTPS enabled and this PR:

http://ci.ros2.org/job/ci_linux/1028/

jacquelinekay:
Looks like this PR fixes the broken build with FastRTPS-only. I fixed the lint error in ab4993c

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -45,6 +45,11 @@ foreach(_extension IN LISTS AMENT_EXTENSIONS_rosidl_generate_interfaces)
   endif()
 endforeach()
 
+if("${_typesupport_impls} " STREQUAL " ")
+  message(WARNING "No valid typesupport implementations found.")
```
dirk-thomas: The message should explicitly mention the Python generator since that is not obvious when reading the message in the CMake output.

jacquelinekay:
e369517 should fix the test failures originating in rosidl_generator_py in http://ci.ros2.org/job/ci_windows/1096/. It will avoid building the rosidl_generator_py test if no valid middlewares were found.

This strikes me as a bit of a hack right now, but I wasn't sure of a better solution.

The other test failures may require implementing rclpy for multiple typesupports or more CMake hacks to avoid installing rclpy if Opensplice was not found, which will probably be faster to implement but accrue more technical debt.

dirk-thomas:
The flag detecting if the Python code generator ran only allows to skip part of the overhead. The message generation is still happening but not used anywhere.

Therefore I think the CMakeLists.txt should determine if message generation for Python is feasible and only if that is the case run all the message generation and register the nosetest.

If the nosetests are not run the whole message generation part is not necessary and should also be skipped.

jacquelinekay:
ok. Since the generator already checks for valid typesupports I will make a macro for finding valid typesupports. That way I can reuse the macro in rclpy and/or rclpy_examples to avoid installing executables that will not work when a valid middleware is not available.

dirk-thomas:
@jacquelinekay Perfect.

jacquelinekay:
Alright, I implemented the macro and opened ros2/rclpy#6 to use it to skip tests in rclpy if no typesupport is found.

http://ci.ros2.org/job/ci_linux/1038/
http://ci.ros2.org/job/ci_osx/857/
http://ci.ros2.org/job/ci_windows/1100/

rosidl_generator_py/cmake/find_valid_typesupport.cmake
```diff
@@ -0,0 +1,34 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+macro(rosidl_generator_py_find_valid_typesupport TYPESUPPORT)
+  set(${TYPESUPPORT} "")
+  foreach(_extension IN LISTS AMENT_EXTENSIONS_rosidl_generate_interfaces)
```
dirk-thomas: It is kind of opaque where `AMENT_EXTENSIONS_rosidl_generate_interfaces` comes from here. Not sure if it is ensured that this works wherever / however the macro is being called.

rosidl_generator_py/cmake/find_valid_typesupport.cmake
```diff
@@ -0,0 +1,34 @@
+# Copyright 2016 Open Source Robotics Foundation, Inc.
```
dirk-thomas: Commonly the CMake filenames match the function / macro name for easier discoverability.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -26,24 +26,12 @@ find_package(PythonExtra MODULE)
 # TODO(esteve): force opensplice and connext C type supports only, uncomment
 # the following line when all typesupport implementations are ported to C
 #get_rmw_typesupport(_typesupport_impls ${rmw_implementation})
```
dirk-thomas: This comment should be moved to the new location in the macro.

rosidl_generator_py/cmake/find_valid_typesupport.cmake
jacquelinekay: yeah, actually, I think this will actually disable the tests even when there is a valid typesupport impl... I think `AMENT_EXTENSIONS_rosidl_generate_interfaces` comes from `rosidl_generator_py_extras`, which calls `ament_register_extension`... Fixing it now

rosidl_generator_py/cmake/find_valid_typesupport.cmake
jacquelinekay: Aesthetically, which do you think is better, prepending the macro with the package name (the macro name) or not (the file name)?

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
jacquelinekay: I actually think this comment is ok. lines 29-46 here, which were moved into the macro, are a substitute for line 28, `get_rmw_typesupport`, which should be reenabled when all typesupport impls are ported to C. The comment above refers to line 28.

rosidl_generator_py/cmake/find_valid_typesupport.cmake
dirk-thomas: Most CMake files have one function / macro per file and their names match: https://github.com/ament/ament_cmake/tree/master/ament_cmake_core/cmake/environment

They all have a namespace like prefix depending on the context it might not be the full package name. E.g. MAny ament functions just start with `ament_`: e.g. https://github.com/ament/ament_cmake/blob/master/ament_cmake_export_libraries/cmake/ament_export_library_names.cmake

For this case maybe `rosidl_generator_py_get_typesupports(...)` (since I don't see a good way to shorten the package name)?

jacquelinekay:
Retriggering with FastRTPS only: http://ci.ros2.org/job/ci_windows/1102/ (pending)

jacquelinekay:
this has been sitting in review for a while. I will rebase and retrigger CI (including a FastRTPS-only build) and squash and merge if the CI looks good.

jacquelinekay:
http://ci.ros2.org/job/ci_linux/1087/
http://ci.ros2.org/job/ci_osx/880/
http://ci.ros2.org/job/ci_windows/1153

http://ci.ros2.org/job/ci_linux/1088 (FastRTPS only again)

wjwwood:
Looks like @dirk-thomas reviewed this and ended up with a +1-ish response. @jacquelinekay do you need more reviewers?

jacquelinekay:
I just need to address the test failures in http://ci.ros2.org/job/ci_linux/1088/

I thought my rclpy pull request would disable those tests. Investigating now.

jacquelinekay:
k, I had to add ros2/examples#94 to fix that failing test. I will remove the workaround once we have support for all middlewares in rclpy.

http://ci.ros2.org/job/ci_linux/1090/

jacquelinekay:
The workaround works for the FastRTPS-only case. I will squash and merge these PRs unless there's an objection.

PR #100, opened 2016-03-01
Added Java generator

esteve:
This adds a rosidl generator for Java.


esteve:
This generator is not complete.

I very much rather this were in a separate repository, so if anyone knows of way to have this working without requiring to merge into the rosidl repository, I'd be happy to move this elsewhere.

I had to add rosidl_generator_java as part of the default generators, so again, if there's a way so that I can generate Java code for messages without forcing this to be a default generator, I'd be okay with withdrawing this PR.

dirk-thomas:
As mentioned in the other PR I don't think we should make our code base require Java.

In order to extend default generators without strict dependencies the dependency system needs to be reworked (and it is still unclear how to do that).

wjwwood:
The options, as I remember them, are to always build the java generator in a separate workspace before the packages which generate messages (forcing the build order), or implementing a more general solution like the "provides" pattern.

The idea of the "provides" pattern is that packages can state that they provide something by name, e.g. the java generator could state (presumably in its `package.xml`) that it provides "message generation", and other packages can depend on packages which provide something by name, e.g. `rosidl_default_generators` could depend on "message generation" providers. The build tool would take these into consideration when doing topological ordering. You would still have explicit dependencies on some generators in `rosidl_default_generators` because the "provides" dependencies would not be considered when building binary packages.

We've discussed this idea on more than one occasion, but never implemented it. I think it has some kinks to work out, but in principal I think it would work.

firesurfer:
From a user perspective I would really appreciate having as many message generators as possible available in the default configuration. So I wouldn't mind adding a message generator for java to the default generators. The only thing I would mind (from a user perspective) is if the framework I want to use requires java even if I don't want to use java. So perhaps having the build system check for an existing java installation would be a nice in-between.

From a developer perspective I would say that it is important to be able to add custom generators, but I would also don't mind having generators for the most common languages (I think about contributing a generator for C# ). 
At this point I have to say that I am not that familiar with the java native interface but I want to point out some solutions you might not have considered yet. 
I think that it wouldn't be necessary to have java as a dependency just for generating fitting message files. It should be possible to generate the interface files without having a java installation on the system. Building the native intefaces could be done at the point where a package wants to use them. At this point the user should have ensured that the system has a valid java installation.

For C# I think it wouldn't even be necessary to generate any native interfaces files, because it is possible to directly map C# structs to their equivalent in C. Perhaps this is also possible in java? (This might render the need of JNI files unnecessary ?!)

esteve:
So, how about adding a dependency in `rosidl_default_generators` to a new package (e.g. `rosidl_external_generators`), which lives in a separate repository and that it's shipped empty by default? Users can always replace that repository/package with their own copy to override it, I think it's a fair compromise and allows some degree of flexibility, without forcing more generators down to our users.

esteve:
I went ahead and implemented the approach I described, see #109 and https://github.com/esteve/rosidl_external_generators

dirk-thomas:
I don't see how introducing the package `rosidl_default_generators` makes it any better. If users want to build with a custom generator they need to provide a custom implementation of that package. As soon as you want to build with two custom generators you can't use this custom package anymore but need to create a custom one for that case anyway.

Therefore I don't think introducing this new package makes it easier then simply adding the extra dependencies into the `rosidl_default_generators` package when needed.

If we want to support "external" generators better we have to design the idea we sketched before and @wjwwood described above.

esteve:
@dirk-thomas I think it's a compromise, until there's support for a `<provides>` tag in package.xml. You can add your own generators without having to fork the entire `rosidl` repository, as this PR demonstrates. This approach would at least allow users to create a custom package that minimizes the changes and the disruption, otherwise they need to fork `rosidl` and keep an up to date copy. For example, I have a local copy of `rosidl_external_generators` that just adds a new dependency to the Java generator, and the rosidl repository is intact, because the `rosidl_generator_java` is in the `ros2_java repository`.

I'm not claiming this should be the perfect solution and I agree with you and @wjwwood about adding a `<provides>` tag, but in the meantime I think this is a valid temporary workaround.

dirk-thomas:
The approach only works for a single external generator. Once you have two you still need a custom package. Also you can not bundle the custom `rosidl_external_generators` in the same repo as your external generator. So I don't see what the advantage is over simply adding the additional dependency and keeping the change in your local working copy.

We have talked about the `provides` tag before. Also about a "catch-up" message generator. I think we should to investigate both of these options and instead of introducing this temporary hack.

esteve:
@dirk-thomas yes, you need a custom package, a local copy as I said. The advantage is that rosidl_default_generators lives in the same repo as the rest of the rosidl stuff, so you need to keep your local copy up to date, whereas with a separate repository, your changes are self contained and do not interfere with the rest of the system (see how small the rosidl_external_generators package is)

esteve:
Instead of merging this PR, document that users who want to use additional generators (e.g. Java), modify their local copy of `rosidl_default_generators`.

The desired solution would be so that the package.xml file for each generator can have a `<provides>` tag so they can be plugged into the dependency system.

PR #102, opened 2016-03-03
Fix pyflakes complaining of imported but unused module.

ernestmc:
Fixes the following error:
```
-- run_test.py: invoking following command in '/home/rosbuild/ci_scripts/ws/src/ros2/rosidl/rosidl_generator_py':
 - /home/rosbuild/ci_scripts/ws/install/ament_pyflakes/bin/ament_pyflakes --xunit-file /home/rosbuild/ci_scripts/ws/build/rosidl_generator_py/test_results/rosidl_generator_py/pyflakes.xunit.xml
rosidl_generator_py/__init__.py:23: 'generate_py' imported but unused
1 errors
rosidl_generator_py/__init__.py

rosidl_generator_py/generate_py_impl.py

rosidl_generator_py/import_type_support_impl.py

test/test_interfaces.py

-- run_test.py: return code 1
-- run_test.py: verify result file '/home/rosbuild/ci_scripts/ws/build/rosidl_generator_py/test_results/rosidl_generator_py/pyflakes.xunit.xml'
```

### DIFF
--- a/rosidl_generator_py/rosidl_generator_py/__init__.py
+++ b/rosidl_generator_py/rosidl_generator_py/__init__.py
@@ -21,6 +21,7 @@ __all__ = ['import_type_support']
 
 try:
     from .generate_py_impl import generate_py
+    assert generate_py
     __all__.append('generate_py')
 except ImportError:
     logger = logging.getLogger('rosidl_generator_py')


rosidl_generator_py/rosidl_generator_py/__init__.py
```diff
@@ -21,6 +21,7 @@
 
 try:
     from .generate_py_impl import generate_py
+    assert generate_py
```
wjwwood: Thanks for looking into this @ernestmc. I believe you can more simple add `  # noqa` to the end of line 23, rather than adding an assert. I think that would be a better way to work around the warning.

ernestmc: @wjwwood I believe this only works with [flake8](https://pypi.python.org/pypi/flake8)

wjwwood: That's another good reason to just use `flake8`... Oh well. Thanks for looking into it @ernestmc.

jacquelinekay:
Launching CI: http://ci.ros2.org/job/ci_linux/1052/

jacquelinekay:
+1

wjwwood:
+1

PR #103, opened 2016-03-03
replace assert with minimal macro exiting

dirk-thomas:
`assert` can not be used in a unit test. When build in release mode the check just goes away.

This will fix the following compiler warnings http://ci.ros2.org/job/ci_linux/1051/warnings17Result/category.-861405076/ and will actually do some testing when building in release mode.

### DIFF
--- a/rosidl_generator_c/test/test_interfaces.c
+++ b/rosidl_generator_c/test/test_interfaces.c
@@ -77,6 +77,8 @@
 
 #define ARRAY_SIZE 7
 
+#define EXPECT_TRUE(expr) if (!(expr)) exit(1)
+
 void test_primitives(void);
 void test_primitives_default_value(void);
 void test_strings(void);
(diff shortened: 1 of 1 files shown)


wjwwood:
You could also use C++ and GTest to write the tests. This what I do in `rcl` for example.

dirk-thomas:
Currently the code is being compiler with a C compiler. And I think it makes sense to keep that since it is a good check that the code is actually compiling with plain C.

wjwwood:
If you want to check that, but also avoid making your own test macros, you can use GTest and C++ in the test and have a separate `.c` file which just includes the header as a test that it compiles with C99 or whatever.

dirk-thomas:
Is can certainly be refactored like this. But for the focus today of fixing the compiler warnings and test failures I think this is sufficient as it is. I have updated the PR to use `EXPECT_TRUE`. So anytime gtest is being added as you described it the custom macro can simply be dropped.

wjwwood:
Sounds good to me.

jacquelinekay:
It looks like this PR introduced some cpplint warnings.

http://ci.ros2.org/job/ci_windows/1117/testReport/rosidl_generator_c/cpplint/readability_check__2___C__J_workspace_ci_windows_ws_src_ros2_rosidl_rosidl_generator_c_test_test_interfaces_c_137_/

dirk-thomas:
New iteration #104 is building.

PR #104, opened 2016-03-03
use macro names cpplint is happy with

dirk-thomas:
Follow up of #103. `cpplint` "known" about `EXPECT_TRUE` and complains. So using custom `EXPECT_EQ` and `EXPECT_NE` now.

http://ci.ros2.org/job/ci_linux/1056/


jacquelinekay:
You would think cpplint would be smart enough to check if gtest is included in the file before giving this warning. Oh well.

PR #105, opened 2016-03-04
Fix python generator warnings

jacquelinekay:
Attempts to fix all "possible loss of data" warnings for float32, int8, int16, uint8, uint16 data types in the generated Python C bindings.

http://ci.ros2.org/job/ci_windows/1120

### DIFF
--- a/rosidl_generator_c/rosidl_generator_c/__init__.py
+++ b/rosidl_generator_c/rosidl_generator_c/__init__.py
@@ -90,6 +90,10 @@ def get_typename_of_base_type(type_):
     return 'rosidl_generator_c__' + suffix
 
 
+def primitive_msg_type_to_c(type_):
+    return MSG_TYPE_TO_C[type_]
+
+
 def msg_type_to_c(type_, name_):
     """
     Convert a message type into the C declaration.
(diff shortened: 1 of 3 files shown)


dirk-thomas:
LGTM.

For a change like this it would be more efficient to run the CI job:
* with less rmw implementation as well as
* with less repositories

jacquelinekay:
Yeah, earlier I was playing around with `--end-with rosidl_generator_py` and with only Opensplice. I will remember that for next time.

dirk-thomas:
The console output already contains the same warnings, e.g. for `:\J\workspace\ci_windows\ws\build\rosidl_generator_py\rosidl_generator_py\rosidl_generator_py\msg\_float32_s.`

Does it make sense to continue the job?

jacquelinekay:
that is strange, they appeared fixed locally. Yeah, I'll cancel it and take a look.

jacquelinekay:
Trying again with fixes:
http://ci.ros2.org/job/ci_windows/1124

jacquelinekay:
argh, I hadn't pushed a local change from my Windows VM so the first build didn't work

http://ci.ros2.org/job/ci_windows/1131

jacquelinekay:
The "possible loss of data" warnings are now fixed.

PR #109, opened 2016-03-05
Added rosidl_external_generators

esteve:
Adds rosidl_external_generators as a dependency, see https://github.com/esteve/rosidl_external_generators


esteve:
Closing in favor of te solution in https://github.com/ros2/rosidl/pull/100#issuecomment-194122422

PR #111, opened 2016-03-25
fix dependencies in Python generator

dirk-thomas:
The first commit just removes an unused argument / parameter which I ran into when debugging this.

The second commit adds the missing target dependency. Otherwise when built with multiple threads each library target can trigger the custom command invoking the code generator. Commonly the code generator gets executed N times where N is the number of threads.

The patch does likely not only fix race conditions like http://ci.ros2.org/job/ci_linux/1111/console but also significantly reduces the resource usage. E.g. building only `rosidl_generator_py` without this patch:

    real  1m49.139s
    user  8m4.513s
    sys   0m18.413s

With this patch:

    real  1m6.204s
    user  2m51.971s
    sys   0m17.512s

* http://ci.ros2.org/job/ci_linux/1117/
* http://ci.ros2.org/job/ci_osx/897/
* http://ci.ros2.org/job/ci_windows/1180/

### DIFF
--- a/rosidl_generator_py/bin/rosidl_generator_py
+++ b/rosidl_generator_py/bin/rosidl_generator_py
@@ -28,16 +28,12 @@ def main(argv=sys.argv[1:]):
         required=True,
         help='The location of the file containing the generator arguments')
     parser.add_argument(
-        '--typesupport-impl',
-        required=True,
-        help='The typesupport implementation targeted by the C interface')
-    parser.add_argument(
         '--typesupport-impls',
         required=True,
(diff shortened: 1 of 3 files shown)


jacquelinekay:
Yeah, I was wondering why the generator often prints the message "Generating Python code for ROS interfaces" multiple times in a row. That was bothering me.

+1

mikaelarguedas:
Yeah that was bothering me too and I started to look into it yesterday. Thanks @dirk-thomas ! lgtm

PR #98, opened 2016-02-24
Consolidate C code of Python extensions

esteve:
Consolidate Python extensions into one per package and per type support implementation

Fixes #96

PR #113, opened 2016-04-03
Add nested msg support

### LINKED ISSUE #96: Look into generating only one Python library per message package
This may reduce compilation time and also simplify our CMake scripts.

mikaelarguedas:
Feature: add nested message support + single python extension per package per type support implementation
Fixes #96 and should replace #98

### DIFF
--- a/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
@@ -38,8 +38,12 @@ set(_output_path
 set(_generated_msg_py_files "")
 set(_generated_msg_c_files "")
 set(_generated_msg_c_common_files "")
-set(_generated_msg_c_ts_files "")
 set(_generated_srv_files "")
+
+foreach(_typesupport_impl ${_typesupport_impls})
+  set(_generated_msg_c_ts_${_typesupport_impl}_files "")
+endforeach()
+
 foreach(_idl_file ${rosidl_generate_interfaces_IDL_FILES})
   get_filename_component(_parent_folder "${_idl_file}" DIRECTORY)
   get_filename_component(_parent_folder "${_parent_folder}" NAME)
@@ -57,13 +61,8 @@ foreach(_idl_file ${rosidl_generate_interfaces_IDL_FILES})
       "${_output_path}/${_parent_folder}/_${_module_name}_s.c"
     )
     foreach(_typesupport_impl ${_typesupport_impls})
-      list(APPEND _generated_msg_c_files
-        "${_output_path}/${_parent_folder}/_${_module_name}_s.ep.${_typesupport_impl}.c"
-      )
-      list(APPEND _generated_msg_c_ts_files
-        "${_output_path}/${_parent_folder}/_${_module_name}_s.ep.${_typesupport_impl}.c"
-      )
-      list(APPEND _type_support_by_generated_msg_c_files "${_typesupport_impl}")
+      list_append_unique(_generated_msg_c_files "${_output_path}/${_parent_folder}/_${PROJECT_NAME}_s.ep.${_typesupport_impl}.c")
+      list_append_unique(_generated_msg_c_ts_${_typesupport_impl}_files "${_output_path}/${_parent_folder}/_${PROJECT_NAME}_s.ep.${_typesupport_impl}.c")
     endforeach()
   elseif("${_parent_folder} " STREQUAL "srv ")
     list(APPEND _generated_srv_files
@@ -158,7 +157,6 @@ set(_extension_dependencies "")
 set(_target_suffix "__py")
 
 add_custom_command(
-  # OUTPUT ${_generated_msg_py_files} ${_generated_msg_c__${_typesupport_impl}_files} ${_generated_srv_files}
   OUTPUT ${_generated_msg_py_files} ${_generated_msg_c_files} ${_generated_srv_files}
   COMMAND ${PYTHON_EXECUTABLE} ${rosidl_generator_py_BIN}
   --generator-arguments-file "${generator_arguments_file}"
@@ -177,105 +175,93 @@ else()
     ${_generated_msg_py_files}
     ${_generated_msg_c_files}
     ${_generated_srv_files}
+    ${_generated_extension_files}
+    ${_extension_dependencies}
+    ${rosidl_generate_interfaces_TARGET}__rosidl_generator_c
   )
 endif()
 
 macro(set_properties _build_type)
-  set_target_properties(${_msg_name}${_pyext_suffix} PROPERTIES
+  set_target_properties(${_target_name} PROPERTIES
     COMPILE_FLAGS "${_extension_compile_flags}"
     PREFIX ""
-    LIBRARY_OUTPUT_DIRECTORY${_build_type} "${_output_path}/${_parent_folder}"
-    RUNTIME_OUTPUT_DIRECTORY${_build_type} "${_output_path}/${_parent_folder}"
-    OUTPUT_NAME "${_base_msg_name}__${_typesupport_impl}${PythonExtra_EXTENSION_SUFFIX}"
+    LIBRARY_OUTPUT_DIRECTORY${_build_type} ${_output_path}
+    RUNTIME_OUTPUT_DIRECTORY${_build_type} ${_output_path}
+    OUTPUT_NAME "${PROJECT_NAME}_s__${_typesupport_impl}${PythonExtra_EXTENSION_SUFFIX}"
     SUFFIX "${PythonExtra_EXTENSION_EXTENSION}")
 endmacro()
 
-foreach(_generated_msg_c_ts_file ${_generated_msg_c_ts_files})
-  get_filename_component(_full_folder "${_generated_msg_c_ts_file}" DIRECTORY)
-  get_filename_component(_parent_folder "${_full_folder}" NAME)
-  get_filename_component(_base_msg_name "${_generated_msg_c_ts_file}" NAME_WE)
-  get_filename_component(_full_extension_msg_name "${_generated_msg_c_ts_file}" EXT)
+foreach(_typesupport_impl ${_typesupport_impls})
+  find_package(${_typesupport_impl} REQUIRED)
 
-  set(_msg_name "${_base_msg_name}${_full_extension_msg_name}")
+  set(_pyext_suffix "__pyext")
+  set(_target_name "${PROJECT_NAME}__${_typesupport_impl}${_pyext_suffix}")
 
-  # foreach(_typesupport_impl ${_typesupport_impls})
-    list(FIND _generated_msg_c_ts_files ${_generated_msg_c_ts_file} _file_index)
-    list(GET _type_support_by_generated_msg_c_files ${_file_index} _typesupport_impl)
-    find_package(${_typesupport_impl} REQUIRED)
-    set(_generated_msg_c_common_file "${_full_folder}/${_base_msg_name}.c")
+  add_library(${_target_name} SHARED
+    ${_generated_msg_c_ts_${_typesupport_impl}_files}
+    ${_generated_msg_c_common_files}
+  )
 
-    set(_pyext_suffix "__pyext")
+  add_dependencies(
+    ${_target_name}
+    ${rosidl_generate_interfaces_TARGET}${_target_suffix}
+    ${rosidl_generate_interfaces_TARGET}__rosidl_generator_c
+  )
 
-    # TODO(esteve): Change the following code so that each file is compiled independently
-    add_library(${_msg_name}${_pyext_suffix} SHARED
-      "${_generated_msg_c_ts_file}"
-      "${_generated_msg_c_common_file}"
+  set(_extension_compile_flags "")
+  if(NOT WIN32)
+    set(_extension_compile_flags "-Wall -Wextra")
+  endif()
+  set_properties("")
+  if(WIN32)
+    set_properties("_DEBUG")
+    set_properties("_MINSIZEREL")
+    set_properties("_RELEASE")
+    set_properties("_RELWITHDEBINFO")
+  endif()
+  target_link_libraries(
+    ${_target_name}
+    ${PythonExtra_LIBRARIES}
+    ${PROJECT_NAME}__${_typesupport_impl}
+  )
+
+  list(APPEND _generated_extension_files
+    "$<TARGET_FILE:${_target_name}>"
+  )
+
+  target_include_directories(${_target_name}
+    PUBLIC
+    ${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_c
+    ${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py
+    ${PythonExtra_INCLUDE_DIRS}
+  )
+
+  foreach(_pkg_name ${rosidl_generate_interfaces_DEPENDENCY_PACKAGE_NAMES})
+    ament_target_dependencies(${_target_name}
+      ${_pkg_name}
     )
+  endforeach()
+  ament_target_dependencies(${_target_name}
+    "rosidl_generator_c"
+    "${_typesupport_impl}"
+  )
 
-    set_properties("")
-    if(WIN32)
-      set_properties("_DEBUG")
-      set_properties("_MINSIZEREL")
-      set_properties("_RELEASE")
-      set_properties("_RELWITHDEBINFO")
-    endif()
+  list(APPEND _extension_dependencies ${_target_name})
 
-    add_dependencies(
-      ${_msg_name}${_pyext_suffix}
-      ${rosidl_generate_interfaces_TARGET}${_target_suffix}
-      ${rosidl_generate_interfaces_TARGET}__rosidl_generator_c
-    )
-
-    set(_extension_compile_flags "")
-    if(NOT WIN32)
-      set(_extension_compile_flags "-Wall -Wextra")
-    endif()
-
-    target_link_libraries(
-      ${_msg_name}${_pyext_suffix}
-      ${PythonExtra_LIBRARIES}
-      ${PROJECT_NAME}__${_typesupport_impl}
-    )
-
-    list(APPEND _generated_extension_files
-      "$<TARGET_FILE:${_msg_name}${_pyext_suffix}>"
-    )
-
-    target_include_directories(${_msg_name}${_pyext_suffix}
-      PUBLIC
-      ${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_c
-      ${PythonExtra_INCLUDE_DIRS}
-    )
-
-    foreach(_pkg_name ${rosidl_generate_interfaces_DEPENDENCY_PACKAGE_NAMES})
-      ament_target_dependencies(
-        ${_msg_name}${_pyext_suffix}
-        ${_pkg_name}
-      )
-    endforeach()
-    ament_target_dependencies(${_msg_name}${_pyext_suffix}
-      "rosidl_generator_c"
-      "${_typesupport_impl}"
-    )
-
-    list(APPEND _extension_dependencies ${_msg_name}${_pyext_suffix})
-
-    ament_target_dependencies(${_msg_name}${_pyext_suffix}
-      ${_typesupport_impl}
-    )
-    add_dependencies(${_msg_name}${_pyext_suffix}
-      ${rosidl_generate_interfaces_TARGET}__${_typesupport_impl}
-    )
-
-  # endforeach()
-  ament_target_dependencies(${_msg_name}${_pyext_suffix}
+  ament_target_dependencies(${_target_name}
+    ${_typesupport_impl}
+  )
+  add_dependencies(${_target_name}
+    ${rosidl_generate_interfaces_TARGET}__${_typesupport_impl}
+  )
+  ament_target_dependencies(${_target_name}
     "rosidl_generator_c"
     "rosidl_generator_py"
     "${PROJECT_NAME}__rosidl_generator_c"
   )
 
   if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
-    install(TARGETS ${_msg_name}${_pyext_suffix}
-      DESTINATION "${PYTHON_INSTALL_DIR}/${PROJECT_NAME}/${_msg_package_dir2}")
+    install(TARGETS ${_target_name}
+      DESTINATION "${PYTHON_INSTALL_DIR}/${PROJECT_NAME}")
   endif()
 endforeach()
--- a/rosidl_generator_py/resource/_msg_support.c.template
+++ b/rosidl_generator_py/resource/_msg_support.c.template
@@ -21,13 +21,11 @@ have_not_included_string = True
 
 @[end if]@
 @[end for]@
-
 @{
 msg_typename = '%s__%s__%s' % (spec.base_type.pkg_name, subfolder, spec.base_type.type)
 }@
 
-/* TODO(esteve): implement conversion function */
-void * @(spec.base_type.pkg_name)_@(module_name)__convert_from_py(PyObject *_pymsg)
+void * @(spec.base_type.pkg_name)_@(module_name)__convert_from_py(PyObject * _pymsg)
 {
   @(msg_typename) * ros_message = @(msg_typename)__create();
   (void)ros_message;
@@ -36,7 +34,6 @@ full_classname = "@(spec.base_type.pkg_name).@(subfolder)._@(module_name).@(spec
 }@
   char full_classname_dest[@(len(full_classname) + 1)];
 
-  /* TODO(esteve): assume 1BYTE encoding for unicode strings. */
   char * class_name = (char *)PyUnicode_1BYTE_DATA(
     PyObject_GetAttrString(PyObject_GetAttrString(_pymsg, "__class__"), "__name__"));
   char * module_name = (char *)PyUnicode_1BYTE_DATA(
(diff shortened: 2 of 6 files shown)


mikaelarguedas:
Talker/listener works for a variety of messages but not all, investigating now

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -76,17 +79,17 @@ endforeach()
 
 if(NOT "${_generated_msg_py_files} " STREQUAL " ")
   list(GET _generated_msg_py_files 0 _msg_file)
-  get_filename_component(_parent_folder "${_msg_file}" DIRECTORY)
+  get_filename_component(_parent_folder2 "${_msg_file}" DIRECTORY)
   list(APPEND _generated_msg_py_files
-    "${_parent_folder}/__init__.py"
+    "${_parent_folder2}/__init__.py"
```
dirk-thomas: These renames seem to be unnecessary? Same below.

mikaelarguedas: I added them because in one case parent_folder is a full path and in the other case it's just the name and in the other case it was the path of the parent folder. Before esteve was not looping over the type support so he could afford to rewrite the variable. But you're right the variable name is confusing and I will rename them to something more explicit

mikaelarguedas:
Two failure use cases:
Uninitialized string in message result in an empty string, thus not ending by '\0' so failure at the C typesupport level
Failure with arrays, should be fixed here in templates

mikaelarguedas:
CI jobs:
http://ci.ros2.org/job/ci_linux/1168/ SUCCESS
http://ci.ros2.org/job/ci_osx/943 FAIL
http://ci.ros2.org/job/ci_windows/1221/ FAIL

mikaelarguedas:
CI jobs:
http://ci.ros2.org/job/ci_linux/1168/ SUCCESS
http://ci.ros2.org/job/ci_osx/943 FAIL
http://ci.ros2.org/job/ci_windows/1221/ FAIL

mikaelarguedas:
Compiling on windows: yay!
CI jobs:
Windows: http://ci.ros2.org/job/ci_windows/1225
OSX: http://ci.ros2.org/job/ci_osx/945/
Linux: http://ci.ros2.org/job/ci_linux/1169

Still no support for arrays yet.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -177,105 +176,94 @@ else()
...
-  set_target_properties(${_msg_name}${_pyext_suffix} PROPERTIES
+  set_target_properties(${_target_name} PROPERTIES
     COMPILE_FLAGS "${_extension_compile_flags}"
     PREFIX ""
-    LIBRARY_OUTPUT_DIRECTORY${_build_type} "${_output_path}/${_parent_folder}"
-    RUNTIME_OUTPUT_DIRECTORY${_build_type} "${_output_path}/${_parent_folder}"
-    OUTPUT_NAME "${_base_msg_name}__${_typesupport_impl}${PythonExtra_EXTENSION_SUFFIX}"
+    LIBRARY_OUTPUT_DIRECTORY${_build_type} ${_output_path}
+    RUNTIME_OUTPUT_DIRECTORY${_build_type} ${_output_path}
+    OUTPUT_NAME "${PROJECT_NAME}_s__${_typesupport_impl}${PythonExtra_EXTENSION_SUFFIX}"
```
jacquelinekay: what's up with the `_s__`?

rosidl_generator_py/resource/_msg_support.c.template
```diff
@@ -52,10 +48,22 @@ full_classname = "@(spec.base_type.pkg_name).@(subfolder)._@(module_name).@(spec
...
   (void)py@(field.name);
 @[if not field.type.is_primitive_type()]@
-// TODO(esteve): add support for nested types
-@[end if]@
-
+@{
+nested_type = '%s__%s__%s' % (field.type.pkg_name,'msg',field.type.type)
+}@
 @[if field.type.is_array]@
+  // TODO implement nested array
```
jacquelinekay: can you add TODO(<username>) here and elsewhere? It's a small style thing to make sure someone is associated with a hanging issue, rather than just letting issues hidden in the source code disappear. This normally is caught by cpplint, but the generated code doesn't get linted.

rosidl_generator_py/resource/_msg_support.c.template
```diff
@@ -123,11 +131,23 @@ PyObject * @(spec.base_type.pkg_name)_@(module_name)__convert_to_py(void * raw_r
...
+  assert(PySequence_Check(py@(field.name)));
+@[else]@
+  PyObject * py@(field.name)_msg_module = PyImport_ImportModule("@(field.type.pkg_name).msg._@convert_camel_case_to_lower_case_underscore(field.type.type)");
+  PyObject* py@(field.name)_convert_to_py = PyObject_GetAttrString(py@(field.name)_msg_module, "_convert_to_py");
+  typedef PyObject *(* convert_to_py_signature)(void *);
+  convert_to_py_signature convert_to_py_@(field.name) = (convert_to_py_signature)PyCapsule_GetPointer(py@(field.name)_convert_to_py,NULL);
+  @(nested_type) pytmp@(field.name) = ros_message->@(field.name);
+  (void) pytmp@(field.name);
```
jacquelinekay: do you need this line if `pytmp@(field.name)` is used in the next line?

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
mikaelarguedas: It seems that this suffix is used everywhere in rosidl_generator_py for the "support" related files (generated c file names, generated library). I didn't know the reason it was there in the first place so I just kept it. But I won't be against removing it everywhere if it can shorten these super long library names.

rosidl_generator_py/resource/_msg_support.c.template
mikaelarguedas: If it's not casted as a void I think the compiler gets angry. But I can double check

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
jacquelinekay: As long as it doesn't collide with the name of another target, I would also support removing it.

jacquelinekay:
Where does the new nested message functionality get tested? Are you testing it locally or do you have something to add to CI? I see that Nested types are already in `test_interfaces.py` of this package but apparently this test does not actually cover the conversion functions since it was passing before:

https://github.com/ros2/rosidl/blob/master/rosidl_generator_py/test/test_interfaces.py

I suppose you could add a nested type to the tests of `rclpy`. Maybe we could make a talker/listeener test fixture that checks different message types following this pattern:

https://github.com/ros2/rclpy/blob/master/rclpy/test/test_talker_listener.py

Besides tests, my style nitpicks, and the warnings on the OSX build, this is looking good to me.

mikaelarguedas:
Yes right now I test them locally with custom talker_py and listener_py.
I'm planning on providing a talker/listener test testing separately nested messages and typesupports before submit this for review.

rosidl_generator_py/resource/_msg_support.c.template
mikaelarguedas: My bad it's a leftover from testing when next line was commented. Removing it

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
mikaelarguedas: I'll give it a try on windows because I think the cmake variable ${PythonExtra_EXTENSION_SUFFIX} is not populated on windows and it may very likely end up colliding with another target

mikaelarguedas: Confirmed, if we remove it, it would break Windows build.
[PEP3149](https://www.python.org/dev/peps/pep-3149/#windows) doesn't define any official suffix for Windows. We could either chose an arbitrary one for windows and remove the `_s` or leave it as is

esteve: Unfortunately, the filename of the Python extension must match the name of the module in https://github.com/ros2/rosidl/blob/add_nested_msg_support/rosidl_generator_py/resource/_msg_support.entry_point.c.template#L32, so you can't a different value for `${PythonExtra_EXTENSION_SUFFIX}` other than the empty string on Windows.

esteve:
@mikaelarguedas I just saw that you included the changes I made in https://github.com/ros2/rosidl/pull/98 I'm sorry I left it stagnant and didn't close it before, I just deleted it.

In any case, that PR didn't seem show any improvement over using one extension per message, at least on my tests. My goal was to reduce the time it took to compile the Python extensions, but there wasn't much of a difference, and the code looked way worse. Have you noticed any improvement in the time it takes to compile the Python extensions?

mikaelarguedas:
@esteve yeah I started all this single extension stuff and I realized that #98 had almost everything I needed. In the end this PR is now only including the changes made in #98 and adding a few lines here and there. We're reaching the ridiculous point of having more commits than line change^^ 
If anything we can say that this PR brings consistency with a single shared library/package/typesupport as it is for c and c++ libs.

Regarding compilation time I didn't notice any significant improvement (didn't measure it neither), it's still super slow. I'll check quantitatively if it makes any difference.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
mikaelarguedas: Thanks for pointing it out! we'll keep the `_s` then.

mikaelarguedas:
POLL:
The last commit allows to properly mark object for garbage collection in python.
I would like some input on the solutions:
huge drawback: 
 * current implementation makes huge code duplication in generated python extension:
number of duplicated lines in a given package if n the number of message in this package:
6n*(6n+1)/2
This will result in a drastic increase of the compile time at some point.
alternative:
 * create nested labels and make the PC jump from a label to the next at each iteration:
`goto label n

labeln:
  do label n stuf
  goto labeln-1

...
label1:
  do label1 stuff
  goto label0
`
This would allow to avoid code duplication. One could argue that nested label is the worst practice in c language programming and another could argue that extensive code duplication within a same file doesn't make any sense.
So let me know what you all think about it and I'll act upon it.

jacquelinekay:
this might be even worse, but maybe you could use macros to reduce the code duplication without use nested labels.

mikaelarguedas:
BTW last CI jobs:
http://ci.ros2.org/job/ci_linux/1183/
http://ci.ros2.org/job/ci_osx/960/
http://ci.ros2.org/job/ci_windows/1238

mikaelarguedas:
@jacquelinekay that's a fair point

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -177,105 +176,94 @@ else()
...
-        ${_pkg_name}
-      )
-    endforeach()
-    ament_target_dependencies(${_msg_name}${_pyext_suffix}
-      "rosidl_generator_c"
-      "${_typesupport_impl}"
+  foreach(_pkg_name ${rosidl_generate_interfaces_DEPENDENCY_PACKAGE_NAMES})
+    ament_target_dependencies(${_target_name}
+      ${_pkg_name}
+      "_${_pkg_name}__${_typesupport_impl}${PythonExtra_EXTENSION_SUFFIX}"
```
wjwwood: See: https://github.com/ros2/rosidl/pull/119/files#r60457523

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -76,17 +75,17 @@ endforeach()
 
 if(NOT "${_generated_msg_py_files} " STREQUAL " ")
   list(GET _generated_msg_py_files 0 _msg_file)
-  get_filename_component(_parent_folder "${_msg_file}" DIRECTORY)
+  get_filename_component(_parent_folder_path "${_msg_file}" DIRECTORY)
   list(APPEND _generated_msg_py_files
-    "${_parent_folder}/__init__.py"
+    "${_parent_folder_path}/__init__.py"
```
dirk-thomas: If not necessary please avoid changing variable names like this since it make reading the diff more difficult since the reader needs to separate essential and non-essential changes.

Same a few lines below.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -151,14 +150,14 @@ if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
   install(FILES ${_generated_msg_py_files}
     DESTINATION "${PYTHON_INSTALL_DIR}/${PROJECT_NAME}/${_msg_package_dir2}"
   )
+
```
dirk-thomas: Same for white space only changes.

rosidl_generator_py/resource/_msg_support.entry_point.c.template
```diff
@@ -1,85 +1,136 @@
...
-            pyconvert_to_py = NULL;
-            Py_XDECREF(m);
-            m = NULL;
-          }
-        }
-      }
-    } else {
+    pyconvert_from_py_@(module_name) = PyCapsule_New(
+                                          (void *)&@(spec.base_type.pkg_name)_@(module_name)__convert_from_py,
+                                          NULL, NULL);
```
wjwwood: Please only do one indention from the original line when you do this, e.g.:

```
pyconvert_from_py_@(module_name) = PyCapsule_New(
  (void *)&@(spec.base_type.pkg_name)_@(module_name)__convert_from_py,
  NULL, NULL);
```

Or:

```
pyconvert_from_py_@(module_name) = PyCapsule_New(
  (void *)&@(spec.base_type.pkg_name)_@(module_name)__convert_from_py,
  NULL, NULL
);
```

rosidl_generator_py/resource/_msg_support.entry_point.c.template
```diff
@@ -1,85 +1,136 @@
...
+      DECREF_@(lastobject.upper())
+      Py_XDECREF(m);
+      m = NULL;
+      return m;
+    }
+
+    PyObject *pyconvert_to_py_@(module_name) = NULL;
+    pyconvert_to_py_@(module_name) = PyCapsule_New(
+                                        (void *)&@(spec.base_type.pkg_name)_@(module_name)__convert_to_py,
+                                        NULL, NULL);
```
wjwwood: same

rosidl_generator_py/resource/_msg_support.entry_point.c.template
```diff
@@ -1,85 +1,136 @@
...
+    if (PyModule_AddObject(m, "convert_to_py_@(module_name)", pyconvert_to_py_@(module_name))) {
+      DECREF_@(lastobject.upper())
+      Py_XDECREF(m);
+      m = NULL;
+      return m;
+    }
+    PyObject *pytype_support_@(module_name) = NULL;
+    pytype_support_@(module_name) = PyCapsule_New(
+                                      (void *)ROSIDL_GET_TYPE_SUPPORT(@(spec.base_type.pkg_name), @(subfolder), @(spec.msg_name)),
+                                    NULL, NULL);
```
wjwwood: same here, also inconsistent. You might want to run our linter over one of the generated files just to see what issues come up. You don't have to have the generated files address all the linter issues (line length is particularly hard to guarantee), but I would appreciate it if you can fix some of the other style issues the linter reports. It might also help you catch non-style issues if you run `cppcheck` over it, even if it's only once.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -177,105 +176,94 @@ else()
...
-
-    set(_pyext_suffix "__pyext")
-
-    # TODO(esteve): Change the following code so that each file is compiled independently
-    add_library(${_msg_name}${_pyext_suffix} SHARED
-      "${_generated_msg_c_ts_file}"
-      "${_generated_msg_c_common_file}"
-    )
+  set(_pyext_suffix "__pyext")
+  set(_target_name "${PROJECT_NAME}${_typesupport_impl}${_pyext_suffix}")
```
dirk-thomas: You might want to use a separator between the project name and the type support name (e.g. double underscore).

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
mikaelarguedas: You're right, not necessary anymore.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
mikaelarguedas: Sorry, another artefact that hasn't been removed

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
mikaelarguedas: noted

rosidl_generator_py/resource/_msg_support.c.template
```diff
@@ -2,7 +2,6 @@
 
 #include <@(spec.base_type.pkg_name)/@(subfolder)/@(module_name)__struct.h>
 #include <@(spec.base_type.pkg_name)/@(subfolder)/@(module_name)__functions.h>
-
```
dirk-thomas: Please don't remove these empty lines. The make the generated code more structured and therefore readable.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -92,6 +99,21 @@ def generate_py(generator_arguments_file, typesupport_impls):
             for module_, type_ in modules[module]:
                 f.write('from %s.msg._%s import %s\n' % (args['package_name'], module_, type_))
 
+    for template_file, generated_filenames in mapping_extension_msgs.items():
+        for generated_filename in generated_filenames:
+            data = {
+                'module_name': module_name, 'package_name': args['package_name'],
```
wjwwood: Please do one dictionary entry per line (I had to double take to notice there were two here).

wjwwood:
Other than the mostly stylistic comments, this lgtm. +1 with passing CI and a squash before merge (if you use the GitHub button to squash make sure the first line of the commit message is descriptive).

rosidl_generator_py/resource/_msg_support.c.template
```diff
@@ -123,11 +130,22 @@ PyObject * @(spec.base_type.pkg_name)_@(module_name)__convert_to_py(void * raw_r
...
+  assert(PySequence_Check(py@(field.name)));
+@[else]@
+  PyObject * py@(field.name)_msg_module = PyImport_ImportModule("@(field.type.pkg_name).msg._@convert_camel_case_to_lower_case_underscore(field.type.type)");
+  PyObject* py@(field.name)_convert_to_py = PyObject_GetAttrString(py@(field.name)_msg_module, "_convert_to_py");
+  typedef PyObject *(* convert_to_py_signature)(void *);
+  convert_to_py_signature convert_to_py_@(field.name) = (convert_to_py_signature)PyCapsule_GetPointer(py@(field.name)_convert_to_py,NULL);
+  @(nested_type) pytmp@(field.name) = ros_message->@(field.name);
+  py@(field.name) = convert_to_py_@(field.name)(&pytmp@(field.name));
+@[end if]@
+@[elif field.type.is_array]@
```
dirk-thomas: Is this template currently working? From reading I can't see to which this `elif` belongs to. But I might just miss it.

dirk-thomas: Nvm. It belongs to the opening `if` in line 132...

rosidl_generator_py/resource/_msg_support.entry_point.c.template
```diff
@@ -1,85 +1,136 @@
...
+  module_name = convert_camel_case_to_lower_case_underscore(type_name)
+  pymodulelist.insert(0, 'pyconvert_from_py_' + module_name)
+  pymodulelist.insert(0, 'pyconvert_to_py_' + module_name)
+  pymodulelist.insert(0, 'pytype_support_' + module_name)
+listlen = len(pymodulelist)
+lastobject = ''
+}@
+
+@[for i in range(listlen)]@
+@[if  i != listlen-1]@
```
dirk-thomas: Spaces around operators.

rosidl_generator_py/resource/_msg_support.entry_point.c.template
```diff
@@ -1,85 +1,136 @@
...
-      if (PyModule_AddObject(m, "convert_from_py", pyconvert_from_py)) {
-        Py_XDECREF(pyconvert_from_py);
-        pyconvert_from_py = NULL;
-        Py_XDECREF(m);
-        m = NULL;
-      }
+  m = PyModule_Create(&@(package_name)__module);
+  if (m == NULL) {
+    return m;
+  }
```
dirk-thomas: I think making the return value more explicit helps reading the code:

```
if (!m) {
  return NULL;
}
```

rosidl_generator_py/resource/_msg_support.entry_point.c.template
```diff
@@ -1,85 +1,136 @@
...
-void * @(spec.base_type.pkg_name)_@(module_name)__convert_from_py(PyObject *_pymsg);
-
-/* TODO(esteve): implement conversion function */
-PyObject * @(spec.base_type.pkg_name)_@(module_name)__convert_to_py(void * raw_ros_message);
-
 PyMODINIT_FUNC
-PyInit__@(module_name)_s__@(typesupport_impl)(void)
+PyInit_@(package_name)_s__@(typesupport_impl)(void)
 {
   PyObject *m = NULL;
```
dirk-thomas: Space after the star.

rosidl_generator_py/resource/_msg_support.entry_point.c.template
```diff
@@ -1,85 +1,136 @@
...
-      }
-    } else {
+    pyconvert_from_py_@(module_name) = PyCapsule_New(
+                                          (void *)&@(spec.base_type.pkg_name)_@(module_name)__convert_from_py,
+                                          NULL, NULL);
+    if (pyconvert_from_py_@(module_name) == NULL) {
+      DECREF_@(lastobject.upper())
       Py_XDECREF(m);
       m = NULL;
+      return m;
```
dirk-thomas: Similar as above. Since `m` is a local variable it doesn't need to be set to `NULL`. But `return NULL` makes it more explicit.

Same below.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -52,15 +56,18 @@ def generate_py(generator_arguments_file, typesupport_impls):
...
     }
     latest_target_timestamp = get_newest_modification_time(args['target_dependencies'])
 
     modules = defaultdict(list)
+    message_specs = []
     for ros_interface_file in args['ros_interface_files']:
         extension = os.path.splitext(ros_interface_file)[1]
         subfolder = os.path.basename(os.path.dirname(ros_interface_file))
-        if extension == '.msg':
+        if extension == '.msg' and subfolder == 'msg':
```
dirk-thomas: What happens to the request and response messages of services if they are being excluded here?

wjwwood: Python doesn't support Services. Perhaps a TODO?

rosidl_generator_py/resource/_msg_support.entry_point.c.template
mikaelarguedas: It makes sense. I didn't see any pep7 linter so I didn't know what I was supposed to run. I'll use `cppcheck`/`cpplint` from now on.

rosidl_generator_py/resource/_msg_support.entry_point.c.template
mikaelarguedas: Should we update the [Guidelines](https://github.com/ros2/ros2/wiki/Developer-Guide#c-1) to make it more obvious that we should use a single indent and not align on the opened bracket? I guess my more general question is in which case should I follow the c++ guidelines and in which case should I follow pep7 ? (number of space indent, wrapping etc).

dirk-thomas: The current guidelines you linked to already state:

> When a function call cannot fit on one line, wrap at the open parenthesis (not in between arguments) and start them on the next line with a 2-space indent. Continue with the 2-space indent on subsequent lines for more arguments.

Please feel free to improve them if any case is not clear based on that.

wjwwood: Yeah, if you think there's an appropriate place to put that in the guidelines, I think it would help. I suppose this is an implicitly stated rule. There are several examples in the guidelines (for C++ at least) which only include a single indentation level after a line which ends with a scoping symbol.

For example, I consider these two cases equivalent when it comes to indentation:

```c++
struct foo_t foo = {
  0,
  1,
  2,
}

{
  auto ret = func(
    param1,
    param2
  );
}
```

Aligning with the opening paranthesis only makes sense (in my opinion) when you have at least one parameter on the original line as so:

```c++
struct foo_t foo = {0,
                    1,
                    2}

{
  auto ret = func(param1,
                  param2);
}
```

But I think we decided as a group that this syntax isn't great because when you have very long original lines you end up with lots of wasted indentation space, like this:

```c++
auto ret = a_really_really_long_function_name_that_sucks(param1,
                                                         param2)
```

If the guidelines don't capture some of all of this then I can help you update it. If you disagree or have more questions we can talk about it more as well 😄. We should all be on the same page if possible.

As for which tool to use for linting, I think we just use uncrustify and cpplint for the C files, even though we say PEP7. I think we either need to clarify this in the guidelines or find a PEP7 checker. Either way, the indentation stuff we're talking about here applies sort of generically to all languages (including C++ and Python).

mikaelarguedas: The C++ guidelines are pretty clear in that matter. I just didn't know if I
was supposed to follow PEP7 or the C++ guidelines.
I honestly am not a big fan of PEP7 for several reasons so I'm totally fine
with following the current C++ guidelines when writing C code.

I just wanted to know if it was an actual rule we should all follow, and if
so maybe state it in the C section of the guidelines.

mikaelarguedas:
Set of CI jobs:
http://ci.ros2.org/job/ci_linux/1213/
http://ci.ros2.org/job/ci_osx/980/
http://ci.ros2.org/job/ci_windows/1258/

Forgot to add the TODO for service messages generation (that were removed from compilation time purposes), adding it now

mikaelarguedas:
I addressed all previous comments in https://github.com/ros2/rosidl/pull/113/commits/240b246ac6d7372d232bd9eee25751af47f3fba8.

Any other comment on this PR?

NOTE: I will test the `ament_target_dependencies` lack of error message when adding a non ament package to the dependency list on a separate branch given that it's not used on this one anymore

rosidl_generator_py/resource/_msg_support.entry_point.c.template
```diff
@@ -1,85 +1,130 @@
...
+@[end for]@
+
+@{
+pymodulelist = []
+for spec, subfolder in message_specs:
+  type_name = spec.base_type.type
+  module_name = convert_camel_case_to_lower_case_underscore(type_name)
+  pymodulelist.insert(0, 'pyconvert_from_py_' + module_name)
+  pymodulelist.insert(0, 'pyconvert_to_py_' + module_name)
+  pymodulelist.insert(0, 'pytype_support_' + module_name)
```
dirk-thomas: Please use four space indentation for Python code.

dirk-thomas:
Beside the last comment it looks good to me.

mikaelarguedas:
Done

dirk-thomas:
This change seems to result in several compiler warnings in the nightly packaging job: http://ci.ros2.org/job/packaging_linux/229/warnings17Result/

If you don't need the branch anymore you might want to delete it.

mikaelarguedas:
I'll investigate the compiler warnings and delete the branch once it's solved. Thanks for pointing it out

PR #115, opened 2016-04-06
Get list of typesupports from ament index

jacquelinekay:
Connects to ros2/rmw#63

Use `get_rmw_typesupport_c` instead of a hardcoded list of typesupport implementations.

There is still a hardcoded ignore rule for fastrtps.

There is also a hardcoded map in `import_type_support_impl.py`. I wasn't sure how to fix this but perhaps `ament_index_python` could be used?


jacquelinekay:
I'm going to address the hardcoded identifier map in a separate pull request.

rosidl_generator_py/rosidl_generator_py/import_type_support_impl.py
```diff
@@ -44,9 +40,10 @@ def import_type_support(pkg_name, subfolder, rosidl_name, rmw_implementation):
     :param rmw_implementation str: name of the rmw implementation
     :returns: the type support Python module for this specific rosidl and rmw implementation pair
     """
-    if rmw_implementation not in type_support_map.keys():
+    if not ament_index_python.has_resource("rmw_typesupport_c", rmw_implementation):
```
dirk-thomas: Single quotes.

Same below.

dirk-thomas:
I removed an extra empty line in ae15542770f8bd57db4cb66e40f20c3be05da0d8

PR #117, opened 2016-04-15
fix hardcoded rmw impl names

jacquelinekay:
The build is broken right now because I changed the identifier names. After the register language of typesupport branches are merged (see ros2/rmw#63) this map can be replaced with the ament index.


wjwwood:
lgtm

PR #119, opened 2016-04-19
add array support

mikaelarguedas:
Connects to https://github.com/ros2/rosidl/issues/114
Relies on https://github.com/ros2/rosidl/pull/113

### DIFF
--- a/rosidl_generator_py/resource/_msg_support.c.template
+++ b/rosidl_generator_py/resource/_msg_support.c.template
@@ -86,7 +86,7 @@ nested_type = '%s__%s__%s' % (field.type.pkg_name, 'msg', field.type.type)
   }
 @[      if field.type.array_size is None]@
   if (!@(nested_type)__Array__init(&(ros_message->@(field.name)), size@(field.name))) {
-    assert(false && "unable to create array ros_message");
+    PyErr_SetString(PyExc_RuntimeError, "unable to create @(nested_type)__Array ros_message");
   }
   memcpy(ros_message->@(field.name).data, tmparray@(field.name), sizeof(@(nested_type)) * size@(field.name));
 @[      else]@
@@ -155,11 +155,11 @@ nested_type = '%s__%s__%s' % (field.type.pkg_name, 'msg', field.type.type)
 @[    if field.type.array_size is None]@
 @[      if field.type.type == 'string']@
   if (!rosidl_generator_c__String__Array__init(&(ros_message->@(field.name)), size@(field.name))) {
-    assert(false && "unable to create array ros_message");
+    PyErr_SetString(PyExc_RuntimeError, "unable to create String__Array ros_message");
   }
 @[      else]@
   if (!rosidl_generator_c__@(field.type.type)__Array__init(&(ros_message->@(field.name)), size@(field.name))) {
-    assert(false && "unable to create array ros_message");
+    PyErr_SetString(PyExc_RuntimeError, "unable to create @(field.type.type)__Array ros_message");
   }
 @[      end if]@
   memcpy(ros_message->@(field.name).data, tmparray@(field.name), sizeof(@primitive_msg_type_to_c(field.type.type)) * size@(field.name));
--- a/rosidl_generator_py/resource/_msg_support.c.template
+++ b/rosidl_generator_py/resource/_msg_support.c.template
@@ -6,21 +6,34 @@
 @{
 have_not_included_primitive_arrays = True
 have_not_included_string = True
+nested_array_dict = {}
 }@
 @[for field in spec.fields]@
-@[if field.type.is_array and have_not_included_primitive_arrays]@
+@[  if field.type.is_array and have_not_included_primitive_arrays]@
 @{have_not_included_primitive_arrays = False}@
 #include <rosidl_generator_c/primitives_array.h>
 #include <rosidl_generator_c/primitives_array_functions.h>
 
-@[end if]@
-@[if field.type.type == 'string' and have_not_included_string]@
+@[  end if]@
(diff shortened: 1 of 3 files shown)


rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -177,105 +176,94 @@ else()
...
-        ${_pkg_name}
-      )
-    endforeach()
-    ament_target_dependencies(${_msg_name}${_pyext_suffix}
-      "rosidl_generator_c"
-      "${_typesupport_impl}"
+  foreach(_pkg_name ${rosidl_generate_interfaces_DEPENDENCY_PACKAGE_NAMES})
+    ament_target_dependencies(${_target_name}
+      ${_pkg_name}
+      "_${_pkg_name}__${_typesupport_impl}${PythonExtra_EXTENSION_SUFFIX}"
```
wjwwood: Unless I'm mistaken the arguments to `ament_target_dependencies` must be a package (which has been `find_package`'ed): https://github.com/ament/ament_cmake/blob/c1d3004db7cb33ded7aed38e00c9a9d888986309/ament_cmake_target_dependencies/cmake/ament_target_dependencies.cmake#L15-L29

So putting what I believe is a library name here doesn't make sense to me. Maybe you meant to pass this library name to `target_link_libraries` for `_target_name`?

dirk-thomas: Running the code should also result in this error message: https://github.com/ament/ament_cmake/blob/c1d3004db7cb33ded7aed38e00c9a9d888986309/ament_cmake_target_dependencies/cmake/ament_target_dependencies.cmake#L40

rosidl_generator_py/resource/_msg_support.c.template
```diff
@@ -49,14 +58,109 @@ full_classname = "@(spec.base_type.pkg_name).@(subfolder)._@(module_name).@(spec
...
+  PyObject* item@(field.name);
+  size_t idx@(field.name);
+@[if field.type.array_size is None]@
+  size_t size@(field.name) = PySequence_Size(py@(field.name));
+@[else]@
+  size_t size@(field.name) = @(field.type.array_size);
+@[end if]@
+  @(nested_type)* array@(field.name)[size@(field.name)];
+  for(idx@(field.name) = 0; idx@(field.name) < size@(field.name); idx@(field.name)++){
+    item@(field.name) = PySequence_Fast_GET_ITEM(seq@(field.name),idx@(field.name));
```
wjwwood: Please follow `,`'s with a space.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
mikaelarguedas: Thanks for noticing it! I think it's actually an artefact from when I was trying to link the generated libraries together I must have forgotten to remove it.
Surprisingly I didnt have an error when running cmake.

To avoid duplicating work on your side, could you comment on this PR https://github.com/ros2/rosidl/pull/113 ?
All the cmake stuff has been changed there and the current PR is just rebased on top of #113.
So once #113 is reviewed, reviewing #119 is going to be much more straight forward.

dirk-thomas: Please try to add debug message in there to check why you don't get an error message. It would be good to ensure that the API works as intended even if you don't need it for this change.

mikaelarguedas:
Now compiles on windows but rclpy tests are failing. http://ci.ros2.org/job/ci_windows/1263/

mikaelarguedas:
http://ci.ros2.org/job/ci_osx/986/
http://ci.ros2.org/job/ci_linux/1221/
http://ci.ros2.org/job/ci_windows/1268/
rclpy original tests not failing anymore, but seems like new tests are failing on windows only.

dirk-thomas:
The type handling code is pretty tough to read but it looks good to me.

rosidl_generator_py/resource/_msg_support.c.template
```diff
@@ -42,73 +55,161 @@ full_classname = "@(spec.base_type.pkg_name).@(subfolder)._@(module_name).@(spec
...
+  @(nested_type) * tmparray@(field.name) = (@(nested_type) *)malloc(sizeof(@(nested_type)) * size@(field.name));
+  size_t idx@(field.name);
+  for (idx@(field.name) = 0; idx@(field.name) < size@(field.name); idx@(field.name)++) {
+    item@(field.name) = (@(nested_type) *) convert_from_py_@(field.name)(
+      PySequence_Fast_GET_ITEM(seq@(field.name), idx@(field.name)));
+    tmparray@(field.name)[idx@(field.name)] = *item@(field.name);
+  }
+@[      if field.type.array_size is None]@
+  if (!@(nested_type)__Array__init(&(ros_message->@(field.name)), size@(field.name))) {
+    assert(false && "unable to create array ros_message");
```
wjwwood: This probably shouldn't use an assert, since it will not be checked in Release mode.

wjwwood:
Other than auditing your use of `assert` (replace if it should check in both release and debug) the changes lgtm.

What's the test coverage like on these new changes? Do we have test messages in place for this and are we exercising them on this code?

wjwwood:
I guess https://github.com/ros2/rclpy/pull/23 adds testing of this.

mikaelarguedas:
Done, now raising `RuntimeError` if arrays can't be initialized.
And yes the associated tests are in ros2/rclpy#23

wjwwood:
lgtm

PR #120, opened 2016-04-21
fix PEP257 warning D211

dirk-thomas:
Since we don't want to follow D203 (ament/ament_lint#47) I assume we will follow D211 (which is checked by newer versions of `pydocstyle`).


wjwwood:
lgtm

PR #122, opened 2016-04-27
Resize for C arrays

jacquelinekay:
This is a work in progress. Here are some issues I thought of while working on this. Tagging @wjwwood since he wrote `rosidl_generator_c` originally, in case he finds this thought-provoking.

It would be convenient to have a "resize" function for `rosidl_generator_c` arrays, especially since unbounded arrays are initialized to have a size of zero.

The Parameter API uses unbounded arrays of `ParameterValue` messages, so I was motivated to look into this while sketching out use cases for C parameters.

The most basic case of "resize" simply calls `fini` then `init` on an existing Array message structure with a new `size` parameter. You can think of another case where the existing data is copied into the newly allocated data but I think it's okay not to implement that for now.

It is consistent with the intention of the design of the message interfaces to make `resize` only available for unbounded arrays.

However the existing array struct definitions do not distinguish between bounded and unbounded arrays.

This initial implementation introduces a very brittle and hackish pattern in which:

If message A has a subfield B where B is an unbounded array, then the header for message A declares a resize method for message B.
All messages define a "resize" method in their .c file.

This prototype compiles, but my question is: is this pattern OK? Should we instead move towards an implementation that distinguishes Bounded and Unbounded arrays (then bounded arrays could enforce a limit on their capacities)? The bounded/unbounded/static definition seems poorly defined in both `rosidl_generator_c` and `rosidl_generator_cpp`.

As a side note, `resize` implicitly calls `malloc` and free because that's what `init` and `fini` use. Should `rosidl_generator_c` have an allocator interface like rcl? I'm ok with not addressing this right now, but it's something to consider.

### DIFF
--- a/rosidl_generator_c/resource/msg__functions.c.template
+++ b/rosidl_generator_c/resource/msg__functions.c.template
@@ -211,14 +211,14 @@ void
 @# array functions
 @#######################################################################
 bool
-@(array_typename)__init(@(array_typename) * array, size_t size, size_t capacity)
+@(array_typename)__init(@(array_typename) * array, size_t size)
 {
   if (!array) {
     return false;
   }
   @(msg_typename) * data = NULL;
   if (size) {
-    data = (@(msg_typename) *)malloc(sizeof(@(msg_typename)) * capacity);
+    data = (@(msg_typename) *)malloc(sizeof(@(msg_typename)) * size);
     if (!data) {
       return false;
     }
@@ -238,13 +238,10 @@ bool
       free(data);
       return false;
     }
-    for (i = size; i < capacity; ++i) {
-      data[i] = {0};  // ???
-    }
   }
   array->data = data;
   array->size = size;
-  array->capacity = capacity;
+  array->capacity = size;
   return true;
 }
 
@@ -255,7 +252,7 @@ void
     return;
   }
   if (array->data) {
-    // ensure that data and capacity values are consistent
+    // ensure that data acapacity values are consistent
     assert(array->capacity > 0);
     // finalize all array elements
     for (size_t i = 0; i < array->capacity; ++i) {
@@ -270,16 +267,6 @@ void
     assert(0 == array->size);
     assert(0 == array->capacity);
   }
-}
-
-bool
-@(array_typename)__resize(@(array_typename) * array, size_t size, size_t capacity)
-{
-  if (!array) {
-    return false;
-  }
-  @(array_typename)__fini(array);
-  return @(array_typename)__init(array, capacity);
 }
 
 @(array_typename) *
@@ -304,4 +291,15 @@ void
     @(array_typename)__fini(array);
   }
   free(array);
+}
+
+bool
+@(array_typename)__resize(@(array_typename) * array, size_t size)
+{
+  if (!array)
+  {
+    return false;
+  }
+  @(array_typename)__fini(array);
+  return @(array_typename)__init(array, size);
 }
--- a/rosidl_generator_c/resource/msg__functions.c.template
+++ b/rosidl_generator_c/resource/msg__functions.c.template
@@ -207,19 +207,18 @@ void
   free(msg);
 }
 
-
 @#######################################################################
 @# array functions
 @#######################################################################
 bool
-@(array_typename)__init(@(array_typename) * array, size_t size)
+@(array_typename)__init(@(array_typename) * array, size_t size, size_t capacity)
 {
   if (!array) {
     return false;
   }
   @(msg_typename) * data = NULL;
   if (size) {
-    data = (@(msg_typename) *)malloc(sizeof(@(msg_typename)) * size);
+    data = (@(msg_typename) *)malloc(sizeof(@(msg_typename)) * capacity);
     if (!data) {
       return false;
     }
@@ -239,10 +238,13 @@ bool
       free(data);
       return false;
     }
+    for (i = size; i < capacity; ++i) {
+      data[i] = {0};  // ???
+    }
   }
   array->data = data;
   array->size = size;
-  array->capacity = size;
+  array->capacity = capacity;
   return true;
 }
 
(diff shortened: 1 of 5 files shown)


rosidl_generator_c/resource/msg__functions.c.template
```diff
@@ -253,7 +252,7 @@ void
     return;
   }
   if (array->data) {
-    // ensure that data and capacity values are consistent
+    // ensure that data acapacity values are consistent
```
wjwwood: typo?

wjwwood:
> The most basic case of "resize" simply calls fini then init on an existing Array message structure with a new size parameter. You can think of another case where the existing data is copied into the newly allocated data but I think it's okay not to implement that for now.

My expectation of a resize function is that it would preserve the existing data unless the new capacity is smaller than the size. So I think it should copy the old data into the new storage (if new storage is required). I think it should also use realloc if it is available (either as a system call or through an allocator interface) which preserves the data in the allocated memory. That brings up a different question, does the resize function reallocate memory if it already has enough capacity? Put another way, does it change the size and the capacity if necessary or does it only change the capacity? Do we need a reserve as well as a resize?

> It is consistent with the intention of the design of the message interfaces to make resize only available for unbounded arrays.

Again this might be a difference in resize versus reserve, but why wouldn't a bounded array be resizeable? If an array is bounded at 1024, for example, then does the capacity have to be 1024 always? Couldn't you have a reserve and/or resize function for bounded arrays that returns an error if you request to reserve or resize to a size greater than the bound?

> If message A has a subfield B where B is an unbounded array, then the header for message A declares a resize method for message B.
All messages define a "resize" method in their .c file.

Why not include the header which defines array operations for the message type? (maybe there isn't one yet, but it seems like we shouldn't define it in multiple places)

> Should `rosidl_generator_c` have an allocator interface like rcl?

Yes, I think @dirk-thomas elected to not do that at the time since we were still trying it out in `rcl`. The intention was to add that to the interface if it worked out in `rcl`.

rosidl_generator_c/resource/msg__functions.h.template
```diff
@@ -86,6 +86,66 @@ ROSIDL_GENERATOR_C_PUBLIC_@(spec.base_type.pkg_name)
...
+#include "rosidl_generator_c/string_functions.h"
+@[  else]
+#include "rosidl_generator_c/primitives_array_functions.h"
+@[  end if]
+@[else]
+@{
+subfield_header_file = '%s/msg/%s__functions.h' % (subfield_pkg_name, get_header_filename_from_msg_name(subfield_type_name))
+}@
+#include "@(subfield_header_file)"
+@[end if]
```
dirk-thomas: All `#include` should be at the top of the file.

rosidl_generator_c/resource/msg__functions.h.template
```diff
@@ -86,6 +86,66 @@ ROSIDL_GENERATOR_C_PUBLIC_@(spec.base_type.pkg_name)
...
+    if field.type.type == 'string':
+        subfield_type_name = 'String'
+    else:
+        subfield_type_name = field.type.type
+
+    subfield_array_typename = '%s__%s__Array' % (subfield_pkg_name, subfield_type_name)
+else:
+    subfield_pkg_name = field.type.pkg_name
+    subfield_type_name = field.type.type
+    subfield_array_typename = '%s__msg__%s__Array' % (subfield_pkg_name, subfield_type_name)
```
dirk-thomas: The Python function `msg_type_to_c` already converts a field type into the C type. I think this should be used here instead.

rosidl_generator_c/resource/msg__functions.h.template
```diff
@@ -86,6 +86,66 @@ ROSIDL_GENERATOR_C_PUBLIC_@(spec.base_type.pkg_name)
 void
 @(msg_typename)__destroy(@(msg_typename) * msg);
 
+@# If the message contains an unbounded non-primitive array subfield, declare a resize function
```
dirk-thomas: Shouldn't the comment be:

> For every non-primitive unbounded array field declare a resize function

rosidl_generator_c/resource/msg__functions.h.template
```diff
@@ -86,6 +86,66 @@ ROSIDL_GENERATOR_C_PUBLIC_@(spec.base_type.pkg_name)
...
+    subfield_array_typename = '%s__%s__Array' % (subfield_pkg_name, subfield_type_name)
+else:
+    subfield_pkg_name = field.type.pkg_name
+    subfield_type_name = field.type.type
+    subfield_array_typename = '%s__msg__%s__Array' % (subfield_pkg_name, subfield_type_name)
+
+subfield_array_decl_guard = '%s__%s__UNBOUNDED_ARRAY__RESIZE_DECL' % (subfield_pkg_name, subfield_type_name)
+}@
+#ifndef @(subfield_array_decl_guard)
+#define @(subfield_array_decl_guard)
```
dirk-thomas: Why is this header guard needed?

rosidl_generator_c/resource/msg__functions.h.template
```diff
@@ -86,6 +86,66 @@ ROSIDL_GENERATOR_C_PUBLIC_@(spec.base_type.pkg_name)
...
+#include "@(subfield_header_file)"
+@[end if]
+
+@[if field.type.is_primitive_type()]@
+ROSIDL_GENERATOR_C_PUBLIC
+@[else]
+ROSIDL_GENERATOR_C_PUBLIC_@(subfield_pkg_name)
+@[end if]
+bool
+@(subfield_array_typename)__resize(@(subfield_array_typename) * array, size_t size);
```
dirk-thomas: I very likely just missed it in the "template war", but where are these defined?

mikaelarguedas:
that would be a pretty cool feature to implement. This would need some rebase and fixups before being able to iterate on

dirk-thomas:
This PR is out of date and can't be merged as is. Since the feature request is tracked in ros2/rosidl#468 I will close this one.

PR #125, opened 2016-05-03
Fix generator being invoked multiple times

### LINKED ISSUE #124: Python generator called repeatedly in message interface packages
The message "Generating Python code for ROS interfaces" is printed repeatedly during compilation of message interface packages. It seems the Python generator is being called repeatedly when it should only be called once.

dirk-thomas:
The reason the custom command was retriggered was that the files being promised to be generated included the Request / Response messages as well as the `__init__` files. But the generator didn't actually generate them.

Fixes #124.


jacquelinekay:
+1. I don't understand every detail but the code looks fine and it fixes the issue for me.

dirk-thomas:
And a follow up commit: 86ad5ad0bda1cd6b3e44261eb0cb50b86ae6026d

PR #126, opened 2016-05-04
rosidl_generator_py:From macros to labels

mikaelarguedas:
This addresses https://github.com/ros2/rosidl/commit/5434124b8e4a80c61e3fc2d0e0c50fb9d4c17d9a#commitcomment-17330447

Relies on #119

PR #127, opened 2016-05-04
refactor generated PyInit function

dirk-thomas:
@mikaelarguedas How does this look?

For each type a separate internal function is generated. The three capsules for each type are generated in a loop.

I only explicitly decrement capsules that are not successfully added to the module since the assumption is that the module will remove all previously added objects when it is being destroyed.

Extending #126.

### DIFF
--- a/rosidl_generator_py/resource/_msg_support.entry_point.c.template
+++ b/rosidl_generator_py/resource/_msg_support.entry_point.c.template
@@ -46,79 +46,64 @@ static struct PyModuleDef @(package_name)__module = {
   NULL,
   NULL,
 };
+@
+@[for spec, subfolder in message_specs]@
+@{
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
looks good to me, you may want to clean up the now unused logic and variables (e.g. `pymodule_list`)

dirk-thomas:
Good point. Gone.

mikaelarguedas:
I don't know if it has it's place in this PR but a test to confirm that there is no memory leak and that all objects are actually DECREFed and collected by GC at module destruction would be sweet. It's kind of a one time test but we would want to test it on all platforms before actually merging it on master.

Otherwise +1

dirk-thomas:
Since there was no such test before I would not block this PR and the others on that.

mikaelarguedas:
Makes sense. Ready for merge then

PR #129, opened 2016-05-06
[rosidl_generator_py] Fix array element checking

mikaelarguedas:
Check if the elements of arrays are of the right instance regardless of the array being bounded or not

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.template
+++ b/rosidl_generator_py/resource/_msg.py.template
@@ -103,9 +103,7 @@ class @(spec.base_type.type)(metaclass=Metaclass):
 @[      end if]@
 @[      if field.type.type == 'byte']@
         self.@(field.name) = kwargs.get('@(field.name)', b'0')
-@[      elif field.type.type in [
-            'char',
-            'string']]@
+@[      elif field.type.type == 'char']@
         self.@(field.name) = kwargs.get('@(field.name)', '\0')
 @[      elif field.type.array_size]
         self.@(field.name) = kwargs.get(
@@ -156,10 +154,8 @@ class @(spec.base_type.type)(metaclass=Metaclass):
 @[      else]@
             and len(value) == @(field.type.array_size) \
 @[      end if]@
-            and all([isinstance(v, @(get_python_type(field.type))) for v in value]))
-@[    else]@
-            )
 @[    end if]@
+            and all([isinstance(v, @(get_python_type(field.type))) for v in value]))
 @[  elif field.type.string_upper_bound]@
             ((isinstance(value, str) or isinstance(value, UserString)) \
             and len(value) <= @(field.type.string_upper_bound))
@@ -178,12 +174,9 @@ class @(spec.base_type.type)(metaclass=Metaclass):
         'int16', 'uint16',
         'int32', 'uint32',
         'int64', 'uint64',
+        'string',
     ]]@
             isinstance(value, @(get_python_type(field.type)))
-@[  elif field.type.type == 'string']@
-            isinstance(value, @(get_python_type(field.type)))
-        if not value.endswith('\0'):
-            value += '\0'
 @[  else]@
             False
 @[  end if]@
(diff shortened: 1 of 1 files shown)


wjwwood:
026ebde813e3747be53ad6f8c767514aba1ffaf1 lgtm

rosidl_generator_py/resource/_msg.py.template
```diff
@@ -103,7 +103,9 @@ class @(spec.base_type.type)(metaclass=Metaclass):
 @[      end if]@
 @[      if field.type.type == 'byte']@
         self.@(field.name) = kwargs.get('@(field.name)', b'0')
-@[      elif field.type.type == 'char']@
+@[      elif field.type.type in [
+            'char',
+            'string']]@
```
dirk-thomas: Why would a Python string field be initialized with a `\0`? Shouldn't it be an empty string?

rosidl_generator_py/resource/_msg.py.template
```diff
@@ -176,9 +178,12 @@ class @(spec.base_type.type)(metaclass=Metaclass):
         'int16', 'uint16',
         'int32', 'uint32',
         'int64', 'uint64',
-        'string',
     ]]@
             isinstance(value, @(get_python_type(field.type)))
+@[  elif field.type.type == 'string']@
+            isinstance(value, @(get_python_type(field.type)))
+        if not value.endswith('\0'):
+            value += '\0'
```
dirk-thomas: In Python a string shouldn't have a trailing `\0`.

rosidl_generator_py/resource/_msg.py.template
```diff
@@ -93,30 +93,32 @@ class @(spec.base_type.type)(metaclass=Metaclass):
...
         from @(field.type.pkg_name).msg import @(field.type.type)
-@[end if]@
-@[if field.type.type == 'byte']@
+@[      end if]@
+@[      if field.type.type == 'byte']@
         self.@(field.name) = kwargs.get('@(field.name)', b'0')
-@[elif field.type.type == 'char']@
+@[      elif field.type.type in [
+            'char',
+            'string']]@
```
dirk-thomas: Why would a Python string field be initialized with a \0? Shouldn't it be an empty string?

rosidl_generator_py/resource/_msg.py.template
```diff
@@ -127,60 +129,63 @@ class @(spec.base_type.type)(metaclass=Metaclass):
...
         'int32', 'uint32',
         'int64', 'uint64',
-        'string',
     ]]@
             isinstance(value, @(get_python_type(field.type)))
-@[else]@
+@[  elif field.type.type == 'string']@
+            isinstance(value, @(get_python_type(field.type)))
+        if not value.endswith('\0'):
+            value += '\0'
```
dirk-thomas: In Python a string shouldn't have a trailing `\0`.

rosidl_generator_py/resource/_msg.py.template
mikaelarguedas: Sorry this is still work in progress, I'm trying to figure out why the strings passed to C don't have the null character at the end (on the C side). This is highly experimental and not ready for review.
I'm curious though, why should a char be initialized to '\0' but not a string ?

dirk-thomas: A char is also one char. And it needs to be initialized with something. The null character seem appropriate given that we initialize all numeric values with zero.

For string on the other hand it doesn't make sense to initialize it to have any content by default, hence it should be initialized as being empty.

mikaelarguedas: Yeah sorry it was a stupid question that's why I edited my comment right after posting it.

mikaelarguedas:
Ok modifying the python template was indeed unnecessary. It seems to be working well now.
http://ci.ros2.org/job/ci_linux/1266/

dirk-thomas:
Since this PR has been reduced to just moving the check I would propose to close this since it is already included in #130.

mikaelarguedas:
Closing this in favor of #130

PR #130, opened 2016-05-07
Lint generated code

dirk-thomas:
This enables linting the generated code and fixes almost all failing tests.

@mikaelarguedas Can you please look into the memory leak reported by cppcheck: http://ci.ros2.org/job/ci_linux/1267/

I will look into the few remaining style related test failures.

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.template
+++ b/rosidl_generator_py/resource/_msg.py.template
@@ -18,7 +18,8 @@ except ImportError:
 
 if __type_support_importable:
     rclpy_implementation = rclpy._rclpy.rclpy_get_rmw_implementation_identifier()
-    module = import_type_support('@(package_name)', '@(subfolder)', '@(module_name)', rclpy_implementation)
+    module = import_type_support(
+        '@(package_name)', '@(subfolder)', '@(module_name)', rclpy_implementation)
     _convert_from_py = module.convert_from_py_@(module_name)
     _convert_to_py = module.convert_to_py_@(module_name)
     _type_support = module.type_support_@(module_name)
@@ -58,7 +59,7 @@ class Metaclass(type):
     @@property
     def @(constant.name)(self):
         """Message constant '@(constant.name)'."""
-        return Constants.__constants['@(constant.name)']
+        return Metaclass.__constants['@(constant.name)']
 @[end for]@
 @[for field in spec.fields]@
 @[  if field.default_value]@
@@ -72,24 +73,27 @@ class Metaclass(type):
 
 
 class @(spec.base_type.type)(metaclass=Metaclass):
+@[if not spec.constants]@
+    """Message class '@(spec.base_type.type)'."""
+@[else]@
     """
     Message class '@(spec.base_type.type)'.
-@[if spec.constants]@
 
     Constants:
 @[  for constant in spec.constants]@
       @(constant.name)
 @[  end for]@
-@[end if]@
     """
+@[end if]@
 
     __slots__ = [
 @[for field in spec.fields]@
         '_@(field.name)',
 @[end for]@
     ]
-
+@
 @[if len(spec.fields) > 0]@
+
     def __init__(self, **kwargs):
         assert all(['_' + key in self.__slots__ for key in kwargs.keys()]), \
             "Invalid arguments passed to constructor: %r" % kwargs.keys()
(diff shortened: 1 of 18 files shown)


rosidl_generator_py/resource/_msg.py.template
```diff
@@ -72,24 +73,27 @@ class Metaclass(type):
...
     """
+@[end if]@
 
     __slots__ = [
 @[for field in spec.fields]@
         '_@(field.name)',
 @[end for]@
     ]
-
+@
```
mikaelarguedas: Is this needed ?

rosidl_generator_py/resource/_msg.py.template
```diff
@@ -145,30 +149,31 @@ class @(spec.base_type.type)(metaclass=Metaclass):
...
-            and len(value) <= @(field.type.array_size) \
+             len(value) <= @(field.type.array_size))
 @[      else]@
-            and len(value) == @(field.type.array_size) \
+             len(value) == @(field.type.array_size))
 @[      end if]@
-            and all([isinstance(v, @(get_python_type(field.type))) for v in value]))
 @[    else]@
-            )
+             all([isinstance(v, @(get_python_type(field.type))) for v in value]))
```
mikaelarguedas: Would it make more sense to move this line out of the else in this PR or should I keep the change in [here](https://github.com/ros2/rosidl/pull/129) ?

mikaelarguedas:
lgtm. 
I'll look into the memory leak shortly

rosidl_generator_py/resource/_msg.py.template
dirk-thomas: Yes, otherwise the generated code contains an extra empty line.

rosidl_generator_py/resource/_msg.py.template
dirk-thomas: Since most of the changes in #129 have been reverted and only contains this single change I would propose to close #129 since it is already included here.

mikaelarguedas: Makes sense, I'll close #129 . Shouldn't this check be moved outside of the `@[    if field.type.array_size]@` statement given that we would like to test the elements for any kind of array ?

dirk-thomas: Good point. Done in 808589c7cea2c8fb4df4f95ac1244f1b6e87dd95

rosidl_generator_py/resource/_msg.py.template
```diff
@@ -98,7 +102,7 @@ class @(spec.base_type.type)(metaclass=Metaclass):
         self.@(field.name) = kwargs.get(
             '@(field.name)', @(spec.base_type.type).@(field.name.upper())__DEFAULT)
 @[    else]@
-@[      if not field.type.is_primitive_type()]@
+@[      if not field.type.is_primitive_type() and (not field.type.is_array or field.type.array_size)]@
```
mikaelarguedas: Could you detail this condition? I you receive a bounded array you import the message module but if it's an unbounded array you don't want to import it ?

dirk-thomas: The condition is the same as for the following code which only uses the type in line 115 (`array_size`) and 120 (`not is_array`).

mikaelarguedas: You're right, the import I was looking for (for all non primitive types) is in Line 131/135.

dirk-thomas:
And finally green - ready for review:

* http://ci.ros2.org/job/ci_linux/1287/
* http://ci.ros2.org/job/ci_osx/1024/
* http://ci.ros2.org/job/ci_windows/1320/

Requires ament/ament_lint#52.

rosidl_generator_py/resource/_msg.py.template
```diff
@@ -72,24 +73,27 @@ class Metaclass(type):
...
 
     __slots__ = [
 @[for field in spec.fields]@
         '_@(field.name)',
 @[end for]@
     ]
-
+@
 @[if len(spec.fields) > 0]@
+
```
mikaelarguedas: removing blank line at L91 to add it L96?

mikaelarguedas:
Except blank line comment, looks good to me. +1

rosidl_generator_py/resource/_msg.py.template
dirk-thomas: Correct, the template needs to generate the right amount of empty line before all methods for every possible input. Therefore it now doesn't generically add a newline in 91 but only adds it in 96 before `def__init__`. Later on each field related method also has a leading empty line.

PR #131, opened 2016-05-07
[rosidl_generator_py] Fix memory leak in generated c code

mikaelarguedas:
this PR adresses memory leak referenced [here](https://github.com/ros2/rosidl/pull/130#issue-153561062)


dirk-thomas:
+1 for freeing the allocated memory.

But why is the temporary variable `tmparray*` necessary in the first place? Since the data is copied into the message field anyway can't it be populated directly?

mikaelarguedas:
If I remember correctly I could populate directly the array if it was a bounded sized one but not an unbounded one. That's why I used this temporary variable. I'll merge this one as is and see if I can get rid of `tmparray*` by initializing the unbounded arrays beforehand and submit another PR if successful.

dirk-thomas:
See #132

PR #133, opened 2016-05-12
update style to pass new uncrustify version

dirk-thomas:
Connect to ament/uncrustify#13

While the parent ticket is blocked I think this could be merged.


dirk-thomas:
Please review this since the change can be merged independently of the uncrustify update.

wjwwood:
lgtm

PR #134, opened 2016-05-13
generate __repr__ for Python messages

dirk-thomas:


mikaelarguedas:
lgtm +1, thanks!

dirk-thomas:
Once per class might be better then once per field :wink: 147519b78b885a46e19eae7b6a32acbc2c06fb07

PR #135, opened 2016-05-13
move custom command to a separate CMake file

### LINKED ISSUE #124: Python generator called repeatedly in message interface packages

dirk-thomas:
After: http://ci.ros2.org/job/ci_windows/1333/consoleFull (search for "Generating Python" which only appears once for the `rosidl_generator_py` package).

Fixes #124.


tfoote:
It seems to work. Crazy workaround. 

+1

PR #137, opened 2016-05-13
move repr to avoid multiple declarations

mikaelarguedas:
This declare the repr function once per messages. I didn't catch it when reviewing #134 .
Should solve new pyflakes complaints cf. http://ci.ros2.org/job/ci_linux/1299/


dirk-thomas:
Sorry, I already committed the fix to master (147519b78b885a46e19eae7b6a32acbc2c06fb07). I saw this too late. Btw. this patch lacks an empty line before `__repr__`.

mikaelarguedas:
no worries, as long as it's fixed upstream

PR #138, opened 2016-05-18
Document string behavior and initialize to the empty string

jacquelinekay:
Attempting to document the behavior of rosidl_generator_c strings that the typesupport libraries expect, and initialize rosidl_generator_c__String to the empty string state as it is understood in the typesupport libraries.

@mikaelarguedas and @wjwwood in particular, does this make sense?

### DIFF
--- a/rosidl_generator_c/include/rosidl_generator_c/string_functions.h
+++ b/rosidl_generator_c/include/rosidl_generator_c/string_functions.h
@@ -25,11 +25,12 @@ extern "C"
 {
 #endif
 
-/* The contents of rosidl_generator_c__String are initialized a single null character '\0'.
- The string initially has size 0 and capacity 1.
- Size represents the size of the contents of the string, while capacity represents the overall
- storage of the string (counting the null terminator).
- All strings must be null-terminated.
+/// Initialize a rosidl_generator_c__String structure.
+/* The contents of rosidl_generator_c__String are initialized to a single null character ('\0').
+ * The string initially has size 0 and capacity 1.
+ * Size represents the size of the contents of the string, while capacity represents the overall
+ * storage of the string (counting the null terminator).
+ * All strings must be null-terminated.
  */
 ROSIDL_GENERATOR_C_PUBLIC
 void
--- a/rosidl_generator_c/include/rosidl_generator_c/string_functions.h
+++ b/rosidl_generator_c/include/rosidl_generator_c/string_functions.h
@@ -33,7 +33,7 @@ extern "C"
  * All strings must be null-terminated.
  */
 ROSIDL_GENERATOR_C_PUBLIC
-void
+bool
 rosidl_generator_c__String__init(rosidl_generator_c__String * str);
 
 ROSIDL_GENERATOR_C_PUBLIC
--- a/rosidl_generator_c/include/rosidl_generator_c/string_functions.h
+++ b/rosidl_generator_c/include/rosidl_generator_c/string_functions.h
@@ -25,6 +25,12 @@ extern "C"
 {
 #endif
 
+/* The contents of rosidl_generator_c__String are initialized a single null character '\0'.
+ The string initially has size 0 and capacity 1.
+ Size represents the size of the contents of the string, while capacity represents the overall
+ storage of the string (counting the null terminator).
+ All strings must be null-terminated.
+ */
 ROSIDL_GENERATOR_C_PUBLIC
 void
 rosidl_generator_c__String__init(rosidl_generator_c__String * str);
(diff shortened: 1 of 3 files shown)


rosidl_generator_c/src/string_functions.c
```diff
@@ -43,7 +44,7 @@ rosidl_generator_c__String__fini(rosidl_generator_c__String * str)
     str->size = 0;
     str->capacity = 0;
   } else {
-    /* ensure that data, szie and and capacity values are consistent */
+    /* ensure that data, size, and capacity values are consistent */
     assert(0 == str->size);
     assert(0 == str->capacity);
```
mikaelarguedas: should the asserts on the capacity use 1 instead of 0 for string checking?
e.g. line 40/41 `assert(str->capacity > 1);` and line 48/49 `assert(1 == str->capacity);`

mikaelarguedas:
The comment is very clear, thanks for documenting it

rosidl_generator_c/src/string_functions.c
jacquelinekay: If `str->data` evaluates to false, then it points to a null pointer and does not have a "null" character allocated to it. Therefore it is in the fini'd state where both size and capacity are 0.

jacquelinekay: Should I add a comment to this effect? Basically this assert is saying "if I got here there the string was already fini'd".

wjwwood: Should this only be checked in Debug mode?

rosidl_generator_c/include/rosidl_generator_c/string_functions.h
```diff
@@ -25,6 +25,12 @@ extern "C"
 {
 #endif
 
+/* The contents of rosidl_generator_c__String are initialized a single null character '\0'.
+ The string initially has size 0 and capacity 1.
+ Size represents the size of the contents of the string, while capacity represents the overall
+ storage of the string (counting the null terminator).
+ All strings must be null-terminated.
+ */
```
wjwwood: Might as well make it Doxygen style, no?

```c++
/// Initialize a rosidl_generator_c__String structure.
/* The contents of rosidl_generator_c__String are initialized to a single null character ('\0').
 * The string initially has size 0 and capacity 1.
 * Size represents the size of the contents of the string, while capacity represents the overall
 * storage of the string (counting the null terminator).
 * All strings must be null-terminated.
 */
```

wjwwood:
Makes sense to me too, thanks!

rosidl_generator_c/src/string_functions.c
jacquelinekay: What other style of error handling would you suggest? Change the return signature to an error code?

wjwwood: You can do `fprintf(stderr, ...)` and `exit(1)`.

rosidl_generator_c/src/string_functions.c
```diff
@@ -24,9 +24,10 @@ rosidl_generator_c__String__init(rosidl_generator_c__String * str)
   if (!str) {
     return;
   }
-  str->data = NULL;
+  str->data = malloc(1);
```
wjwwood: This can fail I think. (`if (!str->data) ... error`)

rosidl_generator_c/src/string_functions.c
wjwwood: If the rest of the code is using `assert`, then I guess you don't have to change it. The code in general doesn't handle errors very consistently.

mikaelarguedas: No need to document it, my bad I was thinking in cython.

wjwwood: It might should be refactored separately.

rosidl_generator_c/src/string_functions.c
jacquelinekay: 300bf2232bd7efad2cffbde7583cf5a0c0e08aa5

dirk-thomas: Shouldn't there be some kind of error signaling here?

jacquelinekay: OK, since everyone is complaining about error checking in this class, I will change all of the function signatures to return bool, and return false on an error state.

jacquelinekay:
Ok, I'm using exit(-1) instead of assert and returning true out of init now.

mikaelarguedas:
lgtm

jacquelinekay:
Another CI job:

http://ci.ros2.org/job/ci_linux/1328/
http://ci.ros2.org/job/ci_osx/1045/
http://ci.ros2.org/job/ci_windows/1349/

PR #141, opened 2016-05-20
Fix nested messages for introspection C

jacquelinekay:
Potential fix for the broken nested messages in introspection C typesupport.

@mikaelarguedas can you see if this branch fixes your issues?

### DIFF
--- a/rosidl_typesupport_introspection_c/resource/msg__introspection_type_support.h.em
+++ b/rosidl_typesupport_introspection_c/resource/msg__introspection_type_support.h.em
@@ -23,15 +23,8 @@ function_prefix = '%s__%s__rosidl_typesupport_introspection_c' % (spec.base_type
 #ifndef @(header_guard_variable)
 #define @(header_guard_variable)
 
-#if __cplusplus
-extern "C"
-{
-#endif
+static_assert(USING_ROSIDL_TYPESUPPORT_CONNEXT_C, "expected rosidl introspection C type support");
 
 #include <rosidl_generator_c/message_type_support.h>
-
-#if __cplusplus
-}
-#endif
 
 #endif  // @(header_guard_variable)
--- a/rosidl_typesupport_introspection_c/resource/msg__introspection_type_support.h.em
+++ b/rosidl_typesupport_introspection_c/resource/msg__introspection_type_support.h.em
@@ -28,11 +28,9 @@ function_prefix = '%s__%s__rosidl_typesupport_introspection_c' % (spec.base_type
 #include "@(spec.base_type.pkg_name)/msg/rosidl_typesupport_introspection_c__visibility_control.h"
 
 @[if subfolder == 'msg']@
-static rosidl_message_type_support_t @(function_prefix)__@(spec.base_type.type)_message_type_support_handle;
-
 ROSIDL_TYPESUPPORT_INTROSPECTION_C_EXPORT_@(spec.base_type.pkg_name)
 const rosidl_message_type_support_t *
-ROSIDL_GET_TYPE_SUPPORT_FUNCTION(@(spec.base_type.pkg_name), @(subfolder), @(spec.msg_name))();
+ROSIDL_GET_TYPE_SUPPORT(@(spec.base_type.pkg_name), @(subfolder), @(spec.msg_name));
 @[end if]@
 
 #endif  // @(header_guard_variable)
--- a/rosidl_typesupport_introspection_c/resource/msg__introspection_type_support.h.em
+++ b/rosidl_typesupport_introspection_c/resource/msg__introspection_type_support.h.em
@@ -29,6 +29,9 @@ function_prefix = '%s__%s__rosidl_typesupport_introspection_c' % (spec.base_type
 
 @[if subfolder == 'msg']@
 static rosidl_message_type_support_t @(function_prefix)__@(spec.base_type.type)_message_type_support_handle;
+
+const rosidl_message_type_support_t *
+ROSIDL_GET_TYPE_SUPPORT_FUNCTION(@(spec.base_type.pkg_name), @(subfolder), @(spec.msg_name))();
 @[end if]@
 
 #endif  // @(header_guard_variable)
--- a/rosidl_typesupport_introspection_c/resource/msg__introspection_type_support.h.em
+++ b/rosidl_typesupport_introspection_c/resource/msg__introspection_type_support.h.em
@@ -28,7 +28,7 @@ function_prefix = '%s__%s__rosidl_typesupport_introspection_c' % (spec.base_type
 #include "@(spec.base_type.pkg_name)/msg/rosidl_typesupport_introspection_c__visibility_control.h"
 
 @[if subfolder == 'msg']@
-ROSIDL_TYPESUPPORT_INTROSPECTION_C_EXPORT_@(spec.base_type.pkg_name)
+ROSIDL_TYPESUPPORT_INTROSPECTION_C_PUBLIC_@(spec.base_type.pkg_name)
 const rosidl_message_type_support_t *
 ROSIDL_GET_TYPE_SUPPORT(@(spec.base_type.pkg_name), @(subfolder), @(spec.msg_name));
 @[end if]@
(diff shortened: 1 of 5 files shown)


rosidl_typesupport_introspection_c/resource/msg__introspection_type_support.h.em
```diff
@@ -29,6 +29,9 @@ function_prefix = '%s__%s__rosidl_typesupport_introspection_c' % (spec.base_type
 
 @[if subfolder == 'msg']@
 static rosidl_message_type_support_t @(function_prefix)__@(spec.base_type.type)_message_type_support_handle;
+
+const rosidl_message_type_support_t *
+ROSIDL_GET_TYPE_SUPPORT_FUNCTION(@(spec.base_type.pkg_name), @(subfolder), @(spec.msg_name))();
```
wjwwood: Is this done for the side-effect?

wjwwood: Sorry, I thought that was calling it, but that's just a function prototype...

jacquelinekay: yep

wjwwood:
makes sense to me, +1

mikaelarguedas:
On a cosmetic side, would it be nicer to use `ROSIDL_GET_TYPE_SUPPORT(@(spec.base_type.pkg_name), @(subfolder), @(spec.msg_name))` instead of `ROSIDL_GET_TYPE_SUPPORT_FUNCTION(@(spec.base_type.pkg_name), @(subfolder), @(spec.msg_name))()` ?

mikaelarguedas:
Seems the generate a bunch of unused variable warnings on my side.
Does that fix the nested message issue on your side ?
Starting testing now, for now it's complaining about type support handle argument being null at publisher creation. I'll look into it soon and let you know

jacquelinekay:
I just did a little more testing my side and observed some issues with strings.

mikaelarguedas:
Works on my side including nested messages with "empty strings"('\0').
Some issue with unbounded arrays but I don't think that it's typesupport related.
Thanks!

jacquelinekay:
Relaunched to see if warnings are fixed: http://ci.ros2.org/job/ci_linux/1320/

jacquelinekay:
Pending CI:
http://ci.ros2.org/job/ci_linux/1329/
http://ci.ros2.org/job/ci_osx/1046
http://ci.ros2.org/job/ci_windows/1350/

jacquelinekay:
There are some Windows warnings...

jacquelinekay:
ok, I still don't know how to fix the inconsistent dll linkage warning. I think it has something to do with using the ROSIDL_GET_TYPE_SUPPORT_FUNCTION for the message fields of a nested message, but I still have no idea how to fix it. I've tried several things (like using `ROSIDL_GENERATOR_C_PUBLIC/ROSIDL_GENERATOR_C_EXPORT` instead of `ROSIDL_TYPESUPPORT_INTROSPECTION_C`). @wjwwood let me know if you have any tips.

wjwwood:
I think 518087fcacd27eea808f893fbda3b0d90016ccf2 should be right. If it's not, then the warning might be coming from somewhere else. I probably won't have time to dig into though... I'm pretty behind on my own issues.

jacquelinekay:
ok, I finally figured it out.

`ROSIDL_GET_TYPE_SUPPORT_FUNCTION` is forward declared in `rosidl_generator_c` with the macro `ROSIDL_GENERATOR_C_PUBLIC_@(pkg_name)`. So the implementation in this package needs to use `ROSIDL_GENERATOR_C_EXPORT_@(pkg_name)`. And in order to avoid warnings about locally defined symbols we need to use `ROSIDL_GENERATOR_C_BUILDING_DLL_@(pkg_name)` in generate_interfaces.  	48d93f7

I'll clean this up and run another Windows job.

jacquelinekay:
http://ci.ros2.org/job/ci_windows/1357/badge/icon

wjwwood:
> http://ci.ros2.org/job/ci_windows/1357/badge/icon

?

jacquelinekay:
I'm trying to use the embedded build status thing but because this job hasn't started yet the link is broken... I guess it's not especially useful right now because they Windows jobs will come up as unstable even if they're on par with the nightly build

mikaelarguedas:
from what I understood it seems to be `[BuildStatus](joburl)`

jacquelinekay:
Personally I'm a fan of doing things without reading the instructions

[Build Status](http://ci.ros2.org/job/ci_windows/1357)

The link will still be broken/won't render until 1356 finishes

dirk-thomas:
The package `rosidl_typesupport_introspection_c` should keep setting `ROSIDL_TYPESUPPORT_INTROSPECTION_C_BUILDING_DLL_${PROJECT_NAME}` and not `ROSIDL_GENERATOR_C_BUILDING_DLL_${PROJECT_NAME} as in 48d93f7b48bcfee51525db3a00cab33de6ebe4b9. The names of these definitions are based on the package names. Setting them across packages will lead to confusion and wrong import / export declarations.

Have you tried to use `ROSIDL_TYPESUPPORT_INTROSPECTION_C_EXPORT_@(spec.base_type.pkg_name)` to define `ROSIDL_GET_TYPE_SUPPORT_FUNCTION` instead?

jacquelinekay:
I did try that, in 518087f, however that commit still had the redundant forward declare in the header provided by rosidl_typesupport_introspection_c, which could have been the cause of the inconsistent DLL linkage warning...

jacquelinekay:
ok, the problem with using `ROSIDL_TYPESUPPORT_INTROSPECTION_C_EXPORT_@(spec.base_type.pkg_name)` to declare `ROSIDL_GET_TYPE_SUPPORT_FUNCTION` (even without a redudant declaration in the header) is that it causes this warning:

```
parameter_event_descriptors__type_support.obj : warning LNK4217: locally defined symbol rosidl_typesupport_introspection_c_get_message__rcl_interfaces__msg
__ParameterDescriptor imported in function rosidl_typesupport_introspection_c_get_message__rcl_interfaces__msg__ParameterEventDescriptors [C:\ros2_ws\build\r
cl_interfaces\rcl_interfaces__rosidl_typesupport_introspection_c.vcxproj]
```
The warning occurs for every nested message which has a subfield that is another message defined in the package.

jacquelinekay:
[Build Status](http://ci.ros2.org/job/ci_windows/1358/)

jacquelinekay:
since there are +1s on this pull request and it's needed to fix nested dynamic messages in Python, I'm going to merge it.

dirk-thomas:
The patch still sets definitions across packages as I mentioned in my comment. I don't think this should stay as-is.

rosidl_typesupport_introspection_c/resource/msg__type_support.c.em
```diff
@@ -198,9 +199,15 @@ static rosidl_message_type_support_t @(function_prefix)__@(spec.base_type.type)_
   &@(function_prefix)__@(spec.base_type.type)_message_members
 };
 
-ROSIDL_TYPESUPPORT_INTROSPECTION_C_EXPORT_@(spec.base_type.pkg_name)
+ROSIDL_GENERATOR_C_EXPORT_@(spec.base_type.pkg_name)
```
dirk-thomas: The c code generated in package `rosidl_typesupport_introspection_c` can not use this export macro. It must use its own.

jacquelinekay: As I've said before, I don't think it's possible to do this without causing a warning on Windows because this function is forward-declared in rosidl_generator_c. I can try again but I'm going to prioritize fixing test failures.

dirk-thomas: Please ticket this since it must be cleaned up.

wjwwood: So long as it's `*_EXPORT*`, you can use any of them (including the one @dirk-thomas is suggesting). `*_EXPORT*` always evaluates to `__declspec(dllexport)` on Windows, regardless of what `-D` options are passed to the compiler.

It is still true, however, that the declaration of this function uses the `ROSIDL_GENERATOR_C_PUBLIC_*` macro and at least in my opinion, it would make sense if it matched, since this is just the definition of that function declared in the `rosidl_generator_c` code and not a new function that is specific to `rosidl_typesupport_introspection_c`.

PR #142, opened 2016-05-25
Only add linter tests if option in rosidl_generate_interfaces is set

dhood:
at the moment it seems that pretty much all packages with interfaces except `rcl_interfaces`, `test_communication` and `test_rclcpp` are outputting cmake warnings because of (I think) https://github.com/ros2/rosidl/commit/454d8d5c70ed54670efe8922d4bd4243f1b0e47a. 
this will turn those warnings from generic 'Could not find a package configuration file' warnings to ones that warn that the generated code isn't being linted

I'll open some PRs to add the linter test dependencies to the relevant packages to make the warnings stop (unless there's a reason not to test the other packages' generated files)


rosidl_generator_c/cmake/rosidl_generator_c_generate_interfaces.cmake
```diff
@@ -174,29 +174,46 @@ if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
...
-      "${_output_path}")
+if(BUILD_TESTING)
+  if(NOT "${_generated_msg_headers}${_generated_msg_sources}${_generated_srv_headers}${_generated_srv_sources} " STREQUAL " ")
+    find_package(ament_cmake_cppcheck QUIET)
+    if(ament_cmake_cppcheck_FOUND)
+      ament_cppcheck(
+        TESTNAME "cppcheck_rosidl_generated_c"
+        "${_output_path}")
+    else()
+      message(WARNING
```
dirk-thomas: I think a warning is not appropriate here. It should be fine for a package to not use those test dependencies and still build without warning.

dhood: ok. perhaps just change it to a message?

dhood: it's a lot of output though. I think it would be better if it were just one message which mentioned all of the packages that couldn't be found?

wjwwood: Something like this would be nice:

```
-- 'X' not found: skipping linting of generated message code
-- 'Y' not found: skipping linting of generated message code
-- One or more linters were not found. Please add them as a <test_depend> to enable linting.
```

The other option would be to just add `QUIET` to the `find_package`, but provide an option to `generate_messages` which switches that to `REQUIRED`. That way new message packages will not complain if the linters are not there, but we can ensure for a few of our core packages that the linters are found and run on the generated code.

dhood: I think that adding an option to `rosidl_generate_interfaces` could be a good idea because otherwise there's not much that can be done if someone wants to use the linters _only_ for their non-generated code. if the linters are there, they're used to test generated code, but that might not be a good fit for everyone.

we could have an option that allows for a) no linting of generated code b) linting with whatever linters are available (`QUIET`) or c) linting with `REQUIRED`

wjwwood: > we could have an option that allows for a) no linting of generated code b) linting with whatever linters are available (`QUIET`) or c) linting with `REQUIRED`

That sounds good to me. What would be the default? Option b or a?

dhood: option a, since it's almost always just a waste of time linting them since every package is using the same template. option b can be removed. option c is useful for core packages to ensure the tests are being run.

wjwwood: I agree, option b doesn't make sense. Either it's a core package and we want to ensure the linting is being done, or it's a user's message package and they probably don't care about linting their generated message code. That's because there is good coverage on the generated code in the core message packages and it makes running all your tests slightly longer.

dhood:
I've added the `LINT_INTERFACES` option to `rosidl_generate_interfaces` which will find the linters with `REQUIRED`, and otherwise not try to find them.

Coupled with a change like [this](https://github.com/ros2/common_interfaces/compare/master...dhood:cmake-lint-warnings) we will have the `common_interfaces` being linted but otherwise no linting of generated interface code and no warnings. linux CI for that branch and this one: http://ci.ros2.org/view/All/job/ci_linux/1350/testReport/

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
```diff
@@ -33,11 +33,14 @@ include(CMakeParseArguments)
 # :type DEPENDENCIES: list of strings
 # :param SKIP_INSTALL: if set skip installing the interface files
 # :type SKIP_INSTALL: option
+# :param LINT_INTERFACES: if set lint the interface files using
+#   the ``ament_lint`` package
```
dhood: this does not do a very good job at specifying what linters are run, but it seems like something that will get out-of-date easily so I thought leaving it generic might be better?

dirk-thomas: Since the macro itself does not perform any linting directly but only registers tests I think this should be reflected in the name of the option: e.g. `ADD_LINTER_TESTS`

dhood: thanks! I'll change it to that

dhood:
@wjwwood could you review please

tfoote:
+1

dhood:
ok will re-run CI and update https://github.com/ros2/rmw_connext/pull/163 and co to use this structure

dhood:
CI's fine. could you just double check my reasoning in https://github.com/ros2/rosidl/pull/142/commits/c987e35d20f52eb890274a6dae318ca872b592f0 then I'll merge all connected PRs

wjwwood:
+1 go for it.

PR #145, opened 2016-06-03
fix bounded arrays

dirk-thomas:
At least compiling the generated code for unbounded array passes with this.

Connect to ros2/system_tests#123.

### DIFF
--- a/rosidl_generator_py/resource/_msg_support.c.em
+++ b/rosidl_generator_py/resource/_msg_support.c.em
@@ -84,7 +84,11 @@ nested_type = '%s__%s__%s' % (field.type.pkg_name, 'msg', field.type.type)
   @(nested_type) * dest@(field.name) = ros_message->@(field.name).data;
 @[      else]@
   size_t size@(field.name) = @(field.type.array_size);
+@[        if field.type.is_upper_bound]@
+  @(nested_type) * dest@(field.name) = ros_message->@(field.name).data;
+@[        else]@
   @(nested_type) * dest@(field.name) = ros_message->@(field.name);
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Please review.

rosidl_generator_py/resource/_msg_support.c.em
```diff
@@ -84,7 +84,11 @@ nested_type = '%s__%s__%s' % (field.type.pkg_name, 'msg', field.type.type)
   @(nested_type) * dest@(field.name) = ros_message->@(field.name).data;
 @[      else]@
   size_t size@(field.name) = @(field.type.array_size);
+@[        if field.type.is_upper_bound]@
+  @(nested_type) * dest@(field.name) = ros_message->@(field.name).data;
+@[        else]@
   @(nested_type) * dest@(field.name) = ros_message->@(field.name);
+@[        end if]@
```
mikaelarguedas: I think that you want to call the `__Array__init()` methods for bounded arrays.
I would suggest checking `is_upper_bound` L79 and if the array is bounded, use `min(PySequence_Size(),upper_bound)` to set the array size.

Same below

dirk-thomas: This PR only fixes the compilation. Getting the code to actually work will take another iteration.

mikaelarguedas: Alright, in this case let's merge the first iteration and get things fixed before actually merging ros2/system_tests#124

mikaelarguedas:
lgtm

PR #146, opened 2016-06-04
do not overwrite init files

mikaelarguedas:
This reverts a hack/fix we introduced [here](https://github.com/ros2/rosidl/pull/125) to avoid the python generator to be invoked multiple times. This would prevent cmake to overwrite generated init files if reran.
It should allow windows to find the python modules.
Not really ready for review yet. But seems to work on linux without introducing multiple generator invocations.

### DIFF
--- a/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
@@ -68,15 +68,15 @@ foreach(_idl_file ${rosidl_generate_interfaces_IDL_FILES})
   endif()
 endforeach()
 
+file(MAKE_DIRECTORY "${_output_path}")
+file(WRITE "${_output_path}/__init__.py" "")
+
 if(NOT "${_generated_msg_py_files} " STREQUAL " ")
   list(GET _generated_msg_py_files 0 _msg_file)
   get_filename_component(_parent_folder "${_msg_file}" DIRECTORY)
   list(APPEND _generated_msg_py_files
     "${_parent_folder}/__init__.py"
   )
-  # TODO(dirk-thomas) since this are being generated at configure time
-  # it should not be a dependency of the custom command
-  file(WRITE "${_parent_folder}/__init__.py" "")
 endif()
 
 if(NOT "${_generated_srv_files} " STREQUAL " ")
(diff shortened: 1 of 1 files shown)


rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -68,15 +68,15 @@ foreach(_idl_file ${rosidl_generate_interfaces_IDL_FILES})
...
+
 if(NOT "${_generated_msg_py_files} " STREQUAL " ")
   list(GET _generated_msg_py_files 0 _msg_file)
   get_filename_component(_parent_folder "${_msg_file}" DIRECTORY)
   list(APPEND _generated_msg_py_files
     "${_parent_folder}/__init__.py"
   )
-  # TODO(dirk-thomas) since this are being generated at configure time
-  # it should not be a dependency of the custom command
-  file(WRITE "${_parent_folder}/__init__.py" "")
```
dirk-thomas: Where is this file supposed to be generated now?

mikaelarguedas: it's generated [here](https://github.com/ros2/rosidl/blob/master/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py#L98). IMO if we got the problem previously is because we were not constraining the scope to the msg folder so it was expecting an __init__.py file in the srv folder and could never find it.

dirk-thomas: If the Python script is in charge of generating them they need to be listed under `add_custom_command(OUTPUT ..)`.

mikaelarguedas: They are added to the list `_generated_msg_py_files` above on line 75/78. Which is used for the custom command.
Is there anywhere else I should add it ?

dirk-thomas: Oh, I missed that. Sorry for the noise.

mikaelarguedas:
Seems to work as expected. No appearance of multiple calls of the generator on windows
If you look for "Generating Python code" it appears only once per package
[Build Status](http://ci.ros2.org/job/ci_windows/1419)
Note: the job was not ran on this branch but on another one that cherry-picked this commit.
We can another round of CI jobs just for this change once the test-fixing party is over if we need more convincing

tfoote:
+1, do you want to merge this or close this in favor of the other branch with it cherry-picked?

mikaelarguedas:
I'll merge it as is and rebase the other branches. Thx

PR #148, opened 2016-06-07
Fix python cmake module

dirk-thomas:
Fix CMake module on Windows.

Before: http://ci.ros2.org/view/nightly/job/nightly_win_deb/142/consoleFull#console-section-73


mikaelarguedas:
works for me. +1

PR #150, opened 2016-06-08
char and byte init and conversion

mikaelarguedas:
Connects to ros2/rclpy#22

Note: this will not fix the crash on char value outside of the 0-127 range. But will work for all the tests added in ros2/system_tests#121

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -49,8 +49,7 @@ class Metaclass(type):
 @[end for]@
 @[for field in spec.fields]@
 @[  if field.default_value]@
-            '@(field.name.upper())__DEFAULT': @value_to_py(field.type, field.default_value,
-                                                           array_as_tuple=True),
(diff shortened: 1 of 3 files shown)


mikaelarguedas:
please review

tfoote:
+1

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -105,17 +104,30 @@ class @(spec.base_type.type)(metaclass=Metaclass):
 @[      if not field.type.is_primitive_type() and (not field.type.is_array or field.type.array_size)]@
         from @(field.type.pkg_name).msg import @(field.type.type)
 @[      end if]@
-@[      if field.type.array_size]
+@[      if field.type.array_size]@
+@[        if field.type.type == 'byte']@
         self.@(field.name) = kwargs.get(
             '@(field.name)',
-            tuple([@(get_python_type(field.type))() for x in range(@(field.type.array_size))])
+            list([bytes([0]) for x in range(@(field.type.array_size))])
```
dirk-thomas: If `x` is not used you should use `_`.

Same below.

PR #151, opened 2016-06-10
fix using only SOABI on Python 3.5, remove obsolete MULTIARCH

dirk-thomas:
In Python 3.5 the SOABI already contains the MULTIARCH.

See http://ci.ros2.org/job/ci_linux/1432/


mikaelarguedas:
lgtm

jacquelinekay:
+1, I tested it locally and it fixes my issues with Python 3.5.

PR #153, opened 2016-06-13
Windows python debug

mikaelarguedas:
After discussion with @dirk-thomas, we will modify the python interpreter only for packages generating/using python extensions and not for the entire build process. If people want to generate executables in Debug mode they'll have to invoke ament_tools with `python_d`.

This adds the suffix and cmake logic to use the python debug interpreter when generating python extensions.

CI with change: http://ci.ros2.org/job/ci_windows/1455/#showFailuresLink
Nightly: http://ci.ros2.org/view/nightly/job/nightly_win_deb/146/#showFailuresLink


python_cmake_module/cmake/Modules/FindPythonExtra.cmake
```diff
@@ -211,7 +222,11 @@ if(PYTHONINTERP_FOUND)
   endif()
 
   if(WIN32)
-    set(PythonExtra_EXTENSION_EXTENSION ".pyd")
+    if("${CMAKE_BUILD_TYPE} " STREQUAL "Debug ")
+      set(PythonExtra_EXTENSION_EXTENSION "_d.pyd")
```
dirk-thomas: Should this also happen when building debug but not having a debug interpreter?

mikaelarguedas: Not necessarily, but if one doesn't have a debug interpreter, Visual Studio won't be able to link against it and the compilation will fail. So the suffix of the extensions would not matter in that case.

dirk-thomas: If this can be detected at CMake time wouldn't it be better to print a clear error message then let it fail later with a not that clear error?

mikaelarguedas:
Round2: the goal here is to use the python_d.exe only when needed instead of for the entire build process.
 * PythonExtra module creates a PYTHON_DBG_EXECUTABLE with the path of python_d.exe and throw an error is it can't be found
 * ament_cmake_nosetests now take the python executable as a parameter to allow using python_d on a per test basis
 * Modules who need the `PYTHON_EXECUTABLE` variable overwrite it in their CMakeLists and restore it to the previous value once done (I'm not sure [the way it's done](https://github.com/ros2/rclpy/pull/34/commits/8bf6aff9babdc768814f9be57787e098a7dc7fc3) is the best way to do it in CMake )

CI passed with the expected result: http://ci.ros2.org/job/ci_windows/1481/

Note: people who want to build ament_python packages to generate executables (setuptools) will need to call `ament.py` with python_d.

python_cmake_module/cmake/Modules/FindPythonExtra.cmake
mikaelarguedas: PythonExtra now performs the existence test of the debug interpreter and throw an error if unable to find it. https://github.com/ros2/rosidl/pull/153/commits/52eb0f4e9d76ddd7d3dfd8f5f39f3cdb764c55d2

dirk-thomas:
The new variable should following the naming scheme of other CMake variables (which append the build type for customization). So I would suggest `PYTHON_EXECUTABLE_DEBUG`.

mikaelarguedas:
Makes sense. Renamed it in all packages using it.

mikaelarguedas:
@ros2/ros2_team please review this and the associated PRs

python_cmake_module/cmake/Modules/FindPythonExtra.cmake
```diff
@@ -128,6 +128,18 @@ if(PYTHONINTERP_FOUND)
     message(STATUS "Using PythonExtra_LIBRARIES: ${PythonExtra_LIBRARIES}")
   else()
     find_package(PythonLibs 3.5 REQUIRED)
+    if(WIN32 AND "${CMAKE_BUILD_TYPE} " STREQUAL "Debug ")
```
dirk-thomas: Please remove the quotes and explicit variable expansion as well as the trailing space.

python_cmake_module/cmake/Modules/FindPythonExtra.cmake
```diff
@@ -128,6 +128,18 @@ if(PYTHONINTERP_FOUND)
...
+      get_filename_component(_python_executable_dir "${PYTHON_EXECUTABLE}" DIRECTORY)
+      get_filename_component(_python_executable_name "${PYTHON_EXECUTABLE}" NAME_WE)
+      get_filename_component(_python_executable_ext "${PYTHON_EXECUTABLE}" EXT)
+      set(_python_executable_debug "${_python_executable_dir}/${_python_executable_name}_d${_python_executable_ext}")
+      if(EXISTS "${_python_executable_debug}")
+        set(PYTHON_EXECUTABLE_DEBUG "${_python_executable_debug}")
+      else()
+        message(FATAL_ERROR ${_python_executable_debug} " doesn't exist")
```
dirk-thomas: Please quote the entire message instead of only parts of it.

dirk-thomas:
Beside the comments +1.

python_cmake_module/cmake/Modules/FindPythonExtra.cmake
mikaelarguedas: done https://github.com/ros2/rosidl/pull/153/commits/11fa14c71351b8c523efc64765903e33ff2f0cb3

python_cmake_module/cmake/Modules/FindPythonExtra.cmake
mikaelarguedas: done https://github.com/ros2/rosidl/pull/153/commits/451ca5aef540c40a1af7151dd0378922082ced82

mikaelarguedas:
needs rebase and a round of CI before merging

mikaelarguedas:
new CI job to validate after rebase: [Build Status](http://ci.ros2.org/job/ci_windows/1508/)

mikaelarguedas:
@dirk-thomas does the +1 covers the associated PRs?

mikaelarguedas:
I'll fix the variable expansion/whitespace in the associated PRs too

dirk-thomas:
I just review the referenced PRs individually.

mikaelarguedas:
thanks!

mikaelarguedas:
Windows Debug job: [Build Status](http://ci.ros2.org/job/ci_windows/1510/)

mikaelarguedas:
CI passed, so I'll merge this now

PR #156, opened 2016-06-22
Fix bug with fastrtps/connext building redundant library

jacquelinekay:
Should fix a bug introduced by FastRTPS C typesupport pointed out by @dirk-thomas in https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/27#issuecomment-227798359.

@mikaelarguedas you also might want to take a look.

### DIFF
--- a/rosidl_generator_py/cmake/rosidl_generator_py_get_typesupports.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_get_typesupports.cmake
@@ -15,7 +15,7 @@
 macro(accumulate_typesupports)
   set(_typesupport_impl "")
   get_rmw_typesupport(_typesupport_impl ${rmw_implementation} LANGUAGE "C")
-  list(APPEND _typesupport_impls ${_typesupport_impl})
+  list_append_unique(_typesupport_impls ${_typesupport_impl})
 endmacro()
 
 macro(rosidl_generator_py_get_typesupports TYPESUPPORT_IMPLS)
--- a/rosidl_generator_py/CMakeLists.txt
+++ b/rosidl_generator_py/CMakeLists.txt
@@ -4,6 +4,7 @@ project(rosidl_generator_py)
 
 find_package(ament_cmake REQUIRED)
 
+ament_export_dependencies(ament_cmake)
 ament_export_dependencies(rosidl_cmake)
 
 ament_python_install_package(${PROJECT_NAME})


rosidl_generator_py/cmake/rosidl_generator_py_get_typesupports.cmake
```diff
@@ -15,7 +15,7 @@
 macro(accumulate_typesupports)
   set(_typesupport_impl "")
   get_rmw_typesupport(_typesupport_impl ${rmw_implementation} LANGUAGE "C")
-  list(APPEND _typesupport_impls ${_typesupport_impl})
+  list_append_unique(_typesupport_impls ${_typesupport_impl})
```
dirk-thomas: This function will require `ament_cmake_core` which is currently not exported by this package and therefore it isn't ensured that it is available to the package calling this macro.

mikaelarguedas:
Looks good to me, do you have a CI job validating the fix ?

rosidl_generator_py/cmake/rosidl_generator_py_get_typesupports.cmake
jacquelinekay: Already an existing problem in this file: https://github.com/ros2/rosidl/pull/156/files#diff-91406625e1eac1b288db266065a388a5L26

I'll add the export to this package in this PR.

jacquelinekay: do

mikaelarguedas: good point, the same macro is used in [generate_interfaces](https://github.com/ros2/rosidl/blob/unique_typesupport/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake) multiple times so I won't expect this to fail but we should fix it indeed

jacquelinekay: the package.xml already exports ament_cmake https://github.com/ros2/rosidl/blob/master/rosidl_generator_py/package.xml#L11

Can I just add `ament_export_dependencies(ament_cmake)` to fulfill this requirement?

dirk-thomas: > Can I just add ament_export_dependencies(ament_cmake) to fulfill this requirement?

Yes.

jacquelinekay:
[Build Status](http://ci.ros2.org/job/ci_linux/1500/)
[Build Status](http://ci.ros2.org/job/ci_osx/1165/)
[Build Status](http://ci.ros2.org/job/ci_windows/1507/)

Here's a set of CI jobs running Connext and FastRTPS only.

602772c adds the missing dependency export.

jacquelinekay:
Lots of CI failures that I'm guessing are chalked up to decayed support for Connext and the FastRTPS release, e.g. http://ci.ros2.org/job/ci_linux/1500/testReport/projectroot.test.profiling/memory/memory_profiling/

Since the build was previously broken with this fix I'm going to merge it.

PR #157, opened 2016-06-22
removed unused import_typesupport parameters

mikaelarguedas:
Based on offline discussion with @dirk-thomas we're going to keep a single python extension containing both msg and srv module typesupport functions.
This cleans up the API before starting the work on service generation


dirk-thomas:
+1

PR #158, opened 2016-06-22
add bounded_vector

dirk-thomas:
The new custom container `BoundedVector` uses a template parameter to specify the upper bound because that enables to internally subclass `std::vector` (which safes us from writing a lot more code). If the upper bound would be passed to the constructor it would need to be stored in a member variable which would make the class not a POD anymore (when still subclassing) and that doesn't work with our type support code for introspection which relies on accessing the message fields using `offsetof`.

* http://ci.ros2.org/job/ci_linux/1516/ (compared to master http://ci.ros2.org/job/ci_linux/1515/)
* http://ci.ros2.org/job/ci_osx/1185/ (compared to master http://ci.ros2.org/job/ci_osx/1183/)
* http://ci.ros2.org/job/ci_windows/1519/ (compared to master http://ci.ros2.org/job/ci_windows/1517/)

### DIFF
--- a/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
+++ b/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
@@ -125,20 +125,17 @@ macro(rosidl_generate_interfaces target)
     ament_index_register_resource("rosidl_interfaces" CONTENT "${_idl_files_lines}")
   endif()
 
-  # collect package names of recursive dependencies
+  # collect package names of recursive dependencies which contain interface files
   set(_recursive_dependencies)
   foreach(_dep ${_ARG_DEPENDENCIES})
-    list_append_unique(_recursive_dependencies "${_dep}")
-    foreach(_dep2 ${${_dep}_RECURSIVE_DEPENDENCIES})
-      list_append_unique(_recursive_dependencies "${_dep2}")
-    endforeach()
-  endforeach()
-
-  # check that all dependencies are actually valid message packages
-  foreach(_dep ${_recursive_dependencies})
-    if(NOT DEFINED ${_dep}_INTERFACE_FILES)
-      message(FATAL_ERROR "The message package '${_dep}' does not declare the interface files")
+    if(DEFINED ${_dep}_INTERFACE_FILES)
+      list_append_unique(_recursive_dependencies "${_dep}")
     endif()
+    foreach(_dep2 ${${_dep}_RECURSIVE_DEPENDENCIES})
+      if(DEFINED ${_dep2}_INTERFACE_FILES)
+        list_append_unique(_recursive_dependencies "${_dep2}")
+      endif()
+    endforeach()
   endforeach()
 
   # generators must be executed in topological order
--- a/rosidl_generator_cpp/include/rosidl_generator_cpp/bounded_vector.hpp
+++ b/rosidl_generator_cpp/include/rosidl_generator_cpp/bounded_vector.hpp
@@ -1,0 +1,712 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ROSIDL_GENERATOR_CPP__BOUNDED_VECTOR_HPP_
+#define ROSIDL_GENERATOR_CPP__BOUNDED_VECTOR_HPP_
+
+#include <algorithm>
+#include <memory>
+#include <stdexcept>
+#include <utility>
+#include <vector>
+
+namespace rosidl_generator_cpp
+{
+
+/**
+ *  @brief  A container based on std::vector but with an upper bound.
+ *
+ *  @tparam  _Tp  Type of element.
+ *  @tparam  _UpperBound  The upper bound for the number of elements.
+ *  @tparam  _Alloc  Allocator type, defaults to allocator<_Tp>.
+ *
+ *  Meets the same requirements as std::vector.
+*/
+template<typename _Tp, std::size_t _UpperBound, typename _Alloc = std::allocator<_Tp>>
+class BoundedVector
+  : protected std::vector<_Tp, _Alloc>
+{
+  using _Base = std::vector<_Tp, _Alloc>;
+
+public:
+  using typename _Base::value_type;
+  using typename _Base::pointer;
+  using typename _Base::const_pointer;
+  using typename _Base::reference;
+  using typename _Base::const_reference;
+  using typename _Base::iterator;
+  using typename _Base::const_iterator;
+  using typename _Base::const_reverse_iterator;
+  using typename _Base::reverse_iterator;
+  using typename _Base::size_type;
+  using typename _Base::difference_type;
+  using typename _Base::allocator_type;
+
+  /**
+   *  @brief  Creates a %BoundedVector with no elements.
+   */
+  BoundedVector()
+  noexcept(std::is_nothrow_default_constructible<_Alloc>::value)
+  : _Base()
+  {}
+
+  /**
+   *  @brief  Creates a %BoundedVector with no elements.
+   *  @param  __a  An allocator object.
+   */
+  explicit
+  BoundedVector(
+    const typename _Base::allocator_type & __a)
+  noexcept
+  : _Base(__a)
+  {}
+
+  /**
+   *  @brief  Creates a %BoundedVector with default constructed elements.
+   *  @param  __n  The number of elements to initially create.
+   *  @param  __a  An allocator.
+   *
+   *  This constructor fills the %BoundedVector with @a __n default
+   *  constructed elements.
+   */
+  explicit
+  BoundedVector(
+    typename _Base::size_type __n,
+    const typename _Base::allocator_type & __a = allocator_type())
+  : _Base(__n, __a)
+  {
+    if (__n > _UpperBound) {
+      throw std::length_error("Exceeded upper bound");
+    }
+  }
+
+  /**
+   *  @brief  Creates a %BoundedVector with copies of an exemplar element.
+   *  @param  __n  The number of elements to initially create.
+   *  @param  __value  An element to copy.
+   *  @param  __a  An allocator.
+   *
+   *  This constructor fills the %BoundedVector with @a __n copies of @a __value.
+   */
+  BoundedVector(
+    typename _Base::size_type __n,
+    const typename _Base::value_type & __value,
+    const typename _Base::allocator_type & __a = allocator_type())
+  : _Base(__n, __value, __a)
+  {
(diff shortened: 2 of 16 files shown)


rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
```diff
@@ -125,22 +125,19 @@ macro(rosidl_generate_interfaces target)
...
+    endif()
     foreach(_dep2 ${${_dep}_RECURSIVE_DEPENDENCIES})
-      list_append_unique(_recursive_dependencies "${_dep2}")
+      if(DEFINED ${_dep2}_INTERFACE_FILES)
+        list_append_unique(_recursive_dependencies "${_dep2}")
+      endif()
     endforeach()
   endforeach()
 
-  # check that all dependencies are actually valid message packages
```
dirk-thomas: This is not the case anymore. Now `rosidl_generator_runtime` exports a package which provides additional C++ header file(s) (containing the `BoundedVector`) necessary for the generated C++ messages to work.

rosidl_generator_cpp/test/test_bounded_vector.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+
+TEST(rosidl_generator_cpp, bounded_vector) {
+  rosidl_generator_cpp::BoundedVector<int, 2> v;
+  ASSERT_EQ(v.size(), 0u);
+  ASSERT_EQ(v.max_size(), 2u);
+  v.push_back(1);
+  v.push_back(2);
+  ASSERT_THROW(v.push_back(3), std::length_error);
+  ASSERT_THROW(v.resize(3), std::length_error);
```
dirk-thomas: At least test that the custom container fulfills its purpose in terms of the upper bound.

rosidl_typesupport_introspection_c/resource/msg__type_support.c.em
```diff
@@ -154,7 +154,7 @@ for index, field in enumerate(spec.fields):
     # bool is_array_
     print('    %s,  // is array' % ('true' if field.type.is_array else 'false'))
     # size_t array_size_
-    print('    %u,  // array size' % (field.type.array_size if field.type.array_size and not field.type.is_upper_bound else 0))
```
dirk-thomas: For bounded dynamic arrays the introspection information must report the `array_size` as not `0` but as the *upper bound*.

dirk-thomas:
I fixed the style problem in the `bounded_vector` (http://ci.ros2.org/job/ci_linux/1516/testReport/rosidl_generator_cpp/uncrustify/).

I assume the tests `test_publisher_subscriber_cpp__BoundedArrayPrimitives` with OpenSplice as the publisher (http://ci.ros2.org/job/ci_linux/1516/testReport/(root)/) are failing for the same reason as the `DynamicArrayPrimitves` ones () so I disabled them the same way in CMake.

jacquelinekay:
I thought about it and I begrudgingly accept this path of inheriting from vector.

Inheriting the vector as `protected` is necessary in order to call the destructor of `vector` on destruction of `BoundedVector`.

This requires that the comparison operators use `reinterpret_cast` in the comparison operators instead of `static_cast` because the inheritance relationship is not visible outside of the class. If public inheritance was used then `static_cast` would work, but then I believe the destructor of vector would not be called.

I think you can actually use a `static_cast` on line 214 here because it's within the definition of the class:
https://github.com/ros2/rosidl/pull/158/files#diff-dd92d0c53b2a6e09d8a33848dd7d5ba7R214

Some more unit tests for e.g. the logical operators, assign, swap etc. would be nice. But I'm decently convinced that this will work.

My only concern is now about restricting the size of allocations made by the `BoundedVector`. The strategy used to resize the capacity of `vector` is not defined by the standard, so different compilers implement it differently. I don't think the current interface can restrict the capacity from going over UpperBound, although I think it does guarantee that the capacity will stop growing when the vector's `size` meets UpperBound. The Allocator template argument controls how new capacity is allocated but it does not provide the "policy" for when to change the capacity. You could probably hack around this by making a pool allocator for `BoundedVector` and restricting its size, but `capacity` would be unaware of this restriction.
(for example: http://stackoverflow.com/questions/11560904/vector-memory-allocation-strategy)

Which is fine for most use cases but we should just be aware that this is a limitation that might not work for extremely memory-constrained environments.

dirk-thomas:
I updated the links in the first comment to the latest builds and they all look "good" (in terms of not worse than before).

Windows has several pub / sub tests failing (but they seem all flaky. One build they work, another they don't across all message type (even `Empty` which isn't affected by the changes in these PRs).

On OS X the `tf2` tests fail being unable to find a library. I guess it fails since `rosidl_generator_runtime` doesn't provide the dependencies on the type support packages anymore. Not sure how to fix that yet.

dirk-thomas:
Instead of removing the type support dependencies from `rosidl_generator_runtime` I fixed the order of the include directories being used in all type support packages (1e4b0741cbd82da5e62b0822ab24f293ede12ba4 https://github.com/ros2/rmw_connext/pull/180/commits/730056a5bd3333959614177cba074ed3482548bf https://github.com/ros2/rmw_opensplice/pull/148/commits/020178a42517499ce1e847ba8f6db5ed690ccdb9) to ensure the own headers are being found before the same ones from other type support packages. That fixed the failing tests on OS X: http://ci.ros2.org/job/ci_osx/1191/

Please review this PR as well as all referenced ones.

rosidl_generator_cpp/include/rosidl_generator_cpp/bounded_vector.hpp
```diff
@@ -0,0 +1,712 @@
...
+
+namespace rosidl_generator_cpp
+{
+
+/**
+ *  @brief  A container based on std::vector but with an upper bound.
+ *
+ *  @tparam  _Tp  Type of element.
+ *  @tparam  _UpperBound  The upper bound for the number of elements.
+ *  @tparam  _Alloc  Allocator type, defaults to allocator<_Tp>.
```
wjwwood: nitpick: this isn't consistent with the styles used in rclcpp and rcl. I'm sort of agnostic as to which style we use (this is the Javadoc style, rclcpp/rcl use the "C++" style), but it should be consistent I think.

dirk-thomas: This is the documentation of stl_vector.h on Ubuntu.

wjwwood: This is a completely new class, regardless of whether or not you copied it from somewhere before modifying it. I think it would be better to use a consistent doc style, even in this case. What do others think?

rosidl_generator_cpp/include/rosidl_generator_cpp/bounded_vector.hpp
```diff
@@ -0,0 +1,712 @@
...
+ *
+ *  Meets the same requirements as std::vector.
+*/
```
wjwwood: nitpick: this is not consistent with the rest of the file -- it should be ` */`

rosidl_generator_cpp/include/rosidl_generator_cpp/bounded_vector.hpp
```diff
@@ -0,0 +1,712 @@
...
+ *  @brief  Vector equality comparison.
+ *  @param  __x  A %BoundedVector.
+ *  @param  __y  A %BoundedVector of the same type as @a __x.
+ *  @return  True if the size and elements of the vectors are equal.
+ *
+ *  This is an equivalence relation.  It is linear in the size of the
+ *  vectors.  Vectors are considered equivalent if their sizes are equal,
+ *  and if corresponding elements compare equal.
+*/
+template<typename _Tp, std::size_t _UpperBound, typename _Alloc>
```
wjwwood: Should it be possible to compare `BoundedVector`'s which have different `_UpperBound`'s? Currently I think this code would not compile:

```c++
BoundedVector<int, 32> bv1;
BoundedVector<int, 64> bv2;
bool is_equal = bv1 == bv2;
```

Maybe I'm wrong about that (I haven't tried it). I suppose the same is true if the allocators are different? Maybe that's fine.

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -43,6 +43,7 @@ cpp_full_name_with_alloc = '%s<ContainerAllocator>' % (cpp_full_name)
 #include <memory>
 #include <string>
 #include <vector>
+#include <rosidl_generator_cpp/bounded_vector.hpp>
```
wjwwood: Does this need to be imported for all messages, or just ones that contain fields with bounded arrays?

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
```diff
@@ -114,10 +114,14 @@ def msg_type_to_cpp(type_):
             (type_.pkg_name, type_.type)
 
     if type_.is_array:
-        if type_.array_size is None:
+        if not type_.array_size:
```
wjwwood: I guess that `type_.array_size` cannot be `0`? Is this change important to the functionality of the bounded array or just another way to check for `None`?

rosidl_generator_cpp/include/rosidl_generator_cpp/bounded_vector.hpp
dirk-thomas: Yes, different types can't be compared. I think that is sufficient for now.

rosidl_generator_cpp/test/test_bounded_vector.cpp
```diff
@@ -0,0 +1,32 @@
...
+  v.resize(1);
+  ASSERT_EQ(v.size(), 1u);
+  v = {1, 2};
+  ASSERT_EQ(v.size(), 2u);
```
wjwwood: Is it possible to test an initializer list that's too long? For example, something like `ASSERT_THROW(v = {1, 2, 3}, std::length_error);`?

rosidl_generator_cpp/resource/msg__struct.hpp.em
dirk-thomas: It is only needed for messages which contain bounded arrays. But the same is true for the other includes above, e.g. `array`, `string`, `vector`.

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
dirk-thomas: An array size of zero is not valid. So the check is the same as checking for `None`.

wjwwood:
I had some questions, but otherwise it lgtm.

rosidl_generator_cpp/test/test_bounded_vector.cpp
dirk-thomas: Sure, but the initializer list must be passed as a variable since the macro wouldn't accept the extra commas: 62a0b2fdeb7d73a84c0e2ec769df60cd0a3bd6a8

The added test passes locally. When shorting the initializer list to only two elements makes the test fails as expected.

rosidl_generator_cpp/include/rosidl_generator_cpp/bounded_vector.hpp
dirk-thomas: See ad78f1e731f57762cb55272ce43f1632523bf13f

wjwwood: Thanks that looks great.

dirk-thomas:
Fixed linter warning in 80e6d2a117091ff27d4701d0f898ca3a1d58f2d1.

PR #159, opened 2016-06-30
fix broken assertion

dirk-thomas:
Fix bug from #73.


mikaelarguedas:
lgtm, thanks for fixing it

PR #162, opened 2016-09-20
Update import_type_support docstring

dhood:
Please check that the terminology I've used is correct because I am still wrapping my head around the details of type support.


mikaelarguedas:
after discussing with @dhood we think that this reached the point where it is understandable. So LGTM but I'd like someone else to +1 it before merging.
Thanks!

PR #165, opened 2016-10-18
fix and test default values for all generators

mikaelarguedas:
note : need to add analog testing to c++ to confirm that changes to `rosidl_parser` don't break anything


rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -445,7 +445,7 @@ def parse_value_string(type_, value_string):
     if type_.is_primitive_type() and not type_.is_array:
         return parse_primitive_value_string(type_, value_string)
 
-    if type_.is_primitive_type() and type_.is_array and type_.type != 'string':
+    if type_.is_primitive_type() and type_.is_array:
```
dirk-thomas: The following code doesn't work for strings yet since it doesn't consider quoting / escaping so this can not be enabled yet.

mikaelarguedas: thanks for the comment, I was wondering what the original reason for this condition was.
reverted and added a todo

mikaelarguedas:
note: Bounded strings in c++ don't seem to be bounded:
[this](https://github.com/ros2/rosidl/blob/default_vals/rosidl_generator_cpp/test/test_interfaces.cpp#L192) is passing while it should fail given that max str size is [10](https://github.com/ros2/rosidl/blob/default_vals/rosidl_generator_cpp/msg/PrimitivesBounded.msg#L14)
this test also passes using `ASSERT_STREQ` rather than `ASSERT_EQ`

mikaelarguedas:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1788)

Investigating non related failures on windows before triggering CI for other platforms

dirk-thomas:
* http://ci.ros2.org/job/ci_linux/1795/
* http://ci.ros2.org/job/ci_osx/1351/
* http://ci.ros2.org/job/ci_windows/1718/

mikaelarguedas:
oops just saw that, cancelling the new round of CI I launched for that purpose.
CI:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1796)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1352/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1718/)

can you please review this PR now that all tests are passing ?

rosidl_generator_py/test/test_interfaces.py
```diff
@@ -24,8 +24,8 @@
 def test_strings():
     a = Strings()
 
-    assert(a.empty_string is str())
-    assert(a.def_string == 'Hello world!')
+    assert a.empty_string is str()
```
dirk-thomas: Should this be `str` (without `()`)?

Same below.

rosidl_generator_cpp/msg/PrimitivesConstants.msg
```diff
@@ -0,0 +1,14 @@
+bool BOOL_CONST=true
+byte BYTE_CONST=50
+char CHAR_CONST=100
+float32 FLOAT32_CONST=1.125
+float64 FLOAT64_CONST=3.14159
```
dirk-thomas: For `float32` we selected a value which is representated exactly the same way (without rounding) in Python and C++. I guess this is not the case for this one. Even if it is only tested in C++ should we try to use a similar "portable" value?

rosidl_generator_py/test/test_interfaces.py
mikaelarguedas: I think the goal here was to confirm that it was the language default initialization value which is more restrictive that checking the type. Would it be less confusing to use `assert a.empty_string is ''` ?

rosidl_generator_cpp/msg/PrimitivesConstants.msg
mikaelarguedas: This is the value we use for `test_communication` because it has the exact same representation in both languages (c.f. https://github.com/ros2/system_tests/commit/0e2dad0d2f4570a7f097c1a58e33433eb815d8cc)
So I think this should be ok

rosidl_generator_py/test/test_interfaces.py
dirk-thomas: Absolutely. I didn't read it like that. I will update it to `''`.

rosidl_generator_cpp/msg/PrimitivesConstants.msg
dirk-thomas: But this line uses `3.14159` and not was we did in the reference patch to test_communication.

mikaelarguedas: oh sorry I misread your comment, I'll update it the the value I used in `test_communication` for multi_rcl testing. thanks for pointing it out!

PR #171, opened 2016-11-02
[rosidl_generator_py] Service generation

mikaelarguedas:
connects to https://github.com/ros2/rosidl/issues/67

### DIFF
--- a/rosidl_generator_py/resource/_msg_support.entry_point.c.em
+++ b/rosidl_generator_py/resource/_msg_support.entry_point.c.em
@@ -7,11 +7,10 @@
 @{
 static_includes = {}
 for spec, subfolder in message_specs:
-  if subfolder not in static_includes.keys():
-    if subfolder == 'msg':
-      static_includes[subfolder] = '#include <rosidl_generator_c/message_type_support.h>'
-    elif subfolder == 'srv':
-      static_includes[subfolder] = '#include <rosidl_generator_c/service_type_support.h>'
+  if subfolder == 'msg':
+    static_includes[subfolder] = '#include <rosidl_generator_c/message_type_support.h>'
+  elif subfolder == 'srv':
+    static_includes[subfolder] = '#include <rosidl_generator_c/service_type_support.h>'
 for value in sorted(static_includes.values()):
   print(value)
 }@
@@ -22,15 +21,13 @@ for spec, subfolder in message_specs:
   type_name = spec.base_type.type
   module_name = convert_camel_case_to_lower_case_underscore(type_name)
   key = '%s/%s/%s' % (spec.base_type.pkg_name, subfolder, module_name)
-  if key not in static_includes.keys():
-    includes[key] = '#include <%s__type_support.h>' % key
+  includes[key] = '#include <%s__type_support.h>' % key
 
 for spec, subfolder in service_specs:
   type_name = convert_camel_case_to_lower_case_underscore(spec.srv_name)
   module_name = convert_camel_case_to_lower_case_underscore(type_name)
   key = '%s/%s/%s' % (spec.pkg_name, subfolder, module_name)
-  if key not in static_includes.keys():
-    includes[key] = '#include <%s.h>' % key
+  includes[key] = '#include <%s.h>' % key
 for v in sorted(includes.values()):
   print(v)
 }@
--- a/rosidl_generator_py/resource/_msg_support.entry_point.c.em
+++ b/rosidl_generator_py/resource/_msg_support.entry_point.c.em
@@ -11,9 +11,10 @@ for spec, subfolder in message_specs:
     static_includes[subfolder] = '#include <rosidl_generator_c/message_type_support.h>'
   elif subfolder == 'srv':
     static_includes[subfolder] = '#include <rosidl_generator_c/service_type_support.h>'
-for value in sorted(static_includes.values()):
-  print(value)
 }@
+@[for value in sorted(static_includes.values())]@
+@(value)
+@[end for]@
 
 @{
 includes = {}
@@ -28,10 +29,10 @@ for spec, subfolder in service_specs:
   module_name = convert_camel_case_to_lower_case_underscore(type_name)
   key = '%s/%s/%s' % (spec.pkg_name, subfolder, module_name)
   includes[key] = '#include <%s.h>' % key
-for v in sorted(includes.values()):
-  print(v)
 }@
-
+@[for v in sorted(includes.values())]@
+@(v)
+@[end for]@
 @[for spec, subfolder in message_specs]@
 @{
 type_name = spec.base_type.type
--- a/rosidl_generator_py/resource/_msg_support.entry_point.c.em
+++ b/rosidl_generator_py/resource/_msg_support.entry_point.c.em
@@ -4,7 +4,17 @@
 #include <Python.h>
 #include <stdint.h>
 
-#include <rosidl_generator_c/message_type_support.h>
+@{
+includes = {}
+for spec, subfolder in message_specs:
+  if subfolder not in includes.keys():
+    if subfolder == 'msg':
+      includes['msg'] = '#include <rosidl_generator_c/message_type_support.h>'
+    elif subfolder == 'srv':
+      includes['srv'] = '#include <rosidl_generator_c/service_type_support.h>'
+for value in sorted(includes.values()):
+  print(value)
+}@
 
 @[for spec, subfolder in message_specs]@
 @{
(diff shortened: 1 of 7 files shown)


rosidl_generator_py/resource/_srv.py.em
```diff
@@ -1,7 +1,7 @@
...
 # generated code does not contain a copyright notice
 
-from @(package_name).srv._@(spec.srv_name)__request import @(spec.srv_name)_Request
-from @(package_name).srv._@(spec.srv_name)__response import @(spec.srv_name)_Response
-@(spec.srv_name)_Request  # unused
-@(spec.srv_name)_Response  # unused
+
+class @(spec.srv_name):
+    from @(package_name).srv._@convert_camel_case_to_lower_case_underscore(spec.srv_name)__request import @(spec.srv_name)_Request as Request
+    from @(package_name).srv._@convert_camel_case_to_lower_case_underscore(spec.srv_name)__response import @(spec.srv_name)_Response as Response
```
dirk-thomas: Maybe add something to prevent instantiation, e.g. `def __init__(self): raise NotImplementedError('...')`

mikaelarguedas: Good point, thanks

rosidl_generator_py/resource/_msg_support.entry_point.c.em
```diff
@@ -4,15 +4,36 @@
 #include <Python.h>
 #include <stdint.h>
 
-#include <rosidl_generator_c/message_type_support.h>
+@{
+static_includes = {}
+for spec, subfolder in message_specs:
+  if subfolder not in static_includes.keys():
```
dirk-thomas: This is redundant. Even without the check the content of the dict is the same. Therefore I would suggest to remove it for shorter, easier to read code.

Same below.

rosidl_generator_py/resource/_srv.py.em
```diff
@@ -1,7 +1,43 @@
...
+        try:
+            import rclpy
+            from rosidl_generator_py import import_type_support
+            __type_support_importable = True
+        except ImportError:
+            logger = logging.getLogger('rosidl_generator_py.@(spec.srv_name)')
+            logger.debug(
+                'Failed to import needed modules for type support:\n' + traceback.format_exc())
+
+        if __type_support_importable:
```
dirk-thomas: Instead of using a custom variable this could just use `else` instead.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -97,16 +99,20 @@ def generate_py(generator_arguments_file, typesupport_impls):
     for module in modules:
         with open(os.path.join(args['output_dir'], module, '__init__.py'), 'w') as f:
             for module_, type_ in modules[module]:
-                f.write('from %s.msg._%s import %s\n' % (args['package_name'], module_, type_))
+                if type_.find('Request') == -1 and type_.find('Response') == -1:
```
dirk-thomas: This condition is likely matching more cases then just request and response class since the substring can be anywhere.

rosidl_generator_py/resource/_srv.py.em
mikaelarguedas: you mean putting everything in the try block, right ?

dirk-thomas: No, putting the code following this line in an `else` block. The `try` block should always be as small as possible in order to not catch exceptions which shouldn't be catched.

rosidl_generator_py/resource/_msg_support.entry_point.c.em
```diff
@@ -4,15 +4,33 @@
...
-#include <rosidl_generator_c/message_type_support.h>
+@{
+static_includes = {}
+for spec, subfolder in message_specs:
+  if subfolder == 'msg':
+    static_includes[subfolder] = '#include <rosidl_generator_c/message_type_support.h>'
+  elif subfolder == 'srv':
+    static_includes[subfolder] = '#include <rosidl_generator_c/service_type_support.h>'
+for value in sorted(static_includes.values()):
+  print(value)
```
wjwwood: nitpick: I prefer to see the code templated it out, i.e. `@[for value in sorted(static_includes.values())]@...`, rather than using `print` within the `@{}@`. But that's just my preference, as I tend to skip over `@{}@` sections when reading, assuming they are just setting up context. You don't have to change this unless you also think it's a good idea.

rosidl_generator_py/resource/_msg_support.entry_point.c.em
```diff
@@ -4,15 +4,33 @@
...
+  key = '%s/%s/%s' % (spec.base_type.pkg_name, subfolder, module_name)
+  includes[key] = '#include <%s__type_support.h>' % key
+
+for spec, subfolder in service_specs:
+  type_name = convert_camel_case_to_lower_case_underscore(spec.srv_name)
+  module_name = convert_camel_case_to_lower_case_underscore(type_name)
+  key = '%s/%s/%s' % (spec.pkg_name, subfolder, module_name)
+  includes[key] = '#include <%s.h>' % key
+for v in sorted(includes.values()):
+  print(v)
```
wjwwood: Same nitpick as above.

rosidl_generator_py/resource/_msg_support.entry_point.c.em
mikaelarguedas: I agree I prefer the generated code to be out of the @{}@ statement that is used to set variables/context. fixed in https://github.com/ros2/rosidl/pull/171/commits/2501e7d3dceb00a27d121f2815c46a4e987b569b

rosidl_generator_py/resource/_srv.py.em
mikaelarguedas: I was not familiar with the try-except-else pattenr, fixed in https://github.com/ros2/rosidl/pull/171/commits/0d35a5ac70b82db44d380ab5a809f43fc3713a1f

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -97,16 +99,20 @@ def generate_py(generator_arguments_file, typesupport_impls):
     for module in modules:
         with open(os.path.join(args['output_dir'], module, '__init__.py'), 'w') as f:
             for module_, type_ in modules[module]:
-                f.write('from %s.msg._%s import %s\n' % (args['package_name'], module_, type_))
+                if type_.endswith('Request') is False and type_.endswith('Response') is False:
```
dirk-thomas: Should this only match files in the `srv` folder?

mikaelarguedas: good catch: https://github.com/ros2/rosidl/pull/171/commits/8ab1a47fa1b0dd018d362b759d7152df8349c13f

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -95,18 +97,27 @@ def generate_py(generator_arguments_file, typesupport_impls):
                     minimum_timestamp=latest_target_timestamp)
 
     for module in modules:
+        importlist = {}
+        for module_, type_ in modules[module]:
+            if module == 'srv' and (
+                    type_.endswith('Request') is True or type_.endswith('Response') is True):
```
dirk-thomas: `is True` is redundant.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -95,18 +97,27 @@ def generate_py(generator_arguments_file, typesupport_impls):
                     minimum_timestamp=latest_target_timestamp)
 
     for module in modules:
+        importlist = {}
+        for module_, type_ in modules[module]:
```
dirk-thomas: The variable names `module_` and `module` don't provide enough information for readers to know what they contain / how they are different.

rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -95,18 +97,27 @@ def generate_py(generator_arguments_file, typesupport_impls):
...
+
         with open(os.path.join(args['output_dir'], module, '__init__.py'), 'w') as f:
-            for module_, type_ in modules[module]:
-                f.write('from %s.msg._%s import %s\n' % (args['package_name'], module_, type_))
-            for module_, type_ in modules[module]:
-                f.write('%s  # unused\n' % type_)
+            for importline in sorted(importlist.values()):
+                f.write(importline)
+            for noqaline in sorted(importlist.keys()):
+                        f.write(noqaline)
```
dirk-thomas: I would suggest adding underscores to these variables: `import_list`, `import_line`, `noqa_line`

mikaelarguedas:
I updated variable names and conditions in https://github.com/ros2/rosidl/pull/171/commits/add5ed670f51be0014fdc206b781251ea1400ab7

dirk-thomas:
The new names are much better :+1:

PR #174, opened 2016-11-07
line length

mikaelarguedas:
adjust line length form  #173 to make linter tests pass


dirk-thomas:
Thank for addressing this. I modified the lines to actually be shorter then 8 chars. The linter just doesn't detect some cases (like the ones still in the file a few lines above).

mikaelarguedas:
yeah I noticed that the first line is not linted in general. Thanks for the iteration

PR #179, opened 2016-12-04
String tests

mikaelarguedas:
This PR adds:
 * [rosidl_generator_py]
   * length enforcement for bounded string in array in Python and the relevant unit tests
 * [rosidl_generator_cpp]
   * a function to generate random string arrays/vector/boundedvector in c++
   * msg files and tests for all String/StringArray types
   * tests for PrimitivesStaticArrays that were not tested before
   * a disabled test for testing bounded string assignment (we currently don't have any way to restrict std::string length)
  * [rosidl_generator_c]
    * a TODO to remind us to test bounded strings once length restriction is enforced


rosidl_generator_cpp/test/test_array_generator.hpp
```diff
@@ -139,4 +140,42 @@ void test_vector_fill(C * container, size_t size,
...
+    std::is_same<typename C::value_type, typename std::string>::value
+  >::type * = nullptr
+>
+void test_vector_fill(C * container, size_t size,
+  int min, int max,
+  int minlength, const int maxlength)
+{
+  std::default_random_engine rand_generator;
+  std::uniform_int_distribution<int> randnum(min, max);
+  std::uniform_int_distribution<int> randlen(minlength, maxlength);
```
dirk-thomas: I would recommend not to use random numbers for any tests. The problem is that the test will be flaky of the result depends on the values. Instead use a "known" pattern, e.g. increment numbers, or ascii characters.

mikaelarguedas: ticketed here https://github.com/ros2/rosidl/issues/181

wjwwood: +1 for not using fuzz values. I'm ok with doing it in a follow up if that's what you want.

mikaelarguedas:
Yeah I'll open a follow-up not to add noise to thise one.
Thanks for reviewing

PR #180, opened 2016-12-05
[rosidl_generator_c] convert to use gtest

mikaelarguedas:
Currently none of the tests in this package generate result files so we have no indicator if they are failing.
This PR converts the tests to use gtest, so that both Jenkins and `ament test_results` are aware of them.
NOTE: maybe it's overkill to now build these as C++ while this package was doing C compilation only. I'm open to suggestions if there is a less intrusive alternative

connects to https://github.com/ros2/ros2/issues/290


dirk-thomas:
There is no need to artificially modify the type of the test. Any executable can be registered using `ament_add_test` and ament will generate the result file based on the return value.

mikaelarguedas:
Yeah I agree that the tests could be registered using `ament_add_test`.
One of the motivation for the modification was that given the current layout of the test. any failure results in `exit(1)` and thus ignoring all the following tests in the same file. Using gtest seemed to be an easy way to have both better error messages and ensuring that all tests are ran.

PR #182, opened 2016-12-07
typesupport reloaded

dirk-thomas:
This together with the other linked PRs switches to use the new `rosidl_typesupport_cpp` in C++. It only stores a map of typesupport names to their respective library / symbol name. It will then on-demand load the library using poco.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/1991/)
* Mac OS [Build Status](http://ci.ros2.org/job/ci_osx/1532/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/1963/)

After this has been merged there will be a few follow up rounds:
* Do the same for C typesupport.
* Invert the package order so that  `rosidl_typesupport_cpp` is being built after all rmw implementations. If only a single rmw implementation is available it can then generate a shortcut which avoids poco.
* Do the same indirection / dispatcher for rmw itself.

Ready for review (either comment on all linked PRs individually or just here with a note that it covers all PRs).

### DIFF
--- a/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
+++ b/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
@@ -31,6 +31,9 @@ include(CMakeParseArguments)
 # :param DEPENDENCIES: the packages from which message types are
 #   being used
 # :type DEPENDENCIES: list of strings
+# :param LIBRARY_NAME: the base name of the library, specific generators might
+#   append their own suffix
+# :type LIBRARY_NAME: string
 # :param SKIP_INSTALL: if set skip installing the interface files
 # :type SKIP_INSTALL: option
 # :param ADD_LINTER_TESTS: if set lint the interface files using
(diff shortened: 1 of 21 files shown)


dirk-thomas:
The Windows job is finally :green_apple: :green_book: :four_leaf_clover: :sparkle: :white_check_mark:

Karsten1987:
Looks good to me (as far as I can understand it :smile: )
One thing though: Is the shortcut in case only one ts is available implemented? 

If the buildfarm doesn't complain and all system tests are running correctly, we could merge that from my point. I only feel that this change needs a lot of incode documentation in order to understand what's happening here.

dirk-thomas:
The shortcut isn't implemented yet. The [beta ticket](ros2/ros2#281) outlines what needs to be done before that. Namely the same for C typesupport. Only then the dependency order can be inverted so that the `rosidl_typesupport_*` packages can depend on all rmw implementation in order to decide if it can shortcut (aka only one rmw impl is available).

I definitely agree with the need for documentation. I would suggest to first finish the next implementation levels and then document the final design / implementation.

The remaining question is: should we merge this as-is or wait for the remaining steps to be implemented. I see the following arguments pro / con merging now:

:heavy_plus_sign: The patch set took already more then two weeks to create and is pretty complex to review. Waiting for the next two levels of implementation will likely require another couple of weeks and will increase the complexity for the review even further. And that is beside the potential of diverging from master and requiring additional rebasing effort.

:heavy_minus_sign: When there is only a single rmw impl available the typesupport will still use `dlopen` to open the specific typesupport library at runtime. In the final implementation that should be resolved by avoiding the mapping / dlopen in that case.

tfoote:
If it's expected to work in all cases, but there's room to optimize the future behavior (aka implement the shortcut). It would make sense to merge to avoid divergence which will make more work to rebase in the future, and ticket/track the feature of the optimization separately.

dirk-thomas:
Then I will move forward an merging this set of PRs this afternoon if nobody raises any concerns.

dirk-thomas:
Continued in ros2/ros2#302...

PR #183, opened 2016-12-08
Control shared/static linking via BUILD_SHARED_LIBS

esteve:
This PR delegates the decision to build a shared or a static library to CMake's BUILD_SHARED_LIBS

Update:

Connects to https://github.com/ros2/ros2/issues/306

### DIFF
--- a/rosidl_generator_c/CMakeLists.txt
+++ b/rosidl_generator_c/CMakeLists.txt
@@ -11,12 +11,12 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
-find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_python REQUIRED)
+find_package(ament_cmake_ros REQUIRED)
 find_package(rosidl_typesupport_interface REQUIRED)
 
 include_directories(include)
-add_library(${PROJECT_NAME} SHARED
+add_library(${PROJECT_NAME}
   "src/message_type_support.c"
   "src/primitives_array_functions.c"
   "src/service_type_support.c"
--- a/rosidl_generator_c/package.xml
+++ b/rosidl_generator_c/package.xml
@@ -7,11 +7,11 @@
   <maintainer email="william@osrfoundation.org">William Woodall</maintainer>
   <license>Apache License 2.0</license>
 
-  <buildtool_depend>ament_cmake</buildtool_depend>
+  <buildtool_depend>ament_cmake_ros</buildtool_depend>
 
   <build_depend>rosidl_typesupport_interface</build_depend>
 
-  <buildtool_export_depend>ament_cmake</buildtool_export_depend>
+  <buildtool_export_depend>ament_cmake_ros</buildtool_export_depend>
   <buildtool_export_depend>rosidl_cmake</buildtool_export_depend>
 
   <build_export_depend>rosidl_typesupport_interface</build_export_depend>
--- a/rosidl_generator_c/package.xml
+++ b/rosidl_generator_c/package.xml
@@ -11,7 +11,7 @@
 
   <build_depend>rosidl_typesupport_interface</build_depend>
 
-  <buildtool_export_depend>ament_cmake_ros</buildtool_export_depend>
+  <buildtool_export_depend>ament_cmake</buildtool_export_depend>
   <buildtool_export_depend>rosidl_cmake</buildtool_export_depend>
 
   <build_export_depend>rosidl_typesupport_interface</build_export_depend>
--- a/rosidl_typesupport_introspection_c/cmake/rosidl_typesupport_introspection_c_generate_interfaces.cmake
+++ b/rosidl_typesupport_introspection_c/cmake/rosidl_typesupport_introspection_c_generate_interfaces.cmake
@@ -114,7 +114,7 @@ list(APPEND _generated_msg_header_files "${_visibility_control_file}")
 
 set(_target_suffix "__rosidl_typesupport_introspection_c")
 
-add_library(${rosidl_generate_interfaces_TARGET}${_target_suffix} SHARED
+add_library(${rosidl_generate_interfaces_TARGET}${_target_suffix}
   ${_generated_msg_header_files} ${_generated_msg_source_files}
   ${_generated_srv_header_files} ${_generated_srv_source_files})
 if(rosidl_generate_interfaces_LIBRARY_NAME)
--- a/rosidl_typesupport_introspection_cpp/cmake/rosidl_typesupport_introspection_cpp_generate_interfaces.cmake
+++ b/rosidl_typesupport_introspection_cpp/cmake/rosidl_typesupport_introspection_cpp_generate_interfaces.cmake
@@ -94,7 +94,7 @@ add_custom_command(
 
 set(_target_suffix "__rosidl_typesupport_introspection_cpp")
 
-add_library(${rosidl_generate_interfaces_TARGET}${_target_suffix} SHARED
+add_library(${rosidl_generate_interfaces_TARGET}${_target_suffix}
   ${_generated_msg_header_files} ${_generated_msg_source_files}
   ${_generated_srv_header_files} ${_generated_srv_source_files})
 if(rosidl_generate_interfaces_LIBRARY_NAME)
--- a/rosidl_generator_c/cmake/rosidl_generator_c_generate_interfaces.cmake
+++ b/rosidl_generator_c/cmake/rosidl_generator_c_generate_interfaces.cmake
@@ -122,7 +122,7 @@ list(APPEND _generated_msg_headers "${_visibility_control_file}")
 
 set(_target_suffix "__rosidl_generator_c")
 
-add_library(${rosidl_generate_interfaces_TARGET}${_target_suffix} SHARED
+add_library(${rosidl_generate_interfaces_TARGET}${_target_suffix}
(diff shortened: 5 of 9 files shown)


rosidl_generator_c/CMakeLists.txt
```diff
@@ -8,10 +8,11 @@ endif()
 
 find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_python REQUIRED)
+find_package(ament_cmake_ros REQUIRED)
```
dirk-thomas: With `ament_cmake_ros` being added you can remove `ament_cmake` (which is a transitive dependency) to keep the CMake file simpler.

rosidl_generator_c/package.xml
```diff
@@ -9,6 +9,7 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
+  <build_depend>ament_cmake_ros</build_depend>
```
dirk-thomas: This needs to be a `buildtool_depend` dependency.

With the above comment the `buildtool_depend` entry for `ament_cmake` can be removed.

rosidl_generator_c/CMakeLists.txt
esteve: Done.

rosidl_generator_c/package.xml
esteve: Done.

esteve:
I've also updated all the other PRs to use `ament_cmake_ros`

dirk-thomas:
The typesupport libraries are expected to be shared when there is more then one available since they are being loaded using `dlopen`. When building static libraries but more than one typesupport is being used by `rosidl_typesupport_c|cpp` I would assume it fails at least at runtime. Have you tried that use case? I think it would be good if this case is being catched at build time to let the user know they are trying an impossible combination.

esteve:
@dirk-thomas sorry, I forgot about this. Unfortunately I've only tested this with `rosidl_typesupport_introspection_c` and `rosidl_typesupport_introspection_cpp` through Fast-RTPS.

The changes need to happen somewhere around here, is that correct?

https://github.com/ros2/rosidl_typesupport/blob/master/rosidl_typesupport_c/cmake/rosidl_typesupport_c_generate_interfaces.cmake#L115-L126
https://github.com/ros2/rosidl_typesupport/blob/master/rosidl_typesupport_cpp/cmake/rosidl_typesupport_cpp_generate_interfaces.cmake#L101-L112

so that if there's more than one typesupport and the `BUILD_SHARED_LIBS` option is set to `OFF`, CMake would bail out, right? If so, let me know and I'll submit a PR to abort or warn the user that they can't use static libraries with more than one typesupport.

Thanks.

dirk-thomas:
I am not sure if there is any use in building all libraries static when requested by still build the typesupport libraries shared if more than one is available?

If that is the case it would be best to choose the library type in each typesupport package to either use `BUILD_SHARED_LIBS` implicitly or if there is multiple typesupport explicitly build the typesupport libraries as `SHARED.

If not, the locations you referenced look like the right place to implement the bailing out.

wjwwood:
> I am not sure if there is any use in building all libraries static when requested by still build the typesupport libraries shared if more than one is available?

Can you rephrase this, I don't exactly understand what you're saying. Maybe that it doesn't make sense to build static libraries if one than one type support is requested/possible?

wjwwood:
edited tha last post, to fix my own typo

dirk-thomas:
> Can you rephrase this, I don't exactly understand what you're saying. Maybe that it doesn't make sense to build static libraries if more than one type support is requested/possible?

The request to build static libraries is problematic if multiple typesupports are available. Multiple typesupports can only be supported with shared libraries. The question is what should happen in that case:

* build typesupport shared despite the request for static libraries
* build static typesupport which implies selecting a single one (which one? how to notify the user?)
* fail the build and ask the user to either switch to shared libraries or reduce the set of typesupports to only one

wjwwood:
When you say multiple type support do you mean `rosidl_typesupport_introspection_c` and `rosidl_typesupport_introspection_cpp` would cause this issue or is it only an issue when "instrospection" and "connext static" (for example) are both requested? Put another way, is this an issue when you have C and C++ support or only when you request two rmw implementations that use two different type support schemes?

> - build typesupport shared despite the request for static libraries

Is that an acceptable solution @esteve? It seems to me like it wouldn't be in this case.

dirk-thomas:
> When you say multiple type support do you mean rosidl_typesupport_introspection_c and rosidl_typesupport_introspection_cpp would cause this issue or is it only an issue when "instrospection" and "connext static" (for example) are both requested? Put another way, is this an issue when you have C and C++ support or only when you request two rmw implementations that use two different type support schemes?

The typesupport dispatch is happening in `rosidl_typesupport_c` as well as in `rosidl_typesupport_cpp`.  Each of them can only use a single typesupport when linking statically.

So having `rosidl_typesupport_introspection_c` and `rosidl_typesupport_introspection_cpp` available at the same time should not be a problem. But having more than one typesupport packages for the same language would be such a conflict.

wjwwood:
> But having more than one typesupport packages for the same language would be such a conflict.

Thanks for clarify, that's important distinction. In that case I think this solution:

> - fail the build and ask the user to either switch to shared libraries or reduce the set of typesupports to only one

Is favorable.

dirk-thomas:
@esteve Can you please comment here once comments have been addressed. Thanks!

esteve:
@dirk-thomas @wjwwood I prefer to fail the build, instead of picking one typesupport implicitly or building shared libraries inconditionally. I've submitted https://github.com/ros2/rosidl_typesupport/pull/10 which aborts the build if there are multiple typesupports and the user wants to build ROS 2.0 statically (i.e. `BUILD_SHARED_LIBS` is disabled), let me know if that'd be fine. Thanks!

dirk-thomas:
> I prefer to fail the build, instead of picking one typesupport implicitly or building shared libraries inconditionally.

Sounds good to me.

rosidl_generator_c/package.xml
```diff
@@ -7,11 +7,11 @@
   <maintainer email="william@osrfoundation.org">William Woodall</maintainer>
   <license>Apache License 2.0</license>
 
-  <buildtool_depend>ament_cmake</buildtool_depend>
+  <buildtool_depend>ament_cmake_ros</buildtool_depend>
 
   <build_depend>rosidl_typesupport_interface</build_depend>
 
-  <buildtool_export_depend>ament_cmake</buildtool_export_depend>
+  <buildtool_export_depend>ament_cmake_ros</buildtool_export_depend>
```
dirk-thomas: Why is this needed?

Same below.

esteve: I incorrectly assumed that ament_cmake_ros had to be exported to other packages, but it doesn't seem to be the case. Removed in https://github.com/ros2/rosidl/pull/183/commits/f6bc5200d3ecf10129d8d81283822614516d3a67

rosidl_typesupport_introspection_c/cmake/rosidl_typesupport_introspection_c_generate_interfaces.cmake
```diff
@@ -114,7 +114,7 @@ list(APPEND _generated_msg_header_files "${_visibility_control_file}")
 
 set(_target_suffix "__rosidl_typesupport_introspection_c")
 
-add_library(${rosidl_generate_interfaces_TARGET}${_target_suffix} SHARED
+add_library(${rosidl_generate_interfaces_TARGET}${_target_suffix}
```
dirk-thomas: Packages not using `ament_cmake_ros` will build this library static. Isn't that a problem?

rosidl_typesupport_introspection_cpp/cmake/rosidl_typesupport_introspection_cpp_generate_interfaces.cmake
```diff
@@ -94,7 +94,7 @@ add_custom_command(
 
 set(_target_suffix "__rosidl_typesupport_introspection_cpp")
 
-add_library(${rosidl_generate_interfaces_TARGET}${_target_suffix} SHARED
+add_library(${rosidl_generate_interfaces_TARGET}${_target_suffix}
```
dirk-thomas: Same here.

rosidl_typesupport_introspection_c/cmake/rosidl_typesupport_introspection_c_generate_interfaces.cmake
esteve: That's true, shall I update the messages packages so that they use `ament_cmake_ros` as well? Thanks for noticing.

dirk-thomas:
Any news on this?

esteve:
@dirk-thomas I was waiting on a response to https://github.com/ros2/rosidl/pull/183#discussion_r114950597 Do you think that updating the message packages to use `ament_cmake_ros` would be a valid solution to avoid the problem you described?

dirk-thomas:
> Do you think that updating the message packages to use ament_cmake_ros would be a valid solution to avoid the problem you described?

I can't say for sure. I see the potential problem if parts of the system are being build with one setting and another part with the opposite. While that might not need to be a use case which works in all possible combinations the behavior should at least be reasonable for the user (e.g. fail with an error message if the combination can't work).

mikaelarguedas:
@clalancette why is this tagged as untargeted?
I think we still want to do this for microcontrollers as well as embedded applications like android/iOS that require static linking

esteve:
Closing in favor of #266 , thanks @dirk-thomas

ros-discourse:

https://discourse.ros.org/t/announcing-objective-c-on-macos-and-ios-for-ros2/946/1

Issue #181, opened 2016-12-06
[rosidl_generator_cpp] Use known patterns instead of random ones for reproducability


mikaelarguedas:
see https://github.com/ros2/rosidl/pull/179#discussion_r91103083

PR #184, opened 2016-12-08
[rosidl_generator_cpp] Deterministic testing

mikaelarguedas:
This PR gets rid of fuzzy testing to get deterministic / reproducible tests.
connects to https://github.com/ros2/rosidl/issues/181

### DIFF
--- a/rosidl_generator_cpp/test/test_array_generator.hpp
+++ b/rosidl_generator_cpp/test/test_array_generator.hpp
@@ -47,7 +47,7 @@ void test_vector_fill(C * container, size_t size, bool val1 = true,
 /**
  * Helper function to generate a test pattern for integer number types.
  * The template type parameter must be an integer number type.
- * Mininum and maximum values for the type and random values in the middle.
+ * Mininum and maximum values for the type and evenly distributed values in the middle.
  * @param C Container (vector, array, etc) to be filled
  * @param size How many elements to fill in. Must size<=container_size
  * @param min Minimum value in the range to fill.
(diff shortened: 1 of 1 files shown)


rosidl_generator_cpp/test/test_array_generator.hpp
```diff
@@ -57,63 +57,25 @@ template<
...
-  typename C::value_type min, typename C::value_type max)
-{
-  std::default_random_engine rand_generator;
-  std::uniform_int_distribution<int> randnum(static_cast<int>(min), static_cast<int>(max));
-
-  if (size > 0) {
-    (*container)[0] = min;
-    for (size_t i = 1; i < size - 1; i++) {
-      (*container)[i] = static_cast<typename C::value_type>(randnum(rand_generator));
+      (*container)[i] = min + i * step;
```
dirk-thomas: The actual `max` value isn't reached.This is basically due to the fact that rounding can occur for `step` which will make the values not cover the whole range. But I would expect it to be crucial that `max` is explicitly tested.

mikaelarguedas: please check the next line

dirk-thomas: Oh, I see. Too far away on this page :wink: Maybe remove the phrase "evenly distributed" from the docblock above then?

mikaelarguedas: I think the docstring states that it contains minimum and maximum values and uses evenly distributed values to opopulate the rest of the array. I'm open to any wording suggestion to make this clearer. I didn't modify the original wording because I didn't come up with anything significantly better

dirk-thomas: This patch introduce the term "evenly distributed" to the docstring. But the values are not actually evenly distributed.

mikaelarguedas: done, thanks

PR #186, opened 2016-12-09
[rosidl_generator_py] enforce value range

mikaelarguedas:
Currently we never check if the values fit and we cast them at the C level before serialization without even throwing a warning.
This prevents users from setting out of range values. This change applies only to message fields, not default values or constants because the parser already enforce these.

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -205,7 +205,7 @@ bound = 2**nbits
 }@
              all([val >= 0 and val < @(bound) for val in value]))
 @[      elif field.type.type == 'char']@
-             all([ord(val) >= -128 and val < 128 for val in value]))
+             all([ord(val) >= -128 and ord(val) < 128 for val in value]))
 @[      else]@
              True)
 @[      end if]@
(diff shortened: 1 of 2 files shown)


rosidl_generator_py/resource/_msg.py.em
```diff
@@ -212,6 +212,19 @@ class @(spec.base_type.type)(metaclass=Metaclass):
         'int64', 'uint64',
         'string',
     ]]@
+@[    if field.type.type.startswith('int')]@
+@{
+nbytes = int(field.type.type[field.type.type.rfind('t') + 1:])
```
dirk-thomas: `nbytes` should be `nbits`.

Instead of `field.type.type.rfind('t') + 1` you can just use `4` since you already ensured the `startswith` before. (`5` in the `uint` case below.)

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -212,6 +212,19 @@ class @(spec.base_type.type)(metaclass=Metaclass):
...
+bound = 2**(nbytes - 1)
+}@
+            (value >= -@(bound) and value < @(bound)) and \
+@[    elif field.type.type.startswith('uint')]@
+@{
+nbytes = int(field.type.type[field.type.type.rfind('t') + 1:])
+bound = 2**nbytes
+}@
+            (value >= 0 and value < @(bound)) and \
+@[    end if]@
```
dirk-thomas: The range check should come after the `isinstance` check from the next line.

mikaelarguedas:
addressed the comments and added same checks in arrays + relevant tests

PR #189, opened 2016-12-12
Unused includes

mikaelarguedas:
Stumbled upon this recently. May not be the only ones in that package but it's a start

### DIFF
--- a/rosidl_generator_cpp/resource/msg__traits.hpp.em
+++ b/rosidl_generator_cpp/resource/msg__traits.hpp.em
@@ -29,10 +29,6 @@ cpp_namespace = '%s::%s::' % (spec.base_type.pkg_name, subfolder)
 }@
 #include <stdint.h>
 
-#include <array>
-#include <string>
-#include <vector>
-
 namespace rosidl_generator_traits
 {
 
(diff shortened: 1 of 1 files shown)


rosidl_generator_cpp/resource/msg__traits.hpp.em
```diff
@@ -29,10 +29,6 @@ cpp_namespace = '%s::%s::' % (spec.base_type.pkg_name, subfolder)
 }@
 #include <stdint.h>
 
-#include <array>
-#include <string>
-#include <vector>
```
wjwwood: You might need to add `#include <type_traits>` here, since `std::integral_constant` is used below, see:

http://en.cppreference.com/w/cpp/types/integral_constant

I don't know if maybe one of these was previously bringing that symbol in for you.

mikaelarguedas:
CI:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/2058)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1591/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/2074/)

PR #197, opened 2017-01-03
Support default values for string arrays

mikaelarguedas:
connects to https://github.com/ros2/rosidl/issues/206
to be merged along https://github.com/ros2/design/pull/111

CI:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/2172/)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1654/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/2174/)


rosidl_generator_c/resource/msg__functions.c.em
```diff
@@ -115,6 +115,21 @@ for field in spec.fields:
...
+            if field.default_value is not None:
+                for i, default_value in enumerate(field.default_value):
+                    if field.type.type == 'string':
+                        lines.append('{')
+                        lines.append('  bool success = rosidl_generator_c__String__assign(&msg->%s[%d], "%s");' % (field.name, i, field.default_value[i]))
+                        lines.append('  if (!success) {')
+                        lines.append('    goto %s%s;' % (label_prefix, last_label_index))
+                        abort_lines[0:0] = [
+                          '  rosidl_generator_c__String__fini(&msg->%s[%d]);' % (field.name, i),
+                          '%s%d:' % (label_prefix, last_label_index),
```
dirk-thomas: Indentation not a multiple of 4.

rosidl_generator_c/resource/msg__functions.c.em
```diff
@@ -134,7 +149,20 @@ for field in spec.fields:
...
             for i, default_value in enumerate(field.default_value):
-                lines.append('msg->%s.data[%d] = %s;' % (field.name, i, primitive_value_to_c(field.type.type, field.default_value[i])))
+                if field.type.type == 'string':
+                    lines.append('{')
+                    lines.append('  bool success = rosidl_generator_c__String__assign(&msg->%s.data[%d], "%s");' % (field.name, i, field.default_value[i]))
+                    lines.append('  if (!success) {')
+                    lines.append('    goto %s%s;' % (label_prefix, last_label_index))
+                    abort_lines[0:0] = [
+                      '  rosidl_generator_c__String__fini(&msg->%s.data[%d]);' % (field.name, i),
+                      '%s%d:' % (label_prefix, last_label_index),
```
dirk-thomas: Indentation not a multiple of 4.

rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -445,8 +445,7 @@ def parse_value_string(type_, value_string):
     if type_.is_primitive_type() and not type_.is_array:
         return parse_primitive_value_string(type_, value_string)
 
-    # TODO(mikaelarguedas) change this condition once string escape function is implemented
-    if type_.is_primitive_type() and type_.is_array and type_.type != 'string':
+    if type_.is_primitive_type() and type_.is_array:
```
dirk-thomas: It looks like this just enables default values for string arrays but without implementing the required string escaping logic? Currently I simply splits on commas - I don't think this is sufficient since the comma can be within a string.

mikaelarguedas: Yes sorry this is still work in progress, the strings are not escaped yet. This is also the case for simple strings. e.g. `string def_string 'Hello w"orld!'` makes the message generation crash (at least the python generator).

For string arrays it gets even trickier because the design document says that:
`string:
  a string value which can optionally be quoted with either single quotes (') or double quotes (")`
We need to make the quoting mandatory to allow differentiation of array elements otherwise it'll be ambiguous

dirk-thomas: Why does the quoting need to mandatory? It should be possible to parse the following without a problem:

string[] names ['foo', "bar", 'foo"bar', 'bar\'baz', "foo'bar\"baz"]

mikaelarguedas: if we allow to not quote strings it is valid to declare:
`string name a, b`
this results in a string containing `"a, b"` and is totally valid.

Using the same logic for arrays it's valid to declare:
`string[] names [a, b, c, d]`
in that case we don't know if it's a single element containing `"a, b, c, d" or four element separated by commas or any other combination.
That's what I meant by making the quoting mandatory. Otherwise we'd need to define a delimiter that is not allowed in strings which doesn't make much sense.

dirk-thomas: Actually we do know what the default values should be in that case. Since the individual items are not quoted the default value contains an array with four elements. Otherwise they would need to be quoted.

wjwwood: Do we need to make it optional? It sounds easier to just require quoting and escaping of interior quotes. It also makes it easier for a human (or syntax highlighter) to determine what is a literal and what is syntax. I'd even say, require double quote for the outside quotes. This isn't Python, and just as many languages, if not more, would use single quotes only for single characters.

dirk-thomas: When considering that we also want to pass values from the command line at some point I don't think only allowing double quotes is convenient.

wjwwood: But this is only for the message definition files right? I don't see what the command line has to do with this.

dirk-thomas: We might want to write only one parser and not multiple for different use cases?

wjwwood: I guess, but I still feel that keeping the idl format as simple as possible will be best because we'll likely want to parse it in other languages at some point.

It's the difference in strings for JSON and strings for YAML. YAML is nice, but it's also complicated.

wjwwood: I don't think allowing single quotes is unnecessarily burdensome, so I'm fine with that, I was just trying to find ways to keep is simple. But not requiring the quotes seems unnecessary to me. For ROS 1, we use yaml syntax on the command line (which worked really well and allowed us to only specify parts of the message). So, if we use that again, I don't think we need a super flexible parser that can work here and on any arbitrary input from the CLI.

mikaelarguedas: I agree with @wjwwood that it would be simpler and clearer if we require users to quote and escape the interior quotes. Not only to reduce parser complexity but also  improve human readability.

I agree that it all comes down to how (what format) we expect people to pass parameters (both CLI and IDL/config files). Do we have a document or discussion somewhere so that we could iterate on and refer to before making a decision on the scope of this parser?

mikaelarguedas:
This should be updated according to https://github.com/ros2/design/pull/111

mikaelarguedas:
@mikaelarguedas look at this someday !

mikaelarguedas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4616/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1517/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3804/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4694/)

CI testing only rosidl_parser with the linter fixup:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4619/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4697/)

rosidl_generator_c/test/test_interfaces.c
```diff
@@ -295,6 +310,10 @@ int test_strings(void)
   EXPECT_EQ(true, res);
   EXPECT_EQ(0, strcmp(strings->empty_string.data, TEST_STRING));
   EXPECT_EQ(0, strcmp(strings->def_string.data, "Hello world!"));
+  EXPECT_EQ(0, strcmp(strings->def_string2.data, "Hello'world!"));
+  EXPECT_EQ(0, strcmp(strings->def_string3.data, "Hello\"world!"));
+  EXPECT_EQ(0, strcmp(strings->def_string4.data, "Hello\'world!"));
```
dirk-thomas: Based on the message definition:

> string def_string4 'Hello\'world!'

Why is there a backslash in the C++ string literal? Should it be just `"Hello'world!"`?

Same for [def_various_quotes.data[1].data](https://github.com/ros2/rosidl/pull/197/files#diff-be6a8e8a931ac60ef4902d04a32978b0R430).

mikaelarguedas: hmmm good question. I guess as the escaping is not necessary in C because we use double quotes it should get "unescaped" by the C generator when it gets it from the parser

dirk-thomas: The extra backslash should be removed then.

mikaelarguedas: Done in efe98ea02a0a9a49f4173ab5c80ca9b9fd5556a8

Issue #193, opened 2016-12-14
[rosidl_generator_py] Support packages with only services but no messages


dirk-thomas:
Currently the generated library can't be loaded in that case since a required symbol isn't available. See ros2/examples#154.

PR #198, opened 2017-01-03
[rosidl_generator_py] Fixes srv only generation

mikaelarguedas:
fixes https://github.com/ros2/rosidl/issues/193
Connects to https://github.com/ros2/rosidl/issues/193
Reason of the failure: the generated file with all python extension symbols was not added to the library when no `.msg` files were present
What this PR does:
 * generated the extension files in the package folder rather than the `msg` subfolder
 * store extension files' paths in a msg agnostic cmake variable and use it for custom command and shared library

CI:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/2173/)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1655/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/2175/)

### DIFF
--- a/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
@@ -269,7 +269,7 @@ endforeach()
 if(BUILD_TESTING AND rosidl_generate_interfaces_ADD_LINTER_TESTS)
   if(
     NOT _generated_msg_py_files STREQUAL "" OR
-    NOT _generated_msg_extension_files STREQUAL "" OR
+    NOT _generated_extension_files STREQUAL "" OR
     NOT _generated_msg_c_files STREQUAL "" OR
     NOT _generated_srv_py_files STREQUAL "" OR
     NOT _generated_srv_c_files STREQUAL ""
--- a/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
+++ b/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
@@ -56,15 +56,7 @@ foreach(_idl_file ${rosidl_generate_interfaces_IDL_FILES})
     list(APPEND _generated_msg_c_files
       "${_output_path}/${_parent_folder}/_${_module_name}_s.c"
     )
-    foreach(_typesupport_impl ${_typesupport_impls})
-      list_append_unique(_generated_extension_files "${_output_path}/_${PROJECT_NAME}_s.ep.${_typesupport_impl}.c")
-      list_append_unique(_generated_extension_${_typesupport_impl}_files "${_output_path}/_${PROJECT_NAME}_s.ep.${_typesupport_impl}.c")
-    endforeach()
   elseif(_parent_folder STREQUAL "srv")
-    foreach(_typesupport_impl ${_typesupport_impls})
-      list_append_unique(_generated_extension_files "${_output_path}/_${PROJECT_NAME}_s.ep.${_typesupport_impl}.c")
-      list_append_unique(_generated_extension_${_typesupport_impl}_files "${_output_path}/_${PROJECT_NAME}_s.ep.${_typesupport_impl}.c")
-    endforeach()
     if("_${_module_name}_s.c" MATCHES "(.*)__response(.*)" OR "_${_module_name}_s.c" MATCHES "(.*)__request(.*)")
       list(APPEND _generated_srv_c_files
         "${_output_path}/${_parent_folder}/_${_module_name}_s.c"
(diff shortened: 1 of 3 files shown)


rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
```diff
@@ -121,7 +121,8 @@ def generate_py(generator_arguments_file, typesupport_impls):
             }
             data.update(functions)
             generated_file = os.path.join(
-                args['output_dir'], 'msg', generated_filename % args['package_name'])
+                args['output_dir'], generated_filename % args['package_name'])
+            print('***generated file: %s' % generated_file)
```
dirk-thomas: You might want to remove this debug output.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -56,14 +56,15 @@ foreach(_idl_file ${rosidl_generate_interfaces_IDL_FILES})
...
       "${_output_path}/${_parent_folder}/_${_module_name}_s.c"
     )
-    list(APPEND _generated_msg_c_common_files
-      "${_output_path}/${_parent_folder}/_${_module_name}_s.c"
-    )
     foreach(_typesupport_impl ${_typesupport_impls})
-      list_append_unique(_generated_msg_c_files "${_output_path}/${_parent_folder}/_${PROJECT_NAME}_s.ep.${_typesupport_impl}.c")
-      list_append_unique(_generated_msg_c_ts_${_typesupport_impl}_files "${_output_path}/${_parent_folder}/_${PROJECT_NAME}_s.ep.${_typesupport_impl}.c")
+      list_append_unique(_generated_extension_files "${_output_path}/_${PROJECT_NAME}_s.ep.${_typesupport_impl}.c")
+      list_append_unique(_generated_extension_${_typesupport_impl}_files "${_output_path}/_${PROJECT_NAME}_s.ep.${_typesupport_impl}.c")
```
dirk-thomas: Instead of repeating the whole value twice (and again twice for services) the variable `_generated_extension_files` could be updated by appending `_generated_extension_${_typesupport_impl}_files` at the end of each `foreach` cycle.

You could even move this out of the loop iterating over the msg / srv files and only set it once if there are any msg or srv files. I think that would be even clearer since these variable only ever contain a single value per typesupport.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -97,6 +89,12 @@ if(NOT _generated_srv_py_files STREQUAL "")
   )
 endif()
 
+if(NOT _generated_msg_c_files STREQUAL "" OR NOT _generated_srv_c_files STREQUAL "")
+    foreach(_typesupport_impl ${_typesupport_impls})
+      list_append_unique(_generated_extension_${_typesupport_impl}_files "${_output_path}/_${PROJECT_NAME}_s.ep.${_typesupport_impl}.c")
+      list_append_unique(_generated_extension_files "${_generated_extension_${_typesupport_impl}_files}")
```
dirk-thomas: While using `list_append_unique` works I would suggest using `list(APPEND` instead since there are by design no duplicates ever.

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -96,6 +89,12 @@ if(NOT _generated_srv_py_files STREQUAL "")
...
```
dirk-thomas: While using `list_append_unique` works for both lines I would suggest using `list(APPEND` instead since there are by design no duplicates ever.

mikaelarguedas:
ok I think this is ready for review now
reran CI just in case:
http://ci.ros2.org/job/ci_linux/2198/
http://ci.ros2.org/job/ci_windows/2191/

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -270,8 +269,8 @@ endforeach()
 if(BUILD_TESTING AND rosidl_generate_interfaces_ADD_LINTER_TESTS)
   if(
     NOT _generated_msg_py_files STREQUAL "" OR
+    NOT _generated_msg_extension_files STREQUAL "" OR
```
dirk-thomas: Where is this variable coming from?

mikaelarguedas: bad search/replace apparently :confounded:

mikaelarguedas:
Yes tried locally using the previously failing service examples

PR #204, opened 2017-02-02
Fix string delimiter

mikaelarguedas:
rosidl_parser uses double quote as string delimiter. So we should escape double quotes instead of single quotes.
Addresses https://github.com/ros2/rosidl/pull/197#discussion_r94443669

Before:
```python
            'DEF_STRING_DELIMITER__DEFAULT': "Hello"world!",
            'DEF_STRING_DELIMITER2__DEFAULT': "Hello\'world!",
            'DEF_STRING_DELIMITER3__DEFAULT': "Hello\'world!",
            'DEF_STRING_DELIMITER4__DEFAULT': "Hello"world!",
```
After
```python
            'DEF_STRING_DELIMITER__DEFAULT': "Hello\"world!",
            'DEF_STRING_DELIMITER2__DEFAULT': "Hello'world!",
            'DEF_STRING_DELIMITER3__DEFAULT': "Hello'world!",
            'DEF_STRING_DELIMITER4__DEFAULT': "Hello\"world!",
```

### DIFF
--- a/rosidl_generator_py/msg/Strings.msg
+++ b/rosidl_generator_py/msg/Strings.msg
@@ -2,8 +2,6 @@ string empty_string
 string def_string "Hello world!"
 string def_string_delimiter 'Hello"world!'
 string def_string_delimiter2 "Hello'world!"
-string def_string_delimiter3 'Hello'world!'
-string def_string_delimiter4 "Hello"world!"
 string<=22 ub_string
 string<=22 ub_def_string "Upper bounded string."
 string<=5[3] ub_string_static_array_value
--- a/rosidl_generator_py/msg/Strings.msg
+++ b/rosidl_generator_py/msg/Strings.msg
@@ -1,5 +1,9 @@
 string empty_string
 string def_string "Hello world!"
+string def_string_delimiter 'Hello"world!'
+string def_string_delimiter2 "Hello'world!"
+string def_string_delimiter3 'Hello'world!'
+string def_string_delimiter4 "Hello"world!"
 string<=22 ub_string
 string<=22 ub_def_string "Upper bounded string."
 string<=5[3] ub_string_static_array_value
--- a/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
+++ b/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
@@ -207,7 +207,7 @@ def constant_value_to_py(type_, value):
 
 def escape_string(s):
     s = s.replace('\\', '\\\\')
-    s = s.replace("'", "\\'")
+    s = s.replace('"', '\\"')
     return s
 
 
--- a/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
+++ b/rosidl_generator_py/rosidl_generator_py/generate_py_impl.py
@@ -208,6 +208,7 @@ def constant_value_to_py(type_, value):
 def escape_string(s):
     s = s.replace('\\', '\\\\')
     s = s.replace('"', '\\"')
+    s = s.replace("'", "\\'")
     return s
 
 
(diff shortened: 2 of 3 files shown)


rosidl_generator_py/msg/Strings.msg
```diff
@@ -1,5 +1,9 @@
 string empty_string
 string def_string "Hello world!"
+string def_string_delimiter 'Hello"world!'
+string def_string_delimiter2 "Hello'world!"
+string def_string_delimiter3 'Hello'world!'
+string def_string_delimiter4 "Hello"world!"
```
dirk-thomas: The last two lines don't look like a valid quoted string to me. How can the quotes used on the left and right of the string appear unescaped in the middle?

mikaelarguedas: because `rosidl_parser` already escapes the inner quotes before feeding the strings to the generators. I thought it was the expected behavior (based on your comment [here](https://github.com/ros2/rosidl/pull/197#discussion_r94444096)).
To clarify we're talking strings and not string arrays here, the question of requiring users to escape inner quotes came up when I started to look into parsing string arrays default values

dirk-thomas: I updated the referenced comment to put the snippet in a code block since some of the backslashes weren't visible. From a generic parsing point of view it can't be `'foo'bar'` imo.

mikaelarguedas: ok it does make more sense now. I can remove these two tests and will ticket the expected behavior of the `rosidl_parser` elsewhere. I'll address the change of behavior of the parser along with implementing parsing for arrays if there is no objection.

I guess that the following (without outer quotes) should be invalid too ?
 ```
string def_string_delimiter3 Hello"world!
string def_string_delimiter4 Hello'world!
```
but 
```
string def_string_delimiter3 Hello\"world!
string def_string_delimiter4 Hello\'world!
```
should be valid.

Is that correct?

Comment (dirk-thomas) [inline rosidl_generator_py/msg/Strings.msg]: I would expect exactly the opposite. The first two lines look valid. Since the string is not quoted there is no need to escape the quote characters. The second two lines might be valid but I would expect the string value to actually contain the backslash.

**Comment by mikaelarguedas** [comment]
Reduced the scope of the test until we change parser to escape strings as expected, ready for review

File: rosidl_generator_py/rosidl_generator_py/generate_py_impl.py [hunk 2]
```diff
@@ -207,6 +207,7 @@ def constant_value_to_py(type_, value):
 
 def escape_string(s):
     s = s.replace('\\', '\\\\')
+    s = s.replace('"', '\\"')
```
dirk-thomas: This escape is not necessary when the template always quotes string with single quotes.

mikaelarguedas: fair enough, fixed in https://github.com/ros2/rosidl/pull/204/commits/1e7941a2808378e5f277f54838320db23cececde

rosidl_generator_py/test/test_interfaces.py
```diff
@@ -111,6 +111,8 @@ def test_default_values():
     assert 'Bye world' == a.def_string
     assert 'Hello world!' == Strings.DEF_STRING__DEFAULT
     assert 'Hello world!' == a.DEF_STRING__DEFAULT
+    assert "Hello\"world!" == a.DEF_STRING_DELIMITER__DEFAULT
```
dirk-thomas: Why use double quotes here?

Next line it would make sense to use double quotes to avoid the need to escape the single quote.

mikaelarguedas:
@dirk-thomas I'll merge this as is if it's fine by you. We can extend the test suite for all generators once we move forward regarding escaping rules

PR #205, opened 2017-02-04
raise on non escaped quotes in strings

mikaelarguedas:
connect to https://github.com/ros2/design/pull/110

### DIFF
--- a/rosidl_parser/rosidl_parser/__init__.py
+++ b/rosidl_parser/rosidl_parser/__init__.py
@@ -566,26 +566,17 @@ def parse_primitive_value_string(type_, value_string):
         for quote in ['"', "'"]:
             if value_string.startswith(quote) and value_string.endswith(quote):
                 value_string = value_string[1:-1]
-                tmp_value_string = value_string
-                clear_escaping = False
-                while True:
-                    index = tmp_value_string.find(quote)
-                    if index < 0:
-                        # no inner quote, proceed
-                        break
-                    elif tmp_value_string[index - 1] != '\\' or index == 0:
-                        # inner quote not escaped, raise an InvalidValue Exception
-                        raise InvalidValue(
-                            primitive_type,
-                            value_string,
-                            'string is not properly escaped, the character `' +
-                            tmp_value_string[index] + '` should be escaped')
-                    else:
-                        clear_escaping = True
-                    tmp_value_string = tmp_value_string[index + 1:]
-                if clear_escaping:
-                    # clears escaping, each language will handle it differently
+                if quote == '"':
+                    match = re.search(r'(?<!\\)"', value_string)
+                else:
+                    match = re.search(r"(?<!\\)'", value_string)
+                if match is None:
                     value_string = value_string.replace('\\' + quote, quote)
+                else:
+                    raise InvalidValue(
+                        primitive_type,
+                        value_string,
+                        'string inner quotes not properly escaped')
                 break
 
         # check that value is in valid range
--- a/rosidl_parser/rosidl_parser/__init__.py
+++ b/rosidl_parser/rosidl_parser/__init__.py
@@ -569,7 +569,7 @@ def parse_primitive_value_string(type_, value_string):
                 tmp_value_string = value_string
                 clear_escaping = False
                 index = 0
-                while index > -1:
+                while index != -1:
                     index = tmp_value_string.find(quote)
                     if index < 0:
                         # no inner quote, proceed
@@ -578,7 +578,7 @@ def parse_primitive_value_string(type_, value_string):
                         # inner quote not escaped, raise an InvalidValue Exception
                         raise InvalidValue(
                             primitive_type,
-                            type_,
+                            value_string,
                             'string is not properly escaped, the character `' +
                             tmp_value_string[index] + '` should be escaped')
                     else:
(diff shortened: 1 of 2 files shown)


rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -569,7 +569,7 @@ def parse_primitive_value_string(type_, value_string):
                 tmp_value_string = value_string
                 clear_escaping = False
                 index = 0
-                while index > -1:
+                while index != -1:
```
dirk-thomas: From reading the code I think this condition doesn't make any difference. It could be changed to `while True`. It is only relevant when the loop enters the first time. After that the `index` is always changed in the next line and then tested in the over next. After that `index` is not changed anymore so if the code doesn't hit the `break` two lines below it will always do another loop.

dirk-thomas:
The logic to unescape the string isn't complete. As soon as an escape character is being used (for the quote the string was started and ended with) that escape character might also be escaped itself. Therefore the unescape needs to take that into account. E.g. consider the following string `'foo\\'bar'`. The backslash before the single quote doesn't make the string valid since it is escaped itself.

You might want to look into existing functions in Python which might provide the necessary functionality rather then implementing it manually.

mikaelarguedas:
In that case we need to define what set of characters we expect users to escape in our IDL. This design and parser update only takes into account the escaping of the inner quotes. So `'foo\\'bar'` is invalid because the `'` is not escaped.
This PR does not target escaping of other special characters like `\` or `%`.

dirk-thomas:
As soon as you insert an escape character you need to take into account escaping of the escape character. Otherwise you can't express certain values anymore.

rosidl_parser/rosidl_parser/__init__.py
mikaelarguedas: good point, it's not necessary now that index is only set within the loop.

mikaelarguedas:
This implementation allows to define any string and enforces only the escaping of inner quotes. It matches the definition in the design document. I can't think of any value that can't be expressed with this parsing.

Arguably escaping only quotes is counter intuitive and not supporting any other escape sequence may be against how users expect the parser to behave. I think that a follow up design document defining the complete set of features we want IDL to support is necessary (multiline strings, complex types default values, etc). And based on that feature-set we should revisit the parser implementation and the message definition itself.

dirk-thomas:
I agree with conclusion to get this change in as an incremental change and tackle the more complicated escaping and support for e.g. newline in another round.

With the PR only checking for unescaped quotes and only replacing escaped quotes with a plain quote I think the patch should be simplified.

A simple regex (of course untested :wink:) can perform the check if there are unescaped quotes: `re.search(r'(^|[^\\])"', default_value)` If this one matches an exception can be raised. If not, a simple string replace can substitute escaped quotes with plain quotes.

rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -566,6 +566,17 @@ def parse_primitive_value_string(type_, value_string):
         for quote in ['"', "'"]:
             if value_string.startswith(quote) and value_string.endswith(quote):
                 value_string = value_string[1:-1]
+                if quote == '"':
+                    match = re.search(r'(?<!\\)"', value_string)
+                else:
+                    match = re.search(r"(?<!\\)'", value_string)
```
dirk-thomas: Instead of repeating the same search twice I would suggest to insert the used `quote` into the pattern using string formatting.

rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -566,6 +566,17 @@ def parse_primitive_value_string(type_, value_string):
...
+                if match is None:
+                    value_string = value_string.replace('\\' + quote, quote)
+                else:
+                    raise InvalidValue(
+                        primitive_type,
+                        value_string,
+                        'string inner quotes not properly escaped')
```
dirk-thomas: If you invert the condition (`if match is not None: raise ...`) you don't need an else block.

dirk-thomas:
The patch looks good to me. It should probably be covered by some tests.

mikaelarguedas:
tested added in https://github.com/ros2/rosidl/pull/205/commits/922406eae63418f6390d9a0f214d54d7dc384d02, ready for review

dirk-thomas:
It would be good to add a few tests covering the case of having valid default values which are quoted (single as well as double) to ensure their "inner" part is extracted correctly.

Other than that the existing tests lgtm.

mikaelarguedas:
you mean like [this](https://github.com/ros2/rosidl/pull/205/files#diff-044611e43a41ec3a6301321e9c031985R151) or do you mean test if a string using single quotes can extract escaped double quotes properly? (`'foo\"bar`)

dirk-thomas:
The referenced test doesn't test a default value which is wrapped in quotes, e.g. `parse_primitive_value_string(Type('string'), '"foo"')`

mikaelarguedas:
ok thanks for clarifying, test added

dirk-thomas:
The new test covers default values which are wrapped in both quotes. It would still be good to have tests for default values wrapped in just one layer of quotes.

mikaelarguedas:
Tests with only one layer of quotes were already there prior to this PR both [simple quotes](https://github.com/ros2/rosidl/pull/205/files#diff-044611e43a41ec3a6301321e9c031985L115) and [double quotes](https://github.com/ros2/rosidl/pull/205/files#diff-044611e43a41ec3a6301321e9c031985L111)

dirk-thomas:
Aha, I see, I missed those - only focused on the diff. Well, now we have even tests for two layers of quoting :wink:

PR #210, opened 2017-03-30
don't import for bounded array

mikaelarguedas:
fix linter error reported in https://github.com/ros2/common_interfaces/pull/30#issuecomment-289942054


mikaelarguedas:
good point, problem of large screens... Fixed in https://github.com/ros2/rosidl/pull/210/commits/314694468a2e43eb6acbeacc62469c3eb08e7c90

Issue #190, opened 2016-12-13
Review primitive types


mikaelarguedas:
This is a meta-ticket for type deprecationg and primitive types supported is ROS 2 IDL.

- [ ] decide what types to support and update design documents accordingly
- [ ] update parsers and generators to match these types
- [ ] update code and example to use the new supported types

mikaelarguedas:
CI:
Linux: [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_linux&build=2047)](http://ci.ros2.org/job/ci_linux/2047)
OSX: [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_osx&build=1581)](http://ci.ros2.org/job/ci_osx/1581/)
Windows: [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_windows&build=2046)](http://ci.ros2.org/job/ci_windows/2046/)

dirk-thomas:
For these changes the existing design docs need to be considered too. There are specific mappings for DDS as well as Python which don't have an equivalent if these would be removed. Therefore I think this needs further consideration and shouldn't be done for the beta 1.

mikaelarguedas:
I'm not advocating to get this in the beta. It would be great but maybe a bit premature.
I'm opening this so that we will have this and the corresponding design PRs opened for reference and will be able to move forward on it as soon as we decide that this is the way to go.

mikaelarguedas:
I never really replied to https://github.com/ros2/rosidl/issues/190#issuecomment-266643123:
- deprecating this means that we don't send the DDS type "char" anymore
- current implementation just remaps to int8/uint8 types. Meaning that in python the corresponding message fields have to be populated as numbers and not as `builtin.bytes` or `buitin.str`. Extra logic can be added in the python getter to allow both types and store the integral equivalent under the hood. I think the getters will always return the integral representation though.

Note: the fact that ROS1 consider `bytes` as **signed** goes against Python's definition of a `byte` that is an **unsigned octet**, that's why I stuck to only remap the types in this PR

mikaelarguedas:
CI:
Linux: [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_linux&build=2492)](http://ci.ros2.org/job/ci_linux/2492)
OSX: [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_osx&build=1959)](http://ci.ros2.org/job/ci_osx/1959/)
[...]

PR #211, opened 2017-04-05
Map ROS char into C signed char

gerkey:
The C standard leaves it to the compiler implementor to decide whether a `char` is signed or unsigned:
* https://en.wikipedia.org/wiki/C_data_types#Basic_types

As a result, it's not portable to assign a negative value to a `char` and then test the result. This came up when building on aarch64:
* http://ci.ros2.org/job/ci_linux-aarch64/31/warnings22Result/
* http://ci.ros2.org/job/ci_linux-aarch64/31/testReport/(root)/projectroot/test_interfaces_c/

This PR changes the mapping of the ROS `char` to be a `signed char` in C. Requires ros2/rmw_fastrtps#99.

I'll also do a PR to update the design doc to specify the new mapping, but these code changes can be reviewed independently of that.

For a smaller test case, consider this program:
~~~
#include <stdio.h>

int main(void)
{
  char c1 = 66;
  char c2 = -66;

  if (c1 == 66)
    printf("positive pass\n");
  else
    printf("positive fail\n");

  if (c2 == -66)
    printf("negative pass\n");
  else
    printf("negative fail\n");
  return 0;
}
~~~

On linux/x86 it looks good:
~~~
$ gcc -Wtype-limits -o foo foo.c
$ ./foo
positive pass
negative pass
~~~

But on linux/aarch64 not so much:
~~~
$ gcc -Wall -Wtype-limits -o foo foo.c
foo.c: In function ‘main’:
foo.c:13:10: warning: comparison is always false due to limited range of data type [-Wtype-limits]
   if (c2 == -66)
          ^
$ ./foo
positive pass
negative fail
~~~


gerkey:
* http://ci.ros2.org/job/ci_linux/2395/
* http://ci.ros2.org/job/ci_osx/1849/
* http://ci.ros2.org/job/ci_windows/2514/

mikaelarguedas:
This looks to me like a good reason to revive the char/byte usefulness in our IDL format.

The [design doc](http://design.ros2.org/articles/interface_definition.html) defines:
```
char:
an integer value in the following interval [-128, 127]
```
I'm not convinced that changing the tests to stop testing the entire range of the char type is a long term solution. Maybe reviving https://github.com/ros2/rosidl/issues/190 and the associated Pull Requests will be more adequate. 

If we are looking for a short term solution I think that we should explicitly assign this to a `signed char` to comply with our spec and compiler on ARM.

mikaelarguedas:
Thanks for iterating on this. Can https://github.com/ros2/rosidl/pull/211/commits/c6e45ca25396a72331c4cc62be35f03c1e9d862b be reverted now that signed char is used ?

rosidl_generator_c/msg/PrimitiveValues.msg
```diff
@@ -1,7 +1,7 @@
 bool def_bool_1 true
 bool def_bool_2 false
 byte def_byte 66
-char def_char -66
+char def_char 67
```
dirk-thomas: This should be reverted.

Same below for test.

rosidl_generator_c/rosidl_generator_c/__init__.py
```diff
@@ -166,7 +166,7 @@ def primitive_value_to_c(type_, value):
     if type_ == 'bool':
         return 'true' if value else 'false'
 
-    if type_ in ['byte', 'char', 'int8', 'int16', 'int32', 'int64']:
+    if type_ in ['byte', 'char', 'signed char', 'int8', 'int16', 'int32', 'int64']:
```
dirk-thomas: `signed char` is not a valid type name. Valid type names are only the keys of the `MSG_TYPE_TO_C` dict.

rosidl_generator_c/msg/PrimitiveValues.msg
gerkey: 00901e50f806a9ddf72bf8afe103cb890cbfdb85

rosidl_generator_c/rosidl_generator_c/__init__.py
gerkey: 00901e50f806a9ddf72bf8afe103cb890cbfdb85

gerkey:
http://ci.ros2.org/job/ci_linux-aarch64/37/ (in queue)

gerkey:
http://ci.ros2.org/job/ci_linux-aarch64/39/

gerkey:
http://ci.ros2.org/job/ci_linux-aarch64/41/

gerkey:
Now that aarch64 is happy, here's full CI:

* http://ci.ros2.org/job/ci_linux/2406/
* http://ci.ros2.org/job/ci_linux-aarch64/42/
* http://ci.ros2.org/job/ci_osx/1855/
* http://ci.ros2.org/job/ci_windows/2523/

PR #217, opened 2017-04-29
enforce string length in byte rather than characters

mikaelarguedas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2567/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/164/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2037/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2676/)

related to ros2/ros2#384

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -211,7 +211,7 @@ bound = 2**nbits
 @[      end if]@
 @[  elif field.type.string_upper_bound]@
             ((isinstance(value, str) or isinstance(value, UserString)) and
-             len(value) <= @(field.type.string_upper_bound))
+             len(value.encode('utf-8')) <= @(field.type.string_upper_bound))
 @[  elif not field.type.is_primitive_type()]@
             isinstance(value, @(field.type.type))
 @[  elif field.type.type == 'byte']@
--- a/rosidl_generator_py/test/test_interfaces.py
+++ b/rosidl_generator_py/test/test_interfaces.py
@@ -134,6 +134,8 @@ def test_check_constraints():
     a.ub_string = 'a' * 22
     assert 'a' * 22 == a.ub_string
     assert_raises(AssertionError, setattr, a, 'ub_string', 'a' * 23)
+    assert len('こ' * 22) == 22
+    assert_raises(AssertionError, setattr, a, 'ub_string', 'こ' * 22)
 
     b = Nested()
     primitives = Primitives()


rosidl_generator_py/resource/_msg.py.em
```diff
@@ -211,7 +211,7 @@ bound = 2**nbits
 @[      end if]@
 @[  elif field.type.string_upper_bound]@
             ((isinstance(value, str) or isinstance(value, UserString)) and
-             len(value) <= @(field.type.string_upper_bound))
+             len(value.encode('utf-8')) <= @(field.type.string_upper_bound))
```
dirk-thomas: Since we currently don't specify the encoding in the design doc I don't this can assume `utf-8` here.

mikaelarguedas: I'm assuming an encoding that uses single byte. I don't think that the actual encoding matters much in this case, does it ? What approach would you suggest here ?

dirk-thomas: I am not sure atm. I think the design doc needs to be clear in terms of strings / wstrings before I know how this should be implemented.

mikaelarguedas:
moving this back in `ready` until a design doc specifying strings and char is updated

nuclearsandwich:
The package this pull request targets has been moved to https://github.com/ros2/rosidl_python.git

sloretz:
Closing since package has been moved

mikaelarguedas:
Can you please link to the corresponding PR opened on the new repository ?

(and delete this branch when you're done)

sloretz:
@mikaelarguedas I'm not sure that there needs to be a PR on the new repo. It seems like the result of [this discussion](https://github.com/ros2/rosidl/pull/217#discussion_r114379545) is we don't know how this should be implemented.  Would you like me to delete the branch anyways?

mikaelarguedas:
My recollection from the Bouncy release meeting was that we wanted to get this discussed/solved by Bouncy or shortly after hence the milestone.

If the decision is to not merge the change as is, I think we would be better off with a follow-up ticket for this to be visible/discussed/addressed rather than by closing this PR without follow-up ticket and leaving the branch around without any reference to it.

sloretz:
@mikaelarguedas I'm not sure what would be into a follow-up ticket. If I understand correctly this PR is a possible solution to ros2/ros2#384 "enforce string length in IDL in all languages" which looks like a ticket to implement ros2/design#117. Are you saying this PR is addressing another issue discussed in the bouncy release meeting?

mikaelarguedas:
oh I see your point, as this PR was linked from the higher level ticket I guess that's fine to just close this and delete the branch :+1:

PR #218, opened 2017-06-06
Add empy dependency.

nuclearsandwich:
Without this dependency use of this package was causing downstream failures.

Here's [a sample failure](http://test.build.ros2.org/job/Rbin_uX64__rosidl_generator_c__ubuntu_xenial_amd64__binary/5/console#console-section-5). Grep for ImportError.

```
16:05:01 Traceback (most recent call last):
16:05:01   File "/tmp/binarydeb/ros-r2b2-rosidl-generator-c-0.0.0/bin/rosidl_generator_c", line 20, in <module>
16:05:01     rosidl_generator_c = loader.load_module()
16:05:01   File "<frozen importlib._bootstrap_external>", line 388, in _check_name_wrapper
16:05:01   File "<frozen importlib._bootstrap_external>", line 809, in load_module
16:05:01   File "<frozen importlib._bootstrap_external>", line 668, in load_module
16:05:01   File "<frozen importlib._bootstrap>", line 265, in _load_module_shim
16:05:01   File "<frozen importlib._bootstrap>", line 626, in _exec
16:05:01   File "<frozen importlib._bootstrap_external>", line 665, in exec_module
16:05:01   File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
16:05:01   File "/tmp/binarydeb/ros-r2b2-rosidl-generator-c-0.0.0/rosidl_generator_c/__init__.py", line 17, in <module>
16:05:01     from rosidl_cmake import convert_camel_case_to_lower_case_underscore
16:05:01   File "/opt/ros/r2b2/lib/python3.5/site-packages/rosidl_cmake/__init__.py", line 21, in <module>
16:05:01     import em
16:05:01 ImportError: No module named 'em'
16:05:01 make[4]: *** [rosidl_generator_c/rosidl_generator_c/msg/empty.h] Error 1
```

### DIFF
--- a/python_cmake_module/package.xml
+++ b/python_cmake_module/package.xml
@@ -10,7 +10,7 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
-  <buildtool_export_depend>python3</buildtool_export_depend>
+  <buildtool_export_depend>python3-dev</buildtool_export_depend>
 
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
--- a/python_cmake_module/package.xml
+++ b/python_cmake_module/package.xml
@@ -10,6 +10,8 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
+  <buildtool_export_depend>python3</buildtool_export_depend>
+
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
 
--- a/python_cmake_module/package.xml
+++ b/python_cmake_module/package.xml
@@ -10,7 +10,7 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
-  <buildtool_export_depend>python3-dev</buildtool_export_depend>
+  <buildtool_export_depend>python3</buildtool_export_depend>
 
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
--- a/python_cmake_module/package.xml
+++ b/python_cmake_module/package.xml
@@ -10,8 +10,6 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
-  <buildtool_export_depend>python3</buildtool_export_depend>
-
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
 
--- a/rosidl_cmake/package.xml
+++ b/rosidl_cmake/package.xml
@@ -12,6 +12,8 @@
 
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
 
+  <depend>python3-empy</depend>
+
   <exec_depend>rosidl_parser</exec_depend>
 
   <test_depend>ament_lint_auto</test_depend>
--- a/rosidl_cmake/package.xml
+++ b/rosidl_cmake/package.xml
@@ -11,8 +11,8 @@
   <buildtool_depend>ament_cmake_python</buildtool_depend>
 
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
+  <buildtool_export_depend>python3-empy</buildtool_export_depend>
 
-  <exec_depend>python3-empy</exec_depend>
   <exec_depend>rosidl_parser</exec_depend>
 
   <test_depend>ament_lint_auto</test_depend>
--- a/rosidl_cmake/package.xml
+++ b/rosidl_cmake/package.xml
@@ -12,8 +12,7 @@
 
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
 
-  <depend>python3-empy</depend>
-
+  <exec_depend>python3-empy</exec_depend>
   <exec_depend>rosidl_parser</exec_depend>
 
   <test_depend>ament_lint_auto</test_depend>


rosidl_cmake/package.xml
```diff
@@ -12,6 +12,8 @@
 
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
 
+  <depend>python3-empy</depend>
```
dirk-thomas: What is Python EmPy needed for at build time in this package?

nuclearsandwich: It likely doesn't I was being overly broad. I've relaxed it to an exec_depend.

rosidl_cmake/package.xml
```diff
@@ -12,6 +12,7 @@
 
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
 
+  <exec_depend>python3-empy</exec_depend>
```
dirk-thomas: Should this be a `buildtool_export_depend` instead?

nuclearsandwich: Should it? I'm not familiar with the tool itself. I just want it to produce a useful package and know it needs empy to do it.

The semantics of the different dependency keys, especially for packages that are themselves build-time dependencies are not entirely clear to me. Maintainer editing is enabled and if you think this needs to change please feel free to make the change.

dirk-thomas: Please see [REP 140](http://www.ros.org/reps/rep-0140.html) which describes the different dependency types. I think `buildtool_export_depend` makes more sense since in a cross-compile scenario `em` must exist for the host platform (not the target platform) since it is being invoked as part of building downstream packages.

python_cmake_module/package.xml
```diff
@@ -10,7 +10,7 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
-  <buildtool_export_depend>python3-dev</buildtool_export_depend>
+  <buildtool_export_depend>python3</buildtool_export_depend>
```
dirk-thomas: Same question as on other PR for the need of a Python 3 dependency.

nuclearsandwich: https://github.com/ament/ament_lint/pull/76#discussion_r122558187

nuclearsandwich: @dirk-thomas this change is somewhat different from the others. The dependency key for python3-dev was already present from https://github.com/ros2/rosidl/commit/1d7b4eb15a0fae579ec4e02f1345156886235200. All I did was normalize the keyname to match the rosdep key rather than the ubuntu package name.

If you don't think it should be there at all, I think that is a separate change to be discussed. Build dependencies for ament_python packages will be added as needed but this is an ament_cmake package and if it needs python at build time, or needs to export a dependency on python at build time, this must remain.

dirk-thomas: I don't know why this package would need `python3-dev` but the package `rosidl_generator_py` in this repo is building a Python extension and therefore needs to have that line: `<buildtool_export_depend>python3</buildtool_export_depend>`

nuclearsandwich:
CI for empy key and removed python3-dev key.

[Linux Build Status](http://ci.ros2.org/job/ci_linux/2786/)
[MacOS Build Status](http://ci.ros2.org/job/ci_macos/2258/)
[Windows Build Status](http://ci.ros2.org/job/ci_windows/2924/)

dirk-thomas:
I am pretty sure my comment about the missing dependency in `rosidl_generator_py` still applies: https://github.com/ros2/rosidl/pull/218#discussion_r123373143

nuclearsandwich:
> I am pretty sure my comment about the missing dependency in rosidl_generator_py still applies: #218 (comment)

[On the buildfarm](http://build.ros2.org/job/Rbin_uX64__rosidl_generator_py__ubuntu_xenial_amd64__binary/5/consoleFull), python3-dev is being pulled in as a dependency of python_cmake_module.

Looking at the existing dependency set, it looks like [rclpy depends on python_cmake_module](https://github.com/ros2/rclpy/blob/151594c5e23f824a0da6f2fda9b4755710094df5/rclpy/package.xml#L11) and not python3-dev (either directly or via python3 rosdep key) as well.

With those two pieces of information, and the fact that this PR prior to these comments was producing packages, I am actually going to restore the dependency on python3-dev (via rosdep key) in python_cmake_module, make no changes to other packages in this repository, and request approval as-is after CI.

@dirk-thomas if you believe the dependencies, which appear to be functioning as desired, are incorrect, can you please either demonstrate this with a case where packages fail to perform or address it separately after the beta.

dirk-thomas:
> On the buildfarm, python3-dev is being pulled in as a dependency of python_cmake_module.

This patch is currently removing `<buildtool_export_depend>python3-dev</buildtool_export_depend>` form the manifest of the `python_cmake_module` package. As you pointed out yourself other packages might need that dependency. So I don't see why it should be removed.

> if you believe the dependencies, which appear to be functioning as desired, are incorrect, can you please either demonstrate this with a case where packages fail to perform or address it separately after the beta.

I would argue that you should come up with a reason why you want to remove that existing dependency and why it isn't necessary.

nuclearsandwich:
> This patch is currently removing <buildtool_export_depend>python3-dev</buildtool_export_depend> form the manifest of the python_cmake_module package. As you pointed out yourself other packages might need that dependency. So I don't see why it should be removed.

My intention is to restore that key for `python_cmake_module`, I'm trying to sort through bloom behavior on this repository and have not yet done so.

The separate change I meant was if you think it should be removed from `python_cmake_module` as [stated here](https://github.com/ros2/rosidl/pull/218#discussion_r123373143) and applied to `rosidl_generator_py` (stated in same comment).

dirk-thomas:
> I'm trying to sort through bloom behavior on this repository and have not yet done so.

How should bloom know that `python3-dev` is required?

> d289672a8aafe2e89f04d45125a332671a8d25ea

Why change `python3-dev` to `python3` here? The dev package is being required for specific reasons.

dirk-thomas:
I will approve this PR if the only change it does is adding the following dep to `rosidl_cmake`:

```
<buildtool_export_depend>python3-empy</buildtool_export_depend>
```

nuclearsandwich:
> Why change python3-dev to python3 here? The dev package is being required for specific reasons.

> How should bloom know that python3-dev is required?

It doesn't, and shouldn't. But bloom doesn't error out when python3-dev is used as a dependency key.


> Why change python3-dev to python3 here? The dev package is being required for specific reasons.

The [current rosdep key name](https://github.com/ros/rosdistro/blob/775248b613cd0f72cae4ac87d99e4d2a9c411364/rosdep/python.yaml#L3448-L3449) for the ubuntu package `python3-dev` is `python3`. That is both the reason for the original change and the reason I am looking at bloom now.

nuclearsandwich:
> I will approve this PR if the only change it does is adding the following dep to rosidl_cmake:

I can make that PR but it will be insufficient to bloom this package with the current rosdep keys.

dirk-thomas:
> I can make that PR but it will be insufficient to bloom this package with the current rosdep keys.

If you also add the missing rosdep key why would that state not work?

nuclearsandwich:
> If you also add the missing rosdep key why would that state not work?

Is it sensible for two rosdep keys to resolve to the same package? Or are you proposing that the existing rosdep key be changed?

dirk-thomas:
I wouldn't change the existing key since it is impossible to check where it is being used. Instead I would introduce a new one since the actual name would make it much clearer what it is being mapped to.

nuclearsandwich:
> I will approve this PR if the only change it does is adding the following dep to rosidl_cmake:
> 
> `<buildtool_export_depend>python3-empy</buildtool_export_depend>`

This is now the case. When https://github.com/ros/rosdistro/issues/15328 lands this can be merged with no additional changes. If the rosdep PR is still pending and the rest of the rosidl PRs land, I may merge this as-is anyway and patch the package.xml in release.

mikaelarguedas:
> I wouldn't change the existing key since it is impossible to check where it is being used.

I agree with the statement in general. In this particular case the key has been added a few days ago for our own use, every ROS distro targets python2 so I think that could be an exception. I think it may end up being confusing for users to have the `python3` rosdep key pointing to `python3-dev` while there is already a `python3-dev` key. So I'd be in favor of removing the `python3` key or modify it to the `python3` package

nuclearsandwich:
Since https://github.com/ros/rosdistro/issues/15328 has merged, this is ready. Because CI will not pull this dependency from the package.xml, is there a need to run CI for this change?

nuclearsandwich:
Wonderful. Thanks folks, particularly @dirk-thomas for working through this with me. 🎉

PR #215, opened 2017-04-20
adds *__destroy_ros_message function for generated message code

yamokosk:
See ros2/rclpy#79

PR #219, opened 2017-06-08
Fix destroy segfault

mikaelarguedas:
this is a follow up of https://github.com/ros2/rclpy/pull/87. It reverts #215 (by not clearing the allocated memory properly) makes the subscriber crash on message destruction for arrays of nested arrays. Leaving this in progress the time I add a regression test for it

connects to ros2/rclpy#87


mikaelarguedas:
Fixed in ros2/rosidl#223 ros2/rmw_fastrtps#126 and ros2/system_tests#211

Issue #216, opened 2017-04-26
Memory leak in nested messages


yamokosk:
This is really a continuation of issue ros2/rclpy#79. I am just continuing to go through top offenders reported by valgrind. I'm still using a simple pub/sub in Python and valgrinding the publisher. This is what I see:

```
==21797== 1,153 (1,152 direct, 1 indirect) bytes in 36 blocks are definitely lost in loss record 848 of 956
==21797==    at 0x4C2DB8F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==21797==    by 0xFBD4DC7: std_msgs__msg__MultiArrayDimension__create (multi_array_dimension__functions.c:43)
==21797==    by 0xF7B3457: std_msgs_multi_array_dimension__convert_from_py (_multi_array_dimension_s.c:15)
==21797==    by 0xF7B4423: std_msgs_multi_array_layout__convert_from_py (_multi_array_layout_s.c:50)
==21797==    by 0xF7ADF6F: std_msgs_float64_multi_array__convert_from_py (_float64_multi_array_s.c:37)
==21797==    by 0x9CF4B6A: rclpy_publish (_rclpy.c:282)
==21797==    by 0x4E9B9E: PyCFunction_Call (in /home/yamokosk/ros2_ws/ros2_env/bin/python3)
==21797==    by 0x524413: PyEval_EvalFrameEx (in /home/yamokosk/ros2_ws/ros2_env/bin/python3)
==21797==    by 0x528813: PyEval_EvalFrameEx (in /home/yamokosk/ros2_ws/ros2_env/bin/python3)
==21797==    by 0x528813: PyEval_EvalFrameEx (in /home/yamokosk/ros2_ws/ros2_env/bin/python3)
==21797==    by 0x52D82E: ??? (in /home/yamokosk/ros2_ws/ros2_env/bin/python3)
==21797==    by 0x529331: PyEval_EvalFrameEx (in /home/yamokosk/ros2_ws/ros2_env/bin/python3)

...

==21797== 1,408 (1,056 direct, 352 indirect) bytes in 33 blocks are definitely lost in loss record 866 of 956
==21797==    at 0x4C2DB8F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==21797==    by 0xFBD516F: std_msgs__msg__MultiArrayLayout__create (multi_array_layout__functions.c:41)
[...]

PR #220, opened 2017-06-16
Check return value of msg `__init` calls

allenh1:
Better cleanup for nested message structures.

There is still some leaked memory, but it's extremely minimal and this doesn't crash.

The leak that still exists comes from `std_msgs/String` and is likely the result of a race
condition, either in the middleware or in the subscriber code (because the publisher is not
affected). But, for now, this is a significant improvement.

See: #215, #216, ros2/rclpy#87

Connects to #216


mikaelarguedas:
Please open a PR on system test with reproducible example and regression test for this. I'll place this PR in review and look at it as soon as I have a minute

allenh1:
@mikaelarguedas got it.

rosidl_generator_c/src/string_functions.c
```diff
@@ -25,10 +25,8 @@ rosidl_generator_c__String__init(rosidl_generator_c__String * str)
   if (!str) {
     return false;
   }
-  str->data = malloc(1);
-  if (!str->data) {
-    return false;
-  }
+  assert(str->data = malloc(1));
```
mikaelarguedas: Why assert here ? all assert lines will be removed if you build in release mode so I don't think that's what you want to do here

allenh1: I figured that if malloc failed you should probably exit immediately. I didn't know that it would be removed if you compiled in release mode, however. I'll fix that.

allenh1: fixed

allenh1:
@Kukanani wrote some unit tests that we could get cleaned up and ready for this, I think that would be best.

mikaelarguedas:
Yes that would be great!

> open a PR on system test with reproducible example and regression test for this

I think that system_tests/test_communication is a good place to add the new custom message and fixtures to test this for regressions

dirk-thomas:
If it is only going to test the generated code of one generator it could probably also go into the generator package itself (likely a little bit more effort).

Kukanani:
I have a standalone test case (including a new message type) that exposes the underlying issue: https://github.com/ros2/system_tests/tree/fix-complex-message-cleanup

Unfortunately, it still crashes at the current state of this PR.

Kukanani:
The branch linked above now includes a C++ implementation as well. I tested all permutations of C++/Python pub/sub manually on my machine, and it is always the Python subscriber node that crashes.

With this PR in place, the GDB output is as follows:

```
Starting program: /usr/bin/python3 src/ros2/system_tests/test_communication/test/subscriber_py.py DynamicArrayPrimitivesNested
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff405c700 (LWP 18817)]
[New Thread 0x7ffff385b700 (LWP 18818)]
[New Thread 0x7ffff305a700 (LWP 18819)]
[New Thread 0x7ffff2859700 (LWP 18820)]
[New Thread 0x7ffff2058700 (LWP 18821)]
[New Thread 0x7ffff1857700 (LWP 18822)]
subscriber: beginning loop
python3: malloc.c:2392: sysmalloc: Assertion `(old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)' failed.

Thread 1 "python3" received signal SIGABRT, Aborted.
0x00007ffff7826428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
54	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
```

Backtrace:
```
#0  0x00007ffff7826428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff782802a in __GI_abort () at abort.c:89
#2  0x00007ffff786e2e8 in __malloc_assert (
    assertion=assertion@entry=0x7ffff7981598 "(old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)", file=file@entry=0x7ffff797dfe5 "malloc.c", line=line@entry=2392, 
    function=function@entry=0x7ffff7981de0 <__func__.11509> "sysmalloc") at malloc.c:301
#3  0x00007ffff7871b06 in sysmalloc (nb=nb@entry=800, av=av@entry=0x7ffff7bb4b20 <main_arena>) at malloc.c:2389
#4  0x00007ffff7872e4b in _int_malloc (av=av@entry=0x7ffff7bb4b20 <main_arena>, bytes=bytes@entry=792) at malloc.c:3825
#5  0x00007ffff78745d4 in __GI___libc_malloc (bytes=792) at malloc.c:2911
#6  0x00000000005d3eb6 in ?? ()
#7  0x00000000005d4248 in PyObject_Realloc ()
#8  0x0000000000554446 in _PyUnicodeWriter_PrepareInternal ()
#9  0x0000000000554b50 in _PyUnicodeWriter_WriteStr ()
#10 0x00000000005a01eb in ?? ()
#11 0x00000000005770a5 in PyObject_Repr ()
#12 0x0000000000528b26 in PyEval_EvalFrameEx ()
#13 0x000000000052d82f in ?? ()
#14 0x0000000000529332 in PyEval_EvalFrameEx ()
#15 0x000000000052d82f in ?? ()
#16 0x0000000000529332 in PyEval_EvalFrameEx ()
#17 0x000000000052e12b in PyEval_EvalCodeEx ()
#18 0x00000000004ebdd7 in ?? ()
#19 0x00000000005b7167 in PyObject_Call ()
#20 0x0000000000620bd8 in ?? ()
#21 0x00000000005b7167 in PyObject_Call ()
#22 0x0000000000528d06 in PyEval_EvalFrameEx ()
#23 0x000000000052d82f in ?? ()
#24 0x0000000000529332 in PyEval_EvalFrameEx ()
#25 0x000000000052d2e3 in ?? ()
#26 0x0000000000529332 in PyEval_EvalFrameEx ()
#27 0x000000000052d2e3 in ?? ()
#28 0x000000000052dfdf in PyEval_EvalCode ()
#29 0x00000000005fd2c2 in ?? ()
#30 0x00000000005ff76a in PyRun_FileExFlags ()
#31 0x00000000005ff95c in PyRun_SimpleFileExFlags ()
#32 0x000000000063e7d6 in Py_Main ()
#33 0x00000000004cfe41 in main ()
```

mikaelarguedas:
How does that compare with the current state of master ?
Does master fail in other scenarios that this one?
If so does it crash consistently at the same place?

dirk-thomas:
I think this needs to split up into two phases:

* Write some C tests which instantiate all message variations and tries to delete them again. This needs to pass with valgrind reporting no leaks.
* Then look at the Python extension which builds on top and do the same there.

mikaelarguedas:
> Write some C tests which instantiate all message variations and tries to delete them again. This needs to pass with valgrind reporting no leaks.
> Then look at the Python extension which builds on top and do the same there.

Agreed

allenh1:
@Kukanani Can I see the assertion that malloc throws for the reverted branch/master branch?

I don't think the reverted branch's leaking is the cause of the crash. I think the crash is actually due to a pointer being incorrectly set up in one of the `convert_to_py` functions, since the talker seems to be fine.

Kukanani:
I updated with the output from `master`, let me know if there's anything else you'd like me to test.

allenh1:
https://github.com/ros2/rosidl/blob/master/rosidl_generator_py/resource/_msg_support.c.em#L244

This might be related.

allenh1:
I'm installing the `python3-dbg` package so I can get some more info.

mikaelarguedas:
Just rebased this to work from a clean diff. I'll focus on the pure C part to see what's wrong before moving up to the python layer

rosidl_generator_c/resource/msg__functions.c.em
```diff
@@ -20,6 +20,7 @@ from rosidl_generator_c import value_to_c
 
 msg_typename = '%s__%s__%s' % (spec.base_type.pkg_name, subfolder, spec.base_type.type)
 array_typename = '%s__Array' % msg_typename
+cleanup_lines = []
```
dirk-thomas: Please move this closer to where it is actually being used. Likely under the definition of `abort_lines`.

rosidl_generator_c/src/string_functions.c
```diff
@@ -48,10 +48,8 @@ rosidl_generator_c__String__fini(rosidl_generator_c__String * str)
         "Exiting.\n");
       exit(-1);
     }
-    if (str->data) {
```
mikaelarguedas: Why do we need to remove this condition?

allenh1: Because we're in an if statement that just checked if `str->data` was `NULL`, so there's no sense in checking it again.

rosidl_generator_c/resource/msg__functions.c.em
```diff
@@ -98,11 +102,16 @@ for field in spec.fields:
...
-
         else:
             # initialize the sub message
-            lines.append('%s__%s__%s__init(&msg->%s);' % (field.type.pkg_name, 'msg', field.type.type, field.name))
-
+            lines.append('if (!%s__%s__%s__init(&msg->%s)) {' % (field.type.pkg_name, 'msg', field.type.type, field.name))
+            lines.append('  return false;')
+            lines.append('}')
+            cleanup_string = ('%s__%s__%s__destroy(&msg->%s);' % (field.type.pkg_name, 'msg', field.type.type, field.name))
+            if field.type.is_primitive_type():
```
dirk-thomas: Isn't this in the `else` block which already ensures that the field is not a primitive type? So this condition should never be true.

rosidl_generator_py/resource/_msg_support.c.em
```diff
@@ -22,10 +22,9 @@ nested_array_dict = {}
 @{have_not_included_string = False}@
 #include <rosidl_generator_c/string.h>
 #include <rosidl_generator_c/string_functions.h>
-
 @[  end if]@
 @{
-if not field.type.is_primitive_type() and field.type.is_array:
+if not field.type.is_primitive_type():
```
allenh1: We don't need to relax it anymore, I think. The original fix needed a bunch of array destroy functions and whatnot.

rosidl_generator_c/resource/msg__functions.c.em
```diff
@@ -81,7 +82,10 @@ for field in spec.fields:
         # non-array field
         if field.type.is_primitive_type():
             if field.type.type == 'string':
-                lines.append('rosidl_generator_c__String__init(&msg->%s);' % field.name)
+                lines.append('if (!rosidl_generator_c__String__init(&msg->%s)) %s' % (field.name, '{'))
+                lines.append('  return false;')
```
dirk-thomas: Isn't it necessary to execute the `cleanup_lines` in case of a `return` to clean up already initialized parts?

Same below.

rosidl_generator_c/resource/msg__functions.c.em
```diff
@@ -112,13 +121,21 @@ for field in spec.fields:
...
             # initialize each array element
             lines.append('for (size_t i = 0; i < %d; ++i) {' % field.type.array_size)
-            lines.append('  %s__init(&msg->%s[i]);' % (get_typename_of_base_type(field.type), field.name))
+            lines.append('  if(!%s__init(&msg->%s[i])) %s' % (get_typename_of_base_type(field.type), field.name, '{'))
+            lines.append('    return false;')
+            lines.append('  }')
             lines.append('}')
 
-    else:
+    else: # dynamic array
```
dirk-thomas: Two space before the `#`.

rosidl_generator_c/resource/msg__functions.c.em
```diff
@@ -112,13 +121,21 @@ for field in spec.fields:
         if not field.type.is_primitive_type() or field.type.type == 'string':
             # initialize each array element
             lines.append('for (size_t i = 0; i < %d; ++i) {' % field.type.array_size)
-            lines.append('  %s__init(&msg->%s[i]);' % (get_typename_of_base_type(field.type), field.name))
+            lines.append('  if(!%s__init(&msg->%s[i])) %s' % (get_typename_of_base_type(field.type), field.name, '{'))
+            lines.append('    return false;')
+            lines.append('  }')
```
dirk-thomas: Does this need corresponding cleanup lines for each item in the array?

rosidl_generator_c/src/string_functions.c
```diff
@@ -84,14 +82,15 @@ rosidl_generator_c__String__assignn(
...
     return false;
   }
-  char * data = (char *)malloc(n + 1);
-  if (!data) {
+  str->data = realloc(str->data, n + 1);
+  if (!str->data) {
+    return false;
+  }
+  str->data = strncpy(str->data, value, n);
+  if (!str->data) {
```
dirk-thomas: In the case it fails here is the `str->data` still guaranteed to be in a consistent state? If it was shrunk it might not be null terminated anymore.

allenh1: Yes, we force a null Terminator in the nth byte (from 0) near the end of that function.

rosidl_generator_c/resource/msg__functions.c.em
mikaelarguedas: :+1:

rosidl_generator_c/src/string_functions.c
mikaelarguedas: oh I didn't unfold the block :+1: sorry for the noise

rosidl_generator_c/src/string_functions.c
dirk-thomas: Which is being ignored when `return` is being invoked in the next line.

rosidl_generator_c/src/string_functions.c
```diff
@@ -101,13 +100,6 @@ bool
...
 {
-  if (!str) {
-    return false;
-  }
-  // a NULL value is not valid
-  if (!value) {
-    return false;
-  }
   return rosidl_generator_c__String__assignn(
     str, value, strlen(value));
```
mikaelarguedas: What happens if value is NULL ?

rosidl_generator_c/src/string_functions.c
allenh1: Right -- I see your point.

allenh1: Wait -- in that case, isn't it `NULL`?

rosidl_generator_c/src/string_functions.c
allenh1: That's checked in `rosidl_geneerator_c__String__assignn`, so it's redundant to check it again here.

rosidl_generator_c/resource/msg__functions.c.em
allenh1: No, since the array cleanup calls `fini`.

rosidl_generator_c/resource/msg__functions.c.em
allenh1: fixed.

rosidl_generator_c/resource/msg__functions.c.em
allenh1: Yes, yes it is. I'll add that in.

rosidl_generator_c/src/string_functions.c
```diff
@@ -84,14 +82,15 @@ rosidl_generator_c__String__assignn(
   if (n == SIZE_MAX) {
     return false;
   }
-  char * data = (char *)malloc(n + 1);
-  if (!data) {
+  str->data = realloc(str->data, n + 1);
```
clalancette: One thing to keep in mind with realloc(); if it fails, it does *not* free the original memory.  If this fails, I think what will happen here is that str->data will be NULL, and thus whatever old memory was in str->data will be leaked.  You may want to assign this to a temporary to avoid that situation.

rosidl_generator_py/resource/_msg_support.c.em
allenh1: Should we have a not in front of the second statement there?

```py
if not field.type.is_primitive_type() and not filed.type.is_array:
```

mikaelarguedas: Not sure what the reason was in the first place but it looks like that this is meant to be looking to nested arrays (c.f. variable name below `nested_array_dict`)

rosidl_generator_c/resource/msg__functions.c.em
```diff
@@ -112,13 +120,23 @@ for field in spec.fields:
...
-    else:
+    else:  # dynamic array
         if field.default_value is None:
             # initialize the dynamic array with a capacity of zero
-            lines.append('%s__Array__init(&msg->%s, 0);' % (get_typename_of_base_type(field.type), field.name))
+            lines.append('if (!%s__Array__init(&msg->%s, 0)) {' % (get_typename_of_base_type(field.type), field.name))
+            lines.append('  %s__destroy(msg);' % array_typename)
+            lines.append('  return false;')
+            lines.append('}')
+            cleanup_line = '%s__Array__destroy(&msg->%s);' % (get_typename_of_base_type(field.type), field.name)
```
mikaelarguedas: Not to sel + @allenh1 : this expects an array so generates warnings when we pass it a non Array message structure

mikaelarguedas:
putting this back in progress the time we're done iterating on it and have regression tests for it, will tag the team for review once it's done

allenh1:
After further investigation, I found the following:
```
Testing test_dynamic_array_nested messages...
Primitives__Array__destroy calls free(0x55ef6a0)
Primitives__Array__destroy calls free(0x55f0060)
DynamicArrayNested__destroy calls free(0x55f0060)
```

`Primitives__Array__destroy` should indeed be called twice, but, in the second call, it's freeing what it knows as `array`, but what is actually `msg` itself. Not sure how we would up with that scenario, but... It's pretty cool.

allenh1:
I see two routes as of now:
 * change the struct to have only pointer fields (this drops support for stack allocated messages, though)
 * Add an additional field into the struct to keep track of whether or not it has been freed (but that doesn't feel very safe to me).

Does anybody have any better ideas?

dirk-thomas:
> change the struct to have only pointer fields

I would say this is a certain no go. There is no reason to force that complexity on users.

> Add an additional field into the struct to keep track of whether or not it has been freed

Existing data structures in C don't need such a flag. As long as the code allocates and frees symmetrically everything is fine. And for extra safety after freeing a pointer it should be set to `null`. Therefore I don't think we should an additional field for this.

> Add another destroy function, msg_type__no_free_destroy (or something) that will do the normal destroy call, just doesn't call free on the passed in argument.

I don't have enough context about where these functions are currently being called and where the newly introduced one should be called. It would probably be best to describe those using an example to clarify the proposed change.

mikaelarguedas:
I don't think any of these two solutions is suitable. My feeling is that the current state of master (calling `destroy` on the highest level and calling `fini` on each individual submember) should be good enough given that each submember will clean itself recursively.

My feeling so far:
The changes of this PR have been making the original problem harder to investigate because we are now debugging these changes on top of the original problem and it's hard to identify the source of the crashes.

New approach:
1) I'm going to create a new branch extending the test suite of this package to expose the original problem. I'll provide the same set of messages + fixtures in test communication to identify any regression in the communication step.

2) I'll create a PR to address the problem once identified by the extended test suite.

3) @allenh1 Will use the branch of this PR to improve the current cleanup logic without introducing additional calls to destroy (aka free) and validate that it doesn't break anything. This PR will be reviewed + merged only after the original problem is identified and fixed

mikaelarguedas:
I reverted the new tests in https://github.com/ros2/rosidl/pull/220/commits/fa915c15d4bcc8153921d722e5db2108e249aff8, the new branch adding the tests can be found at https://github.com/ros2/rosidl/tree/extend_c_test_suite and I'm starting to investigate why some subnested arrays don't seem to be initialized resulting in segfaults

mikaelarguedas:
Update: tests added in a separate PR #222. They all pass even the nested++ messages so I don't think that the problem is in the message generation itself. Same tests added in test_communication on [that branch](https://github.com/ros2/system_tests/tree/c-memleak-testing) and they currently fail for fastrtps (but pass on connext) so I think the issue is in `rmw_fastrtps_cpp` with C typesupport

allenh1:
Ok, I've rebased and renamed this issue.

_New_ summary of changes:
-----------------------------------------
* Checked the return value of the init functions for message subfields
* Removed redundant checks in the string message code
* Fixed a compiler warning in the test code (comparison between signed and unsigned int)

All tests are passing on my end, and this should be ready for merge post beta 2 release.

rosidl_generator_c/resource/msg__functions.c.em
```diff
@@ -81,7 +81,10 @@ for field in spec.fields:
         # non-array field
         if field.type.is_primitive_type():
             if field.type.type == 'string':
-                lines.append('rosidl_generator_c__String__init(&msg->%s);' % field.name)
+                lines.append('if (!rosidl_generator_c__String__init(&msg->%s)) %s' % (field.name, '{'))
+                lines.append('    %s__destroy(msg);' % msg_typename)
+                lines.append('  return false;')
```
dirk-thomas: Returning here will not cleanup initialized / allocation memory of fields handles before.

allenh1: Doesn't destroy call fini, which would then call fini on all the subfields?

dirk-thomas: I haven't tried it. But e.g. in line 129 return is being called without destroy.

allenh1: Yup. I'll fix that.

dirk-thomas: Can you please add a test which ensures that partially constructed messages are cleaned up correctly.

allenh1: Yup!

allenh1: @dirk-thomas I can't think of a good way to do this. Any insight?

dirk-thomas: You could e.g. try to construct a partially initialized message manually and then call destroy on it to check that it is cleaned correctly.

allenh1:
@dirk-thomas I added a test in 8f718c4. Is this the kind of test to which you were referring?

rosidl_generator_c/src/string_functions.c
```diff
@@ -84,12 +82,15 @@ rosidl_generator_c__String__assignn(
   if (n == SIZE_MAX) {
     return false;
   }
-  char * data = (char *)malloc(n + 1);
+  char * data = realloc(str->data, n + 1);
   if (!data) {
+    free(str->data);
+    str->data = NULL;
+    str->size = 0;
+    str->capacity = 0;
```
dirk-thomas: This change will destroy the string if the function fails. The previous behavior would not change the existing string which I think is better behavior. What is the reason to change this?

rosidl_generator_py/resource/_msg_support.c.em
```diff
@@ -22,7 +22,6 @@ nested_array_dict = {}
 @{have_not_included_string = False}@
 #include <rosidl_generator_c/string.h>
 #include <rosidl_generator_c/string_functions.h>
-
```
dirk-thomas: Please keep this empty line. No need to change it in this PR.

rosidl_generator_c/test/test_invalid_initialization.c
```diff
@@ -0,0 +1,76 @@
...
+  if (test_init_null()) {
+    fprintf(stderr, "test_init_null FAILED\n");
+    rc++;
+  }
+  printf("Testing rosidl_generator_c destroy called on partial message...\n");
+  if (test_partial_fini_cleanup()) {
+    fprintf(stderr, "test_partial_fini_cleanup FAILED\n");
+    rc++;
+  }
+  return rc != 0;
```
dirk-thomas: This function is supposed to return an `int`, not a `bool`.

rosidl_generator_c/resource/msg__functions.c.em
```diff
@@ -81,7 +81,10 @@ for field in spec.fields:
         # non-array field
         if field.type.is_primitive_type():
             if field.type.type == 'string':
-                lines.append('rosidl_generator_c__String__init(&msg->%s);' % field.name)
+                lines.append('if (!rosidl_generator_c__String__init(&msg->%s)) %s' % (field.name, '{'))
```
mikaelarguedas: Nit: please place the `{` in the string itself rather that through a `%s` to stay consistent with what you do [line 106](https://github.com/ros2/rosidl/pull/220/files#diff-633804c447473336f244128adc0b96f3R106)

rosidl_generator_c/src/string_functions.c
allenh1: The reason I went for this was to protect againt the case where `realloc` fails. But I suppose that, should `realloc` fail, that doesn't mean that the existing string should be freed. So I'll change this to just `return false;` instead.

rosidl_generator_c/test/test_invalid_initialization.c
allenh1: Good catch!

allenh1:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2861/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/365/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2316/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3001/)

rosidl_generator_c/test/test_invalid_initialization.c
```diff
@@ -0,0 +1,76 @@
...
+#include <limits.h>
+#include <float.h>
+#include <stdint.h>
+#include <string.h>
+
+#include "rosidl_generator_c/string_functions.h"
+#include "rosidl_generator_c/msg/primitives.h"
+
+#define EXPECT_EQ(arg1, arg2) if (arg1 != arg2) return 1
+#define EXPECT_NE(arg1, arg2) if (arg1 == arg2) return 1
```
dirk-thomas: Please put `arg1` as well as `arg2 in parentheses (e.g. `if ((arg1) != (arg2)) return 1` since otherwise the logic might be wrong when passing non trivial arguments.

mikaelarguedas: @allenh1 If you address this comment, can you also propagate the change to the other test files of this package ? thanks!

allenh1: Sounds good to me. Do you want me to apply this to `test_inverfaces.c` as well?

allenh1: @mikaelarguedas Sorry, didn't see your comment.

allenh1:
New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2862/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/366/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2317/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3002/)

allenh1:
Windows building is unstable because of a small indentation issue. Fixed in last commit.

dirk-thomas:
It also has 45 warnings which need to be fixed.

allenh1:
@dirk-thomas Got it! I found the source of them, and think it's ready (but I'm going to thoroughly test it before I push).

allenh1:
Ok, this time should be it:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2864/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/368/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2319/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3004/)

allenh1:
Removed the `strncpy` call, as it generates a warning on Windows.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2865/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/369/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2320/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3005/)

rosidl_generator_c/resource/msg__functions.c.em
```diff
@@ -98,11 +101,12 @@ for field in spec.fields:
             elif field.default_value is not None:
                 # set default value of primitive type
                 lines.append('msg->%s = %s;' % (field.name, value_to_c(field.type, field.default_value)))
-
```
dirk-thomas: As mentioned before please avoid whitespace changes like this and restore the empty line.

allenh1: @dirk-thomas Sorry about that! Reverted.

mikaelarguedas:
Looks like this broke most of the communication tests involving c arrays

allenh1:
Hm... That's not good. I'll see if I can figure out what's going on, then ask for another review.

allenh1:
[crosses-fingers]:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2885/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/374/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2326/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3010/)

allenh1:
@mikaelarguedas / @dirk-thomas could I get one last review on this?

mikaelarguedas:
This intorduced linter failures in the generated C code. Fixed in https://github.com/ros2/rosidl/commit/4b809c184e315c8d2146964465c40d1746affee2

allenh1:
oops -- my b.

PR #221, opened 2017-06-17
Compile options

mikaelarguedas:
This is part of https://github.com/ros2/ros2/issues/362

I open a PR for this one because there is slightly more change than the other repos.

- I tried to get rid of the COMPILE_FLAGS property that is deprecated in cmake. There is one spot I left it (see inline comment) because that would require duplicating code.

Connects to ros2/ros2#362


rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -184,7 +184,7 @@ set_property(
 
 macro(set_properties _build_type)
   set_target_properties(${_target_name} PROPERTIES
-    COMPILE_FLAGS "${_extension_compile_flags}"
+    COMPILE_FLAGS ${_extension_compile_flags}
```
mikaelarguedas: Can't use `COMPILE_OPTIONS` here because the evaluation of `_extension_compile_flags` results in a string.
We could work around it by checking the compiler within this macro and make a different call to "set_target_properties" with COMPILE_DEFINITION only if the compiler supports the flags we want

dirk-thomas: Currently the variable contains a string. Have you tried to convert it into a list and then use `COMPILE_OPTIONS`?

mikaelarguedas: good point! https://github.com/ros2/rosidl/pull/221/commits/2865b6495ee6b28d48f647eb2e2a44b8573d776a

rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
```diff
@@ -212,9 +212,10 @@ foreach(_typesupport_impl ${_typesupport_impls})
 
   set(_extension_compile_flags "")
   set(_PYTHON_EXECUTABLE ${PYTHON_EXECUTABLE})
-  if(NOT WIN32)
-    set(_extension_compile_flags "-Wall -Wextra")
-  elseif("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+  if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    set(_extension_compile_flags "-Wall;-Wextra")
```
dirk-thomas: I would suggest to not use a semicolon here. Just pass both value parts as separate values and the variable will be a list (same as the `add_compile_options` invocation above).

mikaelarguedas: fair enough https://github.com/ros2/rosidl/pull/221/commits/30421427dc298d85b04d10214f61da4a7359c3e8

dirk-thomas:
While this looks good I am not sure CMake across multiple values per property (see https://cmake.org/cmake/help/v3.5/command/set_target_properties.html). Have you checked that all warning levels are being enabled as desired?

mikaelarguedas:
looks like it works as expected:
```cmake
  set(a "")
  get_target_property(a ${_target_name} COMPILE_OPTIONS)
  message(FATAL_ERROR "COMPILE_OPTIONS:${a}")
```
prints:
```
COMPILE_OPTIONS:-Wall;-Wextra
```

mikaelarguedas:
confirmed passing `-Warguedas` in the same line makes the comipler complain. I'm going to squash and merge this

PR #222, opened 2017-06-22
[rosidl_generator_c] Extend c test suite

mikaelarguedas:
This adds a bunch of nested messages and tests to make sure messages can be allocated/populated/destroyed without error.


mikaelarguedas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2798/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/319/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2266/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2934/)

mikaelarguedas:
@allenh1 FYI: you can test your changes using this test suite to confirm the cleanup logic (but please don't push on this branch since I expect this to be merged before any change to the message generation itself)

rosidl_generator_c/msg/DynamicArrayPrimitivesNested.msg
```diff
@@ -0,0 +1 @@
+DynamicArrayPrimitives[] msgs
```
allenh1: No newline at the end of this file

rosidl_generator_c/test/test_interfaces.c
```diff
@@ -896,3 +941,325 @@ int test_submessages(void)
 
   return 0;
```
allenh1: I know this isn't in the diff, but could we change this to `return rc;` (so that we return an error code only when tests fail).

rosidl_generator_c/msg/DynamicArrayPrimitivesNested.msg
mikaelarguedas: Given that it's not code I think we don't care. But feel free to commit a modified version with the newline if you feel like it

mikaelarguedas:
Note: The commented includes will be removed before merging

rosidl_generator_c/test/test_interfaces.c
mikaelarguedas: If think there was a good reason for not doing so. Can't think of it right now and can't test it today. It shouldn't block this PR though

clalancette: I don't think you need to change anything here.  The EXPECT_EQ macro already does a "return 1" if it fails; thus, we just need to return 0 at the end here in case everything passed.

rosidl_generator_c/test/test_interfaces.c
```diff
@@ -24,53 +24,74 @@
...
 #include "rosidl_generator_c/string_functions.h"
 
-#include "rosidl_generator_c/msg/various__struct.h"
+
 #include "rosidl_generator_c/msg/bool__struct.h"
+#include "rosidl_generator_c/msg/bounded_array_nested__struct.h"
+// #include "rosidl_generator_c/msg/bounded_array_primitives__struct.h"
 #include "rosidl_generator_c/msg/byte__struct.h"
 #include "rosidl_generator_c/msg/char__struct.h"
+#include "rosidl_generator_c/msg/constants__struct.h"
```
clalancette: Out of curiousity, is there any reason we prefer to `#include "rosidl_generator_c/msg/constants__struct.h"` (and the corresponding functions.h) rather than including constants.h, which will pull them both in?

rosidl_generator_c/msg/DynamicArrayPrimitivesNested.msg
dirk-thomas: "We" still do: 07b54e77544d95b4c3dcb1c9ff76a75e4111ba87

Otherwise e.g. simply calling `cat <filename>` results in not that nice output.

rosidl_generator_c/test/test_interfaces.c
dirk-thomas: `rc` is not even defined in this function. So I think `0` is the right return code.

rosidl_generator_c/msg/DynamicArrayPrimitivesNested.msg
mikaelarguedas: Thanks for pushing a fix

rosidl_generator_c/test/test_interfaces.c
mikaelarguedas: Good point I'll collapse the include statements while removing the commented ones

mikaelarguedas:
Thanks all for the first round of reviews, I think this is ready for another :eyes:

rosidl_generator_c/test/test_interfaces.c
```diff
@@ -896,3 +904,325 @@ int test_submessages(void)
...
   return 0;
 }
+
+/**
+ * Test message with bounded array nested types
+ */
+int test_bounded_array_nested(void)
+{
+  bool res;
+  size_t size = 4;
```
clalancette: This may be my ignorance of how this section of code works, but why are we setting the size to 4 and then only using the first two elements of the array?

clalancette: Oh yeah, excellent thing to test.  A comment would be helpful, thanks :).

mikaelarguedas: Done in https://github.com/ros2/rosidl/pull/222/commits/4f9033a3d1f78288dbbf27f7adad80e220850b8c thanks for pointing it out

mikaelarguedas:
@ros2/team Still waiting for :+1: or comments to address on this

wjwwood:
Might need new CI since there was additional changes since the last, but I'll leave that up to @mikaelarguedas.

mikaelarguedas:
Ci is still happy, merging

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2819/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/326/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2278/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2958/)

PR #223, opened 2017-06-24
add sizeof message struct to introspection information

dirk-thomas:
Ready for review.


allenh1:
So glad this was an easy fix (though not easy to find).

Should we look into adapting the other rmw implementations to use this new field?

mikaelarguedas:
> Should we look into adapting the other rmw implementations to use this new field?

As discussed offline there is no supported implementation using the introspection typesupport. We had an implementation of rmw_connext using introspection but it has diverged a lot featurewise since we stoped supporting it (~8months ago). I don't think that we should spend time/effort on not supported rmw implementations for now

mikaelarguedas:
We could ticket it though if we ever want to revive connext_dynamic that's good to have a paper trail of what should be done

PR #224, opened 2017-06-27
add messages to all assert calls in Python message classes

dirk-thomas:
The added message argument to the `assert` calls will help significantly when trying to pass wrong values (as in ros2/ros2cli#24).

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2831/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/335/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2287/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2967/)

Ready for review.

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -104,7 +104,7 @@ class @(spec.base_type.type)(metaclass=Metaclass):
 
     def __init__(self, **kwargs):
         assert all(['_' + key in self.__slots__ for key in kwargs.keys()]), \
-            "Invalid arguments passed to constructor: %r" % kwargs.keys()
+            'Invalid arguments passed to constructor: %r' % kwargs.keys()
 @[  for field in spec.fields]@
 @[    if field.default_value]@
         self.@(field.name) = kwargs.get(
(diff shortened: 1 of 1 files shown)


wjwwood:
Looks like there is an issue with the changes. I can't tell with just my phone what the problem is. I'll review the code to see if I can see the issue. But I wanted to make sure you knew.

rosidl_generator_py/resource/_msg.py.em
```diff
@@ -181,45 +181,58 @@ class @(spec.base_type.type)(metaclass=Metaclass):
...
 @[  elif field.type.type == 'byte']@
             ((isinstance(value, bytes) or isinstance(value, ByteString)) and
-             len(value) == 1)
+             len(value) == 1), \
+            "The '@(field.name)' field must of type 'bytes' or 'ByteString' with a length 1"
 @[  elif field.type.type == 'char']@
             ((isinstance(value, str) or isinstance(value, UserString)) and
-             len(value) == 1 and ord(value) >= -128 and ord(value) < 128)
+             len(value) == 1 and ord(value) >= -128 and ord(value) < 128), \
+            "The '@(field.name)' field must of type 'str' or 'UserString' with a length 1 and the character ord() in [-128, 127)"
```
mikaelarguedas: I know we don't check line length on generated code but for constant strings like that I think wrapping would improved readability

dirk-thomas: Wrapped in 61f0c9f4ce66e10b8bd4534cbad931742c5f3f4a

PR #228, opened 2017-08-07
Allocator traits

mikaelarguedas:
opening this for visibility, that can be a starting point if we want to provide allocator traits. Feel free to comment here if you think this is not of interest and should be deleted


dirk-thomas:
Closing due to age / inactivity.

PR #229, opened 2017-08-16
Add has_bounded_size trait (Redo of #30)

mikaelarguedas:
redo of https://github.com/ros2/rosidl/pull/30

Note: I put the logic to determine bounded size in another python loop for readability and sanity checking of the conditions. If we are not willing to take the performace hit of iterating twice I can merge them

### DIFF
--- a/rosidl_generator_cpp/resource/msg__traits.hpp.em
+++ b/rosidl_generator_cpp/resource/msg__traits.hpp.em
@@ -37,6 +37,9 @@ namespace rosidl_generator_traits
 template<typename T>
 struct has_fixed_size : std::false_type {};
 
+template<typename T>
+struct has_bounded_size : std::false_type {};
+
 #endif  // __ROSIDL_GENERATOR_CPP_TRAITS
 
 #include "@(spec.base_type.pkg_name)/@(subfolder)/@(get_header_filename_from_msg_name(spec.base_type.type))__struct.hpp"
@@ -65,6 +68,30 @@ else:
 template<>
 struct has_fixed_size<@(cpp_namespace)@(spec.base_type.type)>
   : std::integral_constant<bool, @(fixed_template_string)>{};
+
+@{
+bounded_template_strings = []
+bounded = True
+
+for field in spec.fields:
+    if field.type.type == 'string' and field.type.string_upper_bound in None:
+        bounded = False
+        break
+    elif field.type.is_dynamic_array() and not field.type.is_upper_bound:
+        bounded = False
+        break
+    if bounded and not field.type.is_primitive_type():
+        bounded_template_strings.append(
+            "has_bounded_size<{}::msg::{}>::value".format(field.type.pkg_name, field.type.type))
+
+if bounded:
+    bounded_template_string = ' && '.join(bounded_template_strings) if bounded_template_strings else 'true'
+else:
+    bounded_template_string = 'false'
+}@
+template<>
+struct has_bounded_size<@(cpp_namespace)@(spec.base_type.type)>
+  : std::integral_constant<bool, @(bounded_template_string)>{};
 
 }  // namespace rosidl_generator_traits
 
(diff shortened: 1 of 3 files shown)


rosidl_generator_cpp/resource/msg__traits.hpp.em
```diff
@@ -67,6 +71,32 @@ template<>
...
+bounded = True
+
+for field in spec.fields:
+    if field.type.type == 'string' and field.type.string_upper_bound is None:
+        bounded = False
+        break
+    elif field.type.is_dynamic_array() and not field.type.is_upper_bound:
+        bounded = False
+        break
+    if bounded and not field.type.is_primitive_type():
```
dirk-thomas: `bounded` is always `True` here.

rosidl_generator_cpp/resource/msg__traits.hpp.em
```diff
@@ -67,6 +71,32 @@ template<>
...
 
+@{
+bounded_template_strings = []
+bounded = True
+
+for field in spec.fields:
+    if field.type.type == 'string' and field.type.string_upper_bound is None:
+        bounded = False
+        break
+    elif field.type.is_dynamic_array() and not field.type.is_upper_bound:
```
dirk-thomas: `if` instead of `elif` since the previous block `break`s anyway.

rosidl_generator_cpp/resource/msg__traits.hpp.em
```diff
@@ -67,6 +71,32 @@ template<>
 struct has_fixed_size<@(cpp_namespace)@(spec.base_type.type)>
   : std::integral_constant<bool, @(fixed_template_string)>{};
 
+@{
+bounded_template_strings = []
+bounded = True
```
dirk-thomas: Instead of setting the flag within the loop this could be written like this:

```
bounded = False:
for field in spec.fields:
    if ...:
        break
    if ...:
        break
    if field.type.is_primitive_type():
        # append to bounded_template_strings
```

That might make it more obvious when the variable is `True` and `False` without having to check where it is assigned within the loop.

[The same could be applied to the existing loop.]

mikaelarguedas: That's a fair point. If the initial value is false, the last (non-breaking) condition needs to set it to True. I think that the first 2 conditions still need to set it to False regardless of the initialization of the variable. I can explicitely set it to True in the other blocks if it improves readability. It would result in something like:
```
fixed = False
if len(spec.fields) == 0:
    fixed = True
else:
    for field in spec.fields:
        if field.type.type == 'string':
            fixed = False
            break
        if field.type.is_dynamic_array():
            fixed = False
            break
        if not field.type.is_primitive_type():
            fixed = True
            tmp_fixed_string = "has_fixed_size<{}::msg::{}>::value".format(
                field.type.pkg_name, field.type.type)
            if tmp_fixed_string not in fixed_template_strings:
                fixed_template_strings.append(tmp_fixed_string)
        else:
            fixed = True
```
Is that preferred ?

rosidl_generator_cpp/resource/msg__traits.hpp.em
mikaelarguedas: good point will fix

rosidl_generator_cpp/resource/msg__traits.hpp.em
mikaelarguedas: good point will fix

rosidl_generator_cpp/resource/msg__traits.hpp.em
dirk-thomas: Sorry, my example was lacking an important part. I updated the previous comment and added the missing `else` to the `for` loop.

mikaelarguedas:
CI just to make sure it builds and passes the tests on all platforms

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3033/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/438/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2439/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3111/)

PR #234, opened 2017-08-30
[rosidl_generator_py] display inclusive boundaries and use closed brackets accordingly

mikaelarguedas:
noticed that when reviewing a PR in this repo. I think it's clearer this way but am open to suggestions if it's not the case for everyone


mikaelarguedas:
I remember I had to lookup online the notation in english when I added the error message in the first place because not all languages use the same.  Given that I needed to fix some boundaries anyway (https://github.com/ros2/rosidl/pull/234/files#diff-9834b895f70443ae69a70ae9da88290eL250) I figured I would use something that seemed more internationally "clear"

PR #236, opened 2017-09-06
Zero initialization

clalancette:
This PR implements the different forms of initialization that we want to allow for the instantiation of messages, as discussed in ros2/ros2#396 and ros2/design#136 .  The default is to do full initialization of primitive message fields with the default values (where those exist) or zeros (for everything else).  Complex types (such as `std::string`, `std::array`, etc) already have their constructors called, so there is no need to do anything specific there.

connects to ros2/ros2#396

CI is also in the above linked issue.


rosidl_generator_cpp/test/test_initialization.cpp
```diff
@@ -0,0 +1,57 @@
...
+}
+
+TEST(Test_initialization, defaults_only_constructor) {
+  rosidl_generator_cpp::msg::PrimitivesSomeDefault def(
+    rosidl_generator_cpp::MessageInitialization::DEFAULTS_ONLY);
+
+  ASSERT_EQ(1.125f, def.float32_value);
+  ASSERT_EQ(2.4, def.float64_value);
+  // skip checking def.uint64_value because it doesn't have a default in this
+  // message and thus isn't initialized with DEFAULTS_ONLY
```
wjwwood: It might be interesting to seed the memory with a known, non-zero, pattern and then do a placement new. Then you could check to make sure `uint64_value` is not `0`. Something like:

```c++
char * memory = new char[sizeof(rosidl_generator_cpp::msg::PrimitivesSomeDefault)];
ASSERT_NE(memory, nullptr);
std::memset(memory, 0xfe, sizeof(rosidl_generator_cpp::msg::PrimitivesSomeDefault));
rosidl_generator_cpp::msg::PrimitivesSomeDefault * def =
  new(memory) (rosidl_generator_cpp::MessageInitialization::DEFAULTS_ONLY);

ASSERT_NE(0ULL, def->uint64_value);

// This should be in a "scope exit" to prevent a memory leak during an assert raising...
def->~PrimitivesSomeDefault();
delete[] memory;
```

This might be fragile, because I don't know what the C++ standard says about this, it might let a compiler initialize it to zero if it wanted to do so during the placement new. But if you wanted to try and test this case, this would be how to do it I think.

rosidl_generator_cpp/package.xml
```diff
@@ -15,6 +15,8 @@
   <!-- This is needed for the rosidl_message_type_support_t struct and visibility macros -->
   <build_export_depend>rosidl_generator_c</build_export_depend>
 
+  <build_depend>rosidl_generator_c</build_depend>
```
dirk-thomas: Please move the build dependency above the build export dependencies (around line 11).

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -81,55 +82,58 @@ struct @(spec.base_type.type)_
...
+    print('    if (init == rosidl_generator_cpp::MessageInitialization::ALL) {')
+    print('      this->initialize_all();')
+    print('    } else if (init == rosidl_generator_cpp::MessageInitialization::SKIP) {')
+    print('      // no work to do')
+    print('    } else if (init == rosidl_generator_cpp::MessageInitialization::ZERO) {')
+    print('      this->zero_initialize();')
+    print('    } else if (init == rosidl_generator_cpp::MessageInitialization::DEFAULTS_ONLY) {')
+    print('      this->initialize_defaults_only();')
+    print('    }')
+    print('  }')
```
dirk-thomas: Can this be refactored to actually use the fact that this is a template? The extra indirection of using `print` here make the code even more difficult to read.

rosidl_generator_cpp/package.xml
clalancette: Will do.

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -81,55 +82,58 @@ struct @(spec.base_type.type)_
...
-            field.type.is_primitive_type() and field.default_value is not None:
-        cpp_value = value_to_cpp(field.type, field.default_value)
-        print('    %s = %s;' % (field.name, cpp_value))
+def init_list(alloc_name):
+    retstring = ":"
+    used_alloc = False
+    for field in spec.fields:
+        # dynamic arrays require allocator if available
+        if field.type.is_array and \
+                (field.type.array_size is None or field.type.is_upper_bound):
```
mikaelarguedas: Tip: `field.type.is_dynamic_array()` (I know this code was just copied but I think we'll gain in readability)

rosidl_generator_cpp/resource/msg__struct.hpp.em
clalancette: I went back and forth on this; it wasn't clear to me that using it as a template actually made it easier to read in all cases.  After looking at the implementation again, I agree with you that in the case of `gen_constructor`, it makes more sense to use it as a template and not as a function with a bunch of `print` statements.  I'll change that.  For `gen_init_func` (below), I don't believe that this is the case; it is mostly control structures, with just a handful of prints embedded, and the templated version is going to be very difficult to read.  For now, I'm just going to change `gen_constructor`, and let's go from there.

rosidl_generator_cpp/resource/msg__struct.hpp.em
clalancette: Will do.

rosidl_generator_cpp/resource/msg__struct.hpp.em
dirk-thomas: > For `gen_init_func` (below), I don't believe that this is the case; it is mostly control structures, with just a handful of prints embedded, and the templated version is going to be very difficult to read.

Then I would recommend to keep the control structure in a function but move the prints out of the function and do them in em. I am pretty sure the result is more readable.

rosidl_generator_cpp/test/test_initialization.cpp
clalancette: That's actually a really good idea, thanks @wjwwood I've gone ahead and implemented this as you pointed out, and it seems to work as advertised, at least on Linux.  I'll give it another CI job run on the other platforms to make sure that it works everywhere.

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -81,55 +82,48 @@ struct @(spec.base_type.type)_
...
-    # if default value is available set it in initializer list
-    elif field.type.is_primitive_type() and field.default_value is not None:
-        cpp_value = value_to_cpp(field.type, field.default_value)
-        if leading_colon == ' ':
-            print(',')
-        print('  %s %s(%s)' % (leading_colon, field.name, cpp_value), end='')
-        leading_colon = ' '
-if leading_colon == ' ':
-    print()
+        init_list += ' %s(_alloc),' % (field.name)
```
dirk-thomas: Instead of building a string incrementally I would suggest to collect the initializations here:

```
init_list = []
for ...:
   if ...:
        init_list.append(field.name + '(_alloc)')
```

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -81,55 +82,48 @@ struct @(spec.base_type.type)_
...
+        used_alloc = True
+
+# Always remove the last element; it will either be a trailing comma we do
+# not want, or a colon we do not want.
+init_list = init_list[:-1]
 }@
-@[  if alloc_name == '_alloc']@
-// *INDENT-ON*
-@[  end if]@
+  explicit @(spec.base_type.type)_(const ContainerAllocator & _alloc, rosidl_generator_cpp::MessageInitialization init = rosidl_generator_cpp::MessageInitialization::ALL)@(init_list)
```
dirk-thomas: This line will get pretty long. Together with the comment above this can be written as:

```
explicit ... MessageInitialization::ALL)
@[if init_list]@
    : @(', '.join(init_list))
@[end if]@
```

If the length of that line still bothers us we could even put each initialization on a separate line.

rosidl_generator_cpp/resource/msg__struct.hpp.em
dirk-thomas: Doesn't `init` need to be passed to the constructor calls of each field?

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -148,6 +142,61 @@ for field in spec.fields:
...
+                elif use_zeros:
+                    if field.type.type == 'string':
+                        cpp_value = '""';
+                    else:
+                        cpp_value = '0'
+                if cpp_value is not None:
+                    init_strs.append('this->%s = %s;' % (field.name, cpp_value))
+            # Arrays are represented either by a std::vector or by a BoundedVector.
+            # In both cases they are not primitive types, so their constructors
+            # will automatically be run and we don't have to do anything.
```
dirk-thomas: An field which is an array of primitives could still have a default value which needs to be initialized based on the `MessageInitialization` argument.

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -148,6 +142,61 @@ for field in spec.fields:
...
+        else:
+            if not field.type.is_array:
+                init_strs.append('this->%s.%s();' % (field.name, funcname))
```
dirk-thomas: The default constructor of complex fields is automatically invoked. It might need to pass the `MessageInitialization` argument though (probably in the initializer list above rather than in the body).

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -148,6 +142,61 @@ for field in spec.fields:
...
+init_default_strings = gen_init_func('initialize_defaults_only', True, False)
+}@
+  void initialize_all()
+  {
+@[for initstr in init_all_strings]@
+    @(initstr)
+@[end for]@
+  }
+
+  void zero_initialize()
```
dirk-thomas: To be consistent should this be named `initialize_zero` instead?

rosidl_generator_cpp/resource/msg__struct.hpp.em
clalancette: Yeah, good call on both of these points.  Will fix.

rosidl_generator_cpp/resource/msg__struct.hpp.em
clalancette: Yeah, definitely, will fix.

rosidl_generator_cpp/resource/msg__struct.hpp.em
clalancette: Yeah, I actually think I prefer to put each on its own line so that we never run into line length problems.  I'll change the code to do that.

clalancette:
What I just pushed has most of the comments addressed.  The two that haven't yet been addressed are @dirk-thomas really good points about complex fields and arrays.  I'm still working on those, and the code may change a bunch because of that, so it is probably a good idea to hold off on reviewing until I have something more.

clalancette:
All right.  I think I have all of the cases done, so I'm going to run another CI.  It's kind of hard to see the big picture amongst the empy control structures, and the fact that things are split in two places, so I'll spell it out below.  Let me know if this logic seems correct, and if I should add this as a comment in the code somewhere.

1.  Arrays with bounded size or unbounded size don't get any special initialization done, since they store no objects at construction time.
1.  Any field which is a non-array, primitive type gets initialized by the initialize_*() functions, controlled by the _init argument.
1.  Arrays of primitives with fixed size get each of their elements initialized by the initialize_*() functions, controlled by the _init argument.
1.  Arrays of non-primitives with fixed size get each of their objects initialized by the braced-initializer-list, and get _init argument passed in.  The allocator versions also get the _alloc argument passed in.
1.  As a special case, when constructing with an _alloc, _init pair, strings use a constructor initializer so we can pass in the allocator into the constructor.

Once CI is clean, I'm going to switch this back to "in review".

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -81,55 +82,62 @@ struct @(spec.base_type.type)_
...
-for field in spec.fields:
-    # default values of dynamic arrays with allocators if available
-    if alloc_name and field.type.is_array and \
-            (field.type.array_size is None or field.type.is_upper_bound) and \
-            field.type.is_primitive_type() and field.default_value is not None:
-        cpp_value = value_to_cpp(field.type, field.default_value)
-        print('    %s = %s;' % (field.name, cpp_value))
-}@
+    if (_init == rosidl_generator_cpp::MessageInitialization::ALL) {
+      this->initialize_all();
```
dirk-thomas: Please use consistent indentation in the Python code.

dirk-thomas:
Looking at the generated code this adds three methods to each message class. Until now we didn't define any methods on message class (beside operators) in order to not restrict the namespace. I would like to keep it that way. The functionality could be moved into "private" functions within the `.cpp` file or if we really like to expose these functions they could additionally be declared in the header.

rosidl_generator_cpp/test/test_initialization.cpp
wjwwood: If you don't run it now, I'd recommend running this test in a Windows debug CI job, because it's the only one that I know of that does something to the memory in order to check for uninitialized memory accesses. Though since you're doing memset before it _should_ be ok.

clalancette: Thanks for the heads-up.  I'll run a Windows Debug job to be sure.

clalancette:
> Looking at the generated code this adds three methods to each message class. Until now we didn't define any methods on message class (beside operators) in order to not restrict the namespace. I would like to keep it that way. The functionality could be moved into "private" functions within the .cpp file or if we really like to expose these functions they could additionally be declared in the header.

As we discussed offline, for now I put a new function into the header file that does the initialization.  This fixes the namespacing issue, with the downside that every file that `#include` any of these headers has to compile the code.  Moving it to a `.cpp` file is a potential optimization in the future.  I'm going to run another CI here and see what we get.

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -219,6 +256,43 @@ for field in spec.fields:
 using @(spec.base_type.type) =
     @(cpp_full_name)<std::allocator<void>>;
 
+@[if need_primitive_init]@
+template<class ContainerAllocator>
+void @(spec.base_type.type)_primitive_initializer(rosidl_generator_cpp::MessageInitialization _init, @(spec.base_type.type)_<ContainerAllocator> * _obj)
+{
+  if (_init == rosidl_generator_cpp::MessageInitialization::ALL) {
+@[ for name,value in init_all_strings]@
```
dirk-thomas: Style: space after comma.

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -219,6 +256,43 @@ for field in spec.fields:
 using @(spec.base_type.type) =
     @(cpp_full_name)<std::allocator<void>>;
 
+@[if need_primitive_init]@
+template<class ContainerAllocator>
+void @(spec.base_type.type)_primitive_initializer(rosidl_generator_cpp::MessageInitialization _init, @(spec.base_type.type)_<ContainerAllocator> * _obj)
```
dirk-thomas: Why `primitive` in the function name? Doesn't it also handle initialization of arrays and potentially complex types?

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -219,6 +256,43 @@ for field in spec.fields:
...
+{
+  if (_init == rosidl_generator_cpp::MessageInitialization::ALL) {
```
dirk-thomas: Style: constant left, variable right of operator.

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -219,6 +256,43 @@ for field in spec.fields:
...
+@[ for name,value in init_all_strings]@
+    _obj->@(name) = @(value);
+@[ end for]@
+@[ for name,value in array_init_all]@
+    for (auto & elem : _obj->@(name)) {
+      elem = @(value);
+    }
```
dirk-thomas: Should this use something like `std::fill` instead?

Same below.

clalancette:
> I don't like the idea that the initialization happens in two different locations. Some part in the constructor, some part in the helper function.
> 
> A user might be tempted to call the global function but would only partially what he expects.

I don't love that bit of it either, but in order to be able to pass the initializer and allocator into the complex types, those must be done in the initializer list, and in order to not pollute the namespace, we need a separate function, not a method of the class (private or otherwise).  The only other way I can see to meet the requirements of not polluting the namespace and initializing all of the primitive types is to do the initialization inline in the constructors.  Do you have another idea on how to do it?

(your other comments are easy enough to fix, I'll address those once we figure this part out)

dirk-thomas:
> The only other way I can see to meet the requirements of not polluting the namespace and initializing all of the primitive types is to do the initialization inline in the constructors. Do you have another idea on how to do it?

I agree with your conclusion that performing all the initialization in the constructor is the only option :+1:

dirk-thomas:
I guess I have to look at the generated code for a few different message definitions to see how this all fits together. With the template it is pretty difficult to imagine how this looks for a complex nested data structure.

I just thought about another case to consider. If a message definition has e.g. a default value for an array field it might not be good to pass the `MessageInitialization` one-to-one to the field constructor. E.g. when the actual array values are later set in the constructor body it would be more efficient to construct the vector without explicit initialization in the constructor initializer list.

clalancette:
The latest commits fix a bunch of bugs and most of the review comments, except for moving the initialization into the constructors.  I'll do that next.

clalancette:
All right.  I've moved the code into the constructors.  The downside to the current implementation is that there is a lot of repetition in the empy file, but I haven't come up with a good way to reduce that yet.  Thoughts/advice welcome.  CI is running on the latest code.

clalancette:
> I guess I have to look at the generated code for a few different message definitions to see how this all fits together. With the template it is pretty difficult to imagine how this looks for a complex nested data structure.

Yeah, this is basically how I've been debugging this.  If you look at `build_isolated/rosidl_generator_cpp/rosidl_generator_cpp/rosidl_generator_cpp/msg/primitives_some_default__struct.hpp`, it should have all of the cases in there (if not, we should add more test cases).

> I just thought about another case to consider. If a message definition has e.g. a default value for an array field it might not be good to pass the MessageInitialization one-to-one to the field constructor. E.g. when the actual array values are later set in the constructor body it would be more efficient to construct the vector without explicit initialization in the constructor initializer list.

I'm not 100% sure if I understand you on this one.  In the case of a bounded or unbounded vector with no defaults, no initialization is done at all (the default constructor for the std::vector/BoundedVector is called, of course, but we don't give it any parameters).  In the case of a bounded or unbounded vector with defaults, we initialize the vector to have a number of elements equivalent to the number of defaults, and then we set those defaults in the body (depending on _init).  In the case of a fixed-size vector, we initialize it with the correct number of elements, and then we assign those elements in the body (depending on _init).  I think the only case the tests aren't currently covering is a bounded/unbounded/fixed-size array of complex types; I'll need to add a test for that, but I don't actually think we support defaults for that anyway.  Does the above cover your concern, or were you thinking about something else?

mikaelarguedas:
Out-of-scope question: I didnt follow the patch in detail, I just wanted to point out that we want to be able to pass default values for not primitive types in the future and we should make sure that this is extensible to allow this use case.

e.g I have 2 messages: MsgA and MsgB

MsgA.msg:
```
int8 a_value 42
int8 another_value
```
MsgB.msg:
```
MsgA msga_value {"another_value":"53"}
```

How will the different initialization strategy behave in that use case?

dirk-thomas:
I took a closer look at the generated code. The conditional block in the constructors are not easy to read. Instead of:

```
if MessageInitialization::ALL
  ...
else if MessageInitialization::SKIP
  // empty
else if MessageInitialization::ZERO
  ...
else if MessageInitialization::DEFAULTS_ONLY
  ...
```
where each section contains partially the same code maybe the following structure is easier to read and understand (I can work on that update since I suggested it if others think its a good idea):

```
if MessageInitialization::ALL || MessageInitialization::DEFAULTS_ONLY
  init field1 with default
  if MessageInitialization::ALL
    init field2 without default
  if MessageInitialization::ALL
    init field3 without default
  init field4 with default
elseif MessageInitialization::ZERO
  ...
else if MessageInitialization::SKIP
  // empty
else
  assert / throw
```

dirk-thomas:
Unrelated question: I assume we also want to provide the same functionality for the C struct (e.g. `MsgName_new(rosidl_runtime_c_message_initialization)`). Is the goal to do that together or at a later point in time?

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -81,55 +83,140 @@ struct @(spec.base_type.type)_
...
-@# generate initializer lists
-@[  if alloc_name == '_alloc']@
-// *INDENT-OFF* (prevent uncrustify from making unnecessary indents here)
-@[  end if]@
 @{
-used_alloc_name = False
-leading_colon = ':'
+def array_init_list(field, args):
+    tmp = [field.type.type + args] * field.type.array_size
+    return "%s{{%s}}" % (field.name, ', '.join(tmp))
```
dirk-thomas: For a static array of size `N` this will generate the following in source code:

```
field_name{{SubmsgClass(_init), SubmsgClass(_init), ... <repeated N times>}}
```

For larger `N` that is pretty extreme. Imagine someone using `geometry_msgs::Vector3[1024] a_few_vectors`. In this case I don't think we can use the initialization list.

dirk-thomas:
@clalancette Regarding my previous comment about nesting: let me try to clarify.

The current implementation passes the `MessageInitialization` type recursively to all complex types. So in the case where the root message doesn't have default values for the complex field that seems to make sense.

But what if the root message defines a default value for something like this: `OtherMsg[3] field_name "somehow describe the default value for field_name"`. If the user chooses to initialize "all" I am worried that the generated code will do something like this which would perform initialization multiple times:

* create the array with three elements
* initialize each element based on the defaults / zeros defined in the constructor of `OtherMsg`
* overwrite the fields of all elements with the values provided by the default value for `field_name`

Maybe we don't have to think too much about it since we don't support default values for complex types yet. It was just something which crossed my mind...

clalancette:
> remove unnecessary NOLINT

I originally put in the NOLINT because cpplint was complaining about something, but it is looking like it doesn't need it anymore.  uncrustify is unhappy with the indentation on that patch, though, so I removed the extra 2 spaces you added.

clalancette:
> Out-of-scope question: I didnt follow the patch in detail, I just wanted to point out that we want to be able to pass default values for not primitive types in the future and we should make sure that this is extensible to allow this use case.

Yeah, good question.  As the code stands, it will not actually do the initialization.  However, it would not be too difficult to add; in your example, when the msgB constructor gets called, it could call the appropriate `->set*()` methods on the embedded msgA.

clalancette:
> Unrelated question: I assume we also want to provide the same functionality for the C struct (e.g. MsgName_new(rosidl_runtime_c_message_initialization)). Is the goal to do that together or at a later point in time?

Yeah, I want to do this for C as well, but I don't think I'm going to get to it in this PR.  This is already growing larger than I wanted :).

clalancette:
> where each section contains partially the same code maybe the following structure is easier to read and understand (I can work on that update since I suggested it if others think its a good idea):
> 
> if MessageInitialization::ALL || MessageInitialization::DEFAULTS_ONLY
>   init field1 with default
>   if MessageInitialization::ALL
>     init field2 without default
>   if MessageInitialization::ALL
>     init field3 without default
>   init field4 with default
> elseif MessageInitialization::ZERO
>   ...
> else if MessageInitialization::SKIP
>   // empty
> else
>   assert / throw

I did a prototype to see how this would look (I had to hand-edit a bit to make it work).  You can compare for yourself here:

Current code:
https://gist.github.com/clalancette/05bf2a3e9a6278a08e773a757b51cf94

Prototype of @dirk-thomas 's idea:
https://gist.github.com/clalancette/446e3e45212be4df681957bcce77b09a

Personally, I don't think the new idea makes it much more readable, and it slightly increases the size of the generated code.  Thus, my vote would be to keep it as-is; I agree that it isn't the greatest, but I'm not quite sure how else to improve it.

rosidl_generator_cpp/resource/msg__struct.hpp.em
clalancette: Yeah, I worried about this, and kicked the can down the road.  Now we've reached it again :).  The real problem is that `std::array` element construction/initialization happens before the body of the constructor, so we can't defer it until then and make sure we pass `_init` to the constructor.  The other way I can think of to fix this is to use a `std::shared_ptr<std::array>` instead of `std::array` (and then do a `make_shared` in the constructor body), but that makes accessing fixed-size arrays different from accessing everything else.  Other ideas?

dirk-thomas:
> I did a prototype to see how this would look (I had to hand-edit a bit to make it work). You can compare for yourself here:

Looking at the two side-by-side I don't like either of them :wink: I think what would help the readability is to keep the fields in order and not have separate sections doing different initialization of the same field. Let me try to make an example:

```
if (ALL || ZERO) {
  fieldWithoutDefault1 = 0  // or std::fill(...)
}

if (ALL || ZERO) {
  fieldWithoutDefault2 = 0  // or std::fill(...)
}

if (ALL || DEFAULT) {
  fieldWithDefault3 = <whatever default value>
else if (ZERO) {
  fieldWithDefault3 = 0  // or std::fill(...)
}

if (ALL || DEFAULT) {
  fieldWithDefault4 = <whatever default value>
else if (ZERO) {
  fieldWithDefault4 = 0  // or std::fill(...)
}
```

Independent of this adjacent fields which are "the same" (both having a default value or both not having one) could use the same `if` / `else` blocks. So the above example would be the "worst" case if all even fields have defaults and all odd fields have no default. In other cases the `if` / `else` blocks are reduced.

rosidl_generator_cpp/resource/msg__struct.hpp.em
clalancette: As we discussed, the idea we are going to try here is:

1.  For fixed-size arrays with small numbers of members (say, 10), do the initialization of them in the initializer list like we are currently doing.
1.  For fixed-sized arrays that are larger than that, let the initializer run as it normally would (i.e. let it call the default constructor of the class, with no _init argument).  In the body of the constructor, we would then hand-construct objects with the appropriate _init argument, and replace the objects in the std::array with the new ones.  This has a bit of a performance impact, but should generally be acceptable.

dirk-thomas: Thinking about this: is that worth the effort and complexity in the template? Or should we just do 2. with the argument that for such small N the double initialization won't hurt us "significantly"?

clalancette:
New CI for all of the latest changes:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3285/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/560/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2616/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3323/)

clalancette:
CI is green, and I've fixed all of the comments from previous reviews.  I still will probably put a bit more documentation into the empy source explaining what it is trying to accomplish, but this is ready for code review.

mikaelarguedas:
You may want to rerun linters locally because this fails with the new version of uncrustify

clalancette:
> You may want to rerun linters locally because this fails with the new version of uncrustify

Heh, yeah, of course.  I'll do that now.

clalancette:
Another CI to deal with the new uncrustify:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3292/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/567/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2628/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3336/)

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -29,9 +29,12 @@ header_guard_variable = '__'.join([x.upper() for x in header_guard_parts]) + '_'
 #endif
 
 @{
+import em
```
dirk-thomas: I don't think an `em` template should import `em` again to evaluate template strings defined in this template.

Instead you should define a function doing the same as what the template is doing but with plain Python. Depending on the size of that code you might want to define that in the Python modules and only pass a function handle to the template.

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -81,55 +86,172 @@ struct @(spec.base_type.type)_
...
+        return False
+
+def default_cpp_value_from_type(type):
+    if type == 'string':
+        return ''
+    elif type in ['float32', 'float64']:
+        return '0.0'
+    elif type == 'bool':
+        return 'false';
+    return '0'
```
dirk-thomas: You might want to move this Python code into a module for better readability, linting, deduplication, etc since they don't seem to contain anything message specific.

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -81,55 +86,172 @@ struct @(spec.base_type.type)_
...
+@[    if memb.zero_need_array_override]@
+      this->@(memb.name).fill(@(memb.type) {@(fill_args)});
+@[    else]@
+      std::fill(this->@(memb.name).begin(), this->@(memb.name).end(), @(memb.zero_value[0]));
+@[    end if]@
+@[   else]@
+      this->@(memb.name) = @(memb.zero_value);
+@[   end if]@
+@[  end for]@
+"""
```
dirk-thomas: See comment about importing `em` above.

rosidl_generator_cpp/resource/msg__struct.hpp.em
clalancette: Good point.  Done in 91b5f94

rosidl_generator_cpp/resource/msg__struct.hpp.em
clalancette: I ended up doing this, and it sped up the generation time quite a bit.  So thanks for the advice.  The downside to what I did in 84e2a51 is that it now uses "print" again to generate code, which we had decided earlier wasn't great.  I'm not quite sure how to reconcile these two competing goals; additional advice welcome.  Also, I ended up moving much of the python code for the creation of the lists into the pure-python place in b49d699, which cleans up the template quite a bit and allows us to get linting, etc support.

dirk-thomas: Instead of printing within the functions you can return a list of string instead. I would also suggest to not embed the indentation within the function. So the functions return something like this:

```
def func():
    lines = []
    lines.append('some code;')
    ...
    return lines
```

Based on that your template can then apply the right indentation necessary in the context where it is being used:

```
    {
@[for line in generate_zero_string(membset, "_alloc, _init")]@
      @line
@[end for]@
    }
```

clalancette: Done in the latest commit 889cb07

rosidl_generator_cpp/msg/PrimitivesSomeDefault.msg
```diff
@@ -0,0 +1,17 @@
+float32 float32_value 1.125
+float64 float64_value 2.4
+uint64 uint64_value
+string string_value "bar"
+float32[10] float32_arr
+float64[3] float64_arr [8.5, 1.2, 3.4]
+float64[2] float64_arr2 [99.0, 2.3]
+string[3] string_arr
+# TODO(clalancette): default values for strings don't seem to work; why?
```
dirk-thomas: What about this todo? Should this be resolved before merging this (at least figuring out the "why")?

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -81,55 +86,107 @@ struct @(spec.base_type.type)_
...
+# http://design.ros2.org/articles/generated_interfaces_cpp.html#constructors
+# for a detailed explanation of the different _init parameters.
+init_list, alloc_list, member_list = create_init_alloc_and_member_lists(spec)
+
+def generate_default_string(membset):
+    strlist = []
+    for memb in membset.members:
+        if memb.default_value is not None:
+            if isinstance(memb.default_value, list):
+                if not memb.default_value or memb.default_value.count(memb.default_value[0]) == len(memb.default_value):
```
dirk-thomas: Maybe rewrite the logic as `all(v == memb.default_value[0] for v in memb.default_value)`. It e.g. can abort early on the first item not matching.

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -81,55 +86,107 @@ struct @(spec.base_type.type)_
...
+                    strlist.append("std::fill(this->%s.begin(), this->%s.end(), %s);" % (memb.name, memb.name, memb.default_value[0]))
+                else:
+                    for index, val in enumerate(memb.default_value):
+                        strlist.append("this->%s[%d] = %s;" % (memb.name, index, val))
```
dirk-thomas: Is this also used for `std::vector` or only for `std::array`? If it is it would be good to resize the vector before to the desired size to avoid repeated reallocation.

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -81,55 +86,107 @@ struct @(spec.base_type.type)_
...
-@{
-for field in spec.fields:
-    # default values of dynamic arrays with allocators if available
-    if alloc_name and field.type.is_array and \
-            (field.type.array_size is None or field.type.is_upper_bound) and \
-            field.type.is_primitive_type() and field.default_value is not None:
-        cpp_value = value_to_cpp(field.type, field.default_value)
-        print('    %s = %s;' % (field.name, cpp_value))
-}@
+@[if len(member_list) == 0]@
```
dirk-thomas: Simpler: `if not member_list:`

Same below.

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
```diff
@@ -176,7 +176,106 @@ def primitive_value_to_cpp(type_, value):
     assert False, "unknown primitive type '%s'" % type_.type
 
 
+def default_cpp_value_from_type(type_):
+    if type_ == 'string':
+        return ''
```
dirk-thomas: Should this be `return '""'`?

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
```diff
@@ -176,7 +176,106 @@ def primitive_value_to_cpp(type_, value):
...
+    elif type_ in ['float32', 'float64']:
+        return '0.0'
+    elif type_ == 'bool':
+        return 'false'
+    return '0'
```
dirk-thomas: Do unsigned types need a trailing `u`?

rosidl_generator_cpp/msg/PrimitivesSomeDefault.msg
dirk-thomas: Maybe update the TODO text then.

rosidl_generator_c/include/rosidl_generator_c/message_initialization.h
```diff
@@ -0,0 +1,28 @@
...
+#define ROSIDL_GENERATOR_C__MESSAGE_INITIALIZATION_H_
+
+// TODO(clalancette): this should be moved out into a separate runtime package
+
+enum rosidl_runtime_c_message_initialization
+{
+  ROSIDL_RUNTIME_C_MSG_INIT_ALL,
+  ROSIDL_RUNTIME_C_MSG_INIT_SKIP,
+  ROSIDL_RUNTIME_C_MSG_INIT_ZERO,
+  ROSIDL_RUNTIME_C_MSG_INIT_DEFAULTS_ONLY,
```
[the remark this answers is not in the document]
clalancette: Done in 114078e

rosidl_generator_cpp/msg/PrimitivesSomeDefault.msg
clalancette: Done in 114078e

rosidl_generator_cpp/resource/msg__struct.hpp.em
clalancette: Done in 114078e

rosidl_generator_cpp/resource/msg__struct.hpp.em
clalancette: It is used both for std::vector and for std::array in the case that default values are given.  I'll take a look at pre-allocation for vectors.

rosidl_generator_cpp/resource/msg__struct.hpp.em
clalancette: Done in 114078e

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
clalancette: As it stands, no.  The reason is that these values will all be passed through `primitive_value_to_cpp` (either directly or indirectly through `value_to_cpp`) below, which will make these "real" C++ code, including the quotes and any trailing modifiers.

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
clalancette: Same as above, `primitive_value_to_cpp` will handle this.

rosidl_generator_cpp/resource/msg__struct.hpp.em
clalancette: This is done in 01bccad.

clalancette:
I think I've addressed all of the review comments, so this should be ready for another review.

rosidl_generator_cpp/CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ project(rosidl_generator_cpp)
 
 find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_python REQUIRED)
+find_package(rosidl_generator_c REQUIRED)
+include_directories(${rosidl_generator_c_INCLUDE_DIRS})
```
dirk-thomas: Should these two lines be moved into the `BUILD_TESTING` block (and the `include_directories` call be merged with the existing call there)?

rosidl_generator_cpp/CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ project(rosidl_generator_cpp)
 
 find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_python REQUIRED)
+find_package(rosidl_generator_c REQUIRED)
```
dirk-thomas: Since the header of this package transitively exposes the header from `rosidl_generator_c` does it need to be an export dependency?

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -81,55 +86,111 @@ struct @(spec.base_type.type)_
...
+
+def generate_default_string(membset):
+    strlist = []
+    for memb in membset.members:
+        if memb.default_value is not None:
+            if memb.num_prealloc > 0:
+                strlist.append("this->%s.resize(%d);" % (memb.name, memb.num_prealloc))
+            if isinstance(memb.default_value, list):
+                if not memb.default_value or all(v == memb.default_value[0] for v in memb.default_value):
+                    strlist.append("std::fill(this->%s.begin(), this->%s.end(), %s);" % (memb.name, memb.name, memb.default_value[0]))
```
dirk-thomas: In the case the first condition is true (`not memb.default_value`) the list is empty. Would accessing `memb.default_value[0]` not fail?

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
```diff
@@ -176,7 +176,108 @@ def primitive_value_to_cpp(type_, value):
...
+                self.members.append(memb)
+                return True
+            new_def_none = memb.default_value is None
+            last_def_none = self.members[-1].default_value is None
+            new_zero_none = memb.zero_value is None
+            last_zero_none = self.members[-1].zero_value is None
+            if (new_def_none == last_def_none) and (new_zero_none == last_zero_none):
+                self.members.append(memb)
+                return True
+            return False
```
dirk-thomas: This code seems overly complicated to me. I would recommend something like this:

```
class Member:
    def same_default_and_zero_value(self, other):
        return self.default_value == other.default_value and self.zero_value == other.zero_value
```

This makes the code much easier to read:

```
class CommonMemberSet:
    def add_member(self, memb):
        if not self.members or self.members[-1].same_default_and_zero_value(memb):
            self.members.append(memb)
            return True
        return False
```

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -81,55 +86,111 @@ struct @(spec.base_type.type)_
...
+# The creation of the constructors for messages is a bit complicated.  The goal
+# is to have a constructor where the user can control how the fields of the
+# message get initialized via the _init parameter to the constructor.  See
+# http://design.ros2.org/articles/generated_interfaces_cpp.html#constructors
+# for a detailed explanation of the different _init parameters.
+init_list, alloc_list, member_list = create_init_alloc_and_member_lists(spec)
+
+def generate_default_string(membset):
+    strlist = []
+    for memb in membset.members:
```
dirk-thomas: We usually try to not abbreviate variable names. I don't see the two saved characters to be crucial here. Therefore I would suggest using `member` throughout the code.

rosidl_generator_cpp/CMakeLists.txt
clalancette: I agree with you, but the odd thing is that it was working without it (I would have expected the builds to fail to find the right header file).  In any case, I've now changed this to have an `ament_export_dependency()`.

rosidl_generator_cpp/CMakeLists.txt
clalancette: Done.

rosidl_generator_cpp/resource/msg__struct.hpp.em
clalancette: Sure, I changed to `member` everywhere.

rosidl_generator_cpp/resource/msg__struct.hpp.em
clalancette: Yeah, this is a great catch.  I'm not sure what I was trying to do now.  I've now hoisted the `if memb.default_value` up to a higher level, which should get rid of this concern.

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
clalancette: Yeah, that is a much better implementation.  I've implemented it this way instead.

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -81,55 +86,112 @@ struct @(spec.base_type.type)_
...
+    return strlist
+
+def generate_zero_string(membset, fill_args):
+    strlist = []
+    for member in membset.members:
+        if isinstance(member.zero_value, list):
+            if member.num_prealloc > 0:
+                strlist.append('this->%s.resize(%d);' % (member.name, member.num_prealloc))
+            if member.zero_need_array_override:
+                strlist.append('this->%s.fill(%s {%s});' % (member.name, member.type, fill_args))
```
dirk-thomas: Is there an example with this line expanded and being used? I just can't imagine how it is supposed to look like.

clalancette: If you run

`src/ament/ament_tools/scripts/ament.py test --build-tests --isolated --symlink-install --only rosidl_generator_cpp`

And then take a look at the `vec3_fixed` field in `./build_isolated/rosidl_generator_cpp/rosidl_generator_cpp/rosidl_generator_cpp/msg/primitives_some_default__struct.hpp`, that's an example that uses it.

dirk-thomas: The line expands to:

```
this->vec3_fixed.fill(MyVector3 {_init});
```

Looks good to me.

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -81,55 +86,112 @@ struct @(spec.base_type.type)_
...
-        cpp_value = value_to_cpp(field.type, field.default_value)
-        print('    %s = %s;' % (field.name, cpp_value))
-}@
+@[if not member_list]@
+    (void)_init;
+@[end if]@
+@[for membset in member_list]@
+@[ if membset.members[0].default_value is not None]@
+    if (rosidl_generator_cpp::MessageInitialization::ALL == _init ||
+        rosidl_generator_cpp::MessageInitialization::DEFAULTS_ONLY == _init) {
```
dirk-thomas: This line doesn't pass `uncrustify`. It needs to look like lines 151-152.

clalancette: Good point.  It should be fixed b41a4d00

clalancette:
Now that we are a closer, here's another CI run to make sure everything is happy:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3327/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/589/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2655/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3367/)

rosidl_generator_cpp/resource/msg__struct.hpp.em
dirk-thomas: The generated code still doesn't pass uncrustify. The open curly brace needs to be on the next line.

The `rosidl_generate_interfaces()` in the testing section should pass `ADD_LINTER_TESTS` to automatically lint the generated code.

clalancette: I'm not sure why it is passing locally, but not when run elsewhere.  I'll add the `ADD_LINTER_TESTS` to the code and see what turns up.

mikaelarguedas: @clalancette It seems that you addressed only the comment on this line and not the "same below" part, it still fails uncrustify locally

clalancette:
I've added ADD_LINTER_TESTS, and I'm now clean locally.  Another CI run to check:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3349/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/607/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2684/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3405/)

rosidl_generator_cpp/CMakeLists.txt
```diff
@@ -21,11 +23,14 @@ if(BUILD_TESTING)
 
     "msg/Empty.msg"
 
+    "msg/MyVector3.msg"
```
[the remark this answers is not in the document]
clalancette: Yeah, this is needed because I wanted to make sure we tested a "complex" type with some of the fields set to a default value, and some unset.  I didn't find any other messages that had this particular semantic in `rosidl_generator_cpp/msg`.  I agree with you that the name would be better as something else; I'll take your suggestion and change the name to `FieldsWithSameTypeSomeDefaults`.  As far as keeping the testing closer to the other generators, I would prefer to add additional tests to the other generators (including adding this message) in another PR.  Does that seem reasonable to you?

rosidl_generator_cpp/package.xml
```diff
@@ -7,6 +7,8 @@
   <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
   <license>Apache License 2.0</license>
 
+  <build_depend>rosidl_generator_c</build_depend>
```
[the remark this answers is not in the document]
clalancette: After this PR, when the code for messages is generated, they all have:

```
#include <rosidl_generator_cpp/message_initialization.hpp>
```

The `message_initialization.hpp` file has:

```
#include <rosidl_generator_c/message_initialization.h>
```

Thus, in order to build these messages, a dependency on `rosidl_generator_c` is required.  However, it is not required for building *this* package.  So I think you are correct; we only need this as a `build_export_depend` and a `test_depend`.  I'll make that change: b6fe0bf

rosidl_generator_cpp/test/test_initialization.cpp
```diff
@@ -0,0 +1,241 @@
...
+TEST(Test_initialization, zero_constructor) {
+  rosidl_generator_cpp::msg::PrimitivesSomeDefault def(
+    rosidl_generator_cpp::MessageInitialization::ZERO);
+
+  ASSERT_EQ(0.0f, def.float32_value);
+  ASSERT_EQ(0.0, def.float64_value);
+  ASSERT_EQ(0ULL, def.uint64_value);
+  ASSERT_EQ("", def.string_value);
+  ASSERT_TRUE(std::all_of(def.float32_arr.begin(), def.float32_arr.end(), [](float i) {
+      return i == 0.0f;
```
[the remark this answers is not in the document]
clalancette: Fixed in af73047

rosidl_generator_cpp/test/test_initialization.cpp
```diff
@@ -0,0 +1,241 @@
...
+    }));
+  ASSERT_TRUE(std::all_of(def.float64_arr.begin(), def.float64_arr.end(), [](double i) {
+      return i == 0.0;
+    }));
+  ASSERT_TRUE(std::all_of(def.string_arr.begin(), def.string_arr.end(), [](std::string i) {
+      return i == "";
+    }));
+  ASSERT_EQ(2UL, def.unbounded.size());
+  ASSERT_EQ(def.unbounded[0], 0.0f);
```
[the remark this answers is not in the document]
clalancette: Fixed in af73047

rosidl_generator_cpp/CMakeLists.txt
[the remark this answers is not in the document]
mikaelarguedas: Yeah totally, I meant having a message definition we could copy to the other generators packages for testing. It makes total sense not to touch the other generators in this PR but in follow-up PRs when we add the initialization features to them :+1: .

rosidl_generator_cpp/resource/msg__struct.hpp.em
mikaelarguedas: Looking at this a bit more: I have a question that I'm not sure to be able to answer just based on the design doc for interfaces.
Should I be able to use a message from the same package in a service definition without specifying the package it comes from?
Example because I know the previous sentence is not clear:
I have a package as follow:
```
├── CMakeLists.txt
├── msg
│   └── foo.msg
├── package.xml
└── srv
    └── bar.srv
```
Does bar.srv **have to** be:
```
pkgA/foo[] fools
---
pkgA/foo[3] threefools
```
or is the following a valid definition?
```
foo[] fools
---
foo[3] threefools
```

Reading the [design doc](http://design.ros2.org/articles/interface_definition.html):
```
The base type can be one of the following:

- a primitive type from the list above: e.g. int32
- a string with an upper boundary: string<=N to limit the length of the string to N characters
- a complex type referencing another message:
- an absolute reference of a message: e.g. some_package/SomeMessage
- a relative reference of a message within the same package: e.g. OtherMessage
```
 it seems that the answer should be "both are valid". 

If both are valid, we need to update this part of the PR (and maybe others?) to use the full c++ namespace of the message and not only the type name.
`this->vec3_fixed.fill(MyVector3 {_init});` -> `this->vec3_fixed.fill(rosidl_generator_cpp::msg::MyVector3 {_init});`

If only the definition using "absolute references" is allowed, we should update the design document to make it clearer.

Example branch that fails to compile with the current patch: https://github.com/ros2/rcl_interfaces/tree/more_services

rosidl_generator_cpp/msg/PrimitivesSomeDefault.msg
```diff
@@ -0,0 +1,18 @@
...
+# TODO(clalancette): default values for strings aren't currently implemented;
+# see https://github.com/ros2/rosidl/issues/206
+#string[3] string_def_arr ["hello", "there", "world"]
+float32[] unbounded [1, 2]
+float32[<=3] bounded_no_def
+float32[<=3] bounded_def [3, 4]
+FieldsWithSameTypeSomeDefaults vec3
+FieldsWithSameTypeSomeDefaults[2] vec3_fixed
+FieldsWithSameTypeSomeDefaults[] vec3_unbounded
+FieldsWithSameTypeSomeDefaults[<=2] vec3_bounded
```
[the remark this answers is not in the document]
clalancette: The point of this particular .msg file is to put as many different things together as possible (that way, if you are debugging this stuff, there are a limited number of places to look).  So I'd prefer to leave them here, unless I misunderstood your point.  Despite the fact that they are now of type `FieldsWithSameTypeSomeDefaults`, they are still vectors, and `vec3` is short :).  However, if you think I should rename them, I can do that.  Any suggestions?

rosidl_generator_cpp/resource/msg__struct.hpp.em
clalancette: This is an excellent point.  Based on the design document, I agree with your reasoning, and relative should be allowed.  I'll take a closer look at your branch and see what I have to do to fix it here.  Thanks for the heads up.

rosidl_generator_cpp/msg/PrimitivesSomeDefault.msg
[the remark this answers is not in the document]
mikaelarguedas: Oh yeah the vec3 could be renamed but it's not a big deal.
I was referring to the name of this file that is "PrimitivesSomeDefault" and this message contains non-primitive types. If we want this to be a message definition with various types (primitives and complex fields) we may want to rename it. Some other packages have a [Various](https://github.com/ros2/rosidl/blob/master/rosidl_generator_py/msg/Various.msg) message definition for example. That's not a blocker, but something we may want to consider when updating the other generators, to have the same message definitions everywhere and test the same things on all generators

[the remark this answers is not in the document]
clalancette: Ah, I see, I misunderstood before.  Now it makes sense.  I'll rename the top-level to something more descriptive after investigating the bigger issue about namespaces below.

clalancette:
It turned out that the problem that @mikaelarguedas was talking about wasn't really a design problem so much as a bug in my implementation.  The problem was that `srv` and `msg` are in different namespaces, so when doing the zero-fill array thing in a `srv`, it didn't have access to the `msg` symbols directly.  The fix was to generate the appropriate namespace-complete symbol, which fixes the problem (I also added a test for this case).  Thanks for catching that @mikaelarguedas!  I think this is ready for another review now.

dirk-thomas:
Why do the services need a separate initialization test? Each part of a service is just a message so their initialization should be covered by the message tests already.

clalancette:
> Why do the services need a separate initialization test? Each part of a service is just a message so their initialization should be covered by the message tests already.

The service test is to ensure that the generator correctly generates fully-namespaced objects when re-initializing zero-filled objects.  This is a bit dense, so let me show an example based on what Mikael talked about earlier.

Let's assume you have a service with a fixed-size array of a complex type for the request (like https://github.com/ros2/rosidl/pull/236/files#diff-690610169303fba9f4575f67fb3590a9).  The generated C++ type for that field will be `std::array<FieldsWithSameTypeSomeDefaults, 2>`.  Because we ended up not wanting to have huge initializer lists, those complex fields will always have their default constructor run, which will initialize all of the fields to the default values.  In the case that the user passed `init = ZERO`, we re-initialize those fields with a zero-based object.  Before the latest changes, that generated code looked like the following:

```
this->req.fill(FieldsWithSameTypeSomeDefaults_<ContainerAllocator>{_init});
```

However, this actually fails to compile.  The reason is that the service is in the namespace `rosidl_generator_cpp::srv`, while the complex type is in the namespace `rosidl_generator_cpp::msg`, with no using statement.  The latest code changes the above code to:

```
this->req.fill(rosidl_generator_cpp::msg::FieldsWithSameTypeSomeDefaults_<ContainerAllocator>{_init});
```

And the new test is there to ensure that the rosidl generator is generating the correct code.  To be honest, this test is mostly a compile-test; if the compilation succeeds, we've generated the correct thing, and we don't need to go much further than that.  However, I thought it wouldn't hurt much to add a couple of checks in there as well.  I can remove the bulk of the code from `test_srv_initialization.cpp` and add a comment explaining that this is a compile test if you think that is better.  Other suggestions on how to do this compile test welcome.

mikaelarguedas:
I don't feel strongly about testing the content or not given that the initialization itself is the same as for messages, but as you said I can't hurt.
If the decision is to make this a compile test only, we could get rid of the test code completely and just keep the generation of the service and corresponding Request/Response messages

PR #237, opened 2017-09-06
fix ref count of Python messages

dirk-thomas:
Connect to ros2/rclpy#114.

Fixes the ref counts of the message fields.

### DIFF
--- a/rosidl_generator_py/resource/_msg_support.c.em
+++ b/rosidl_generator_py/resource/_msg_support.c.em
@@ -61,10 +61,18 @@ full_classname = '%s.%s._%s.%s' % (spec.base_type.pkg_name, subfolder, module_na
 }@
   char full_classname_dest[@(len(full_classname) + 1)];
 
-  char * class_name = (char *)PyUnicode_1BYTE_DATA(
-    PyObject_GetAttrString(PyObject_GetAttrString(_pymsg, "__class__"), "__name__"));
-  char * module_name = (char *)PyUnicode_1BYTE_DATA(
-    PyObject_GetAttrString(PyObject_GetAttrString(_pymsg, "__class__"), "__module__"));
+  char * class_name = NULL;
+  char * module_name = NULL;
+  {
+    PyObject * attr1 = PyObject_GetAttrString(_pymsg, "__class__");
+    PyObject * attr2 = PyObject_GetAttrString(attr1, "__name__");
+    class_name = (char *)PyUnicode_1BYTE_DATA(attr2);
+    Py_DECREF(attr2);
+    attr2 = PyObject_GetAttrString(attr1, "__module__");
+    module_name = (char *)PyUnicode_1BYTE_DATA(attr2);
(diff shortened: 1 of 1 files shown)


rosidl_generator_py/resource/_msg_support.c.em
```diff
@@ -61,10 +61,18 @@ full_classname = '%s.%s._%s.%s' % (spec.base_type.pkg_name, subfolder, module_na
...
+  {
+    PyObject * attr1 = PyObject_GetAttrString(_pymsg, "__class__");
+    PyObject * attr2 = PyObject_GetAttrString(attr1, "__name__");
+    class_name = (char *)PyUnicode_1BYTE_DATA(attr2);
+    Py_DECREF(attr2);
+    attr2 = PyObject_GetAttrString(attr1, "__module__");
+    module_name = (char *)PyUnicode_1BYTE_DATA(attr2);
+    Py_DECREF(attr2);
+    Py_DECREF(attr1);
+  }
```
mikaelarguedas: Question: why is the new logic scoped everywhere in this PR ?

dirk-thomas: That makes the scope of these temporary variables as small as possible. It makes it clear that they are not used beyond these few lines.

rosidl_generator_py/resource/_msg_support.c.em
```diff
@@ -327,7 +346,8 @@ nested_type = '%s__%s__%s' % (field.type.pkg_name, 'msg', field.type.type)
 @[    elif field.type.type == 'string']@
     PyList_SetItem(py@(field.name), idx@(field.name), PyUnicode_FromString(tmpmessagedata@(field.name)[idx@(field.name)].data));
 @[    elif field.type.type == 'bool']@
-    PyList_SetItem(py@(field.name), idx@(field.name), tmpmessagedata@(field.name)[idx@(field.name)] ? Py_True : Py_False);
+@# using PyBool_FromLong because PyList_SetItem will steal ownership of the passed item
+    PyList_SetItem(py@(field.name), idx@(field.name), PyBool_FromLong(tmpmessagedata@(field.name)[idx@(field.name)] ? 1 : 0));
```
dirk-thomas: This was slightly tricky to find :angry:

Same below.

mikaelarguedas: pretty unconvenient that we need to use this hack because `Py_False` and `Py_True` don't give the right refcount here... Looking at the doc that seems the be the only suitable workaround :+1:

dirk-thomas: Well, we could also not call `Py_DECREF` for this case - but that would make our template logic more complex. So I would say it is only necessary since we want to treat all types uniformly.

mikaelarguedas: Agreed, and I think we agree that the template logic doesnt need additional complexity ;)

PR #240, opened 2017-10-16
return NULL in case some Python function fails

dirk-thomas:
Follow up of ros2/rclpy#95.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3341/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/600/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2671/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3397/)

Ready for review.

### DIFF
--- a/rosidl_generator_py/resource/_msg_support.c.em
+++ b/rosidl_generator_py/resource/_msg_support.c.em
@@ -402,8 +402,13 @@ nested_type = '%s__%s__%s' % (field.type.pkg_name, 'msg', field.type.type)
   assert(false);
 @[  end if]@
   assert(py@(field.name) != NULL);
-  PyObject_SetAttrString(_pymessage, "@(field.name)", py@(field.name));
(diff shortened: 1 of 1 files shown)


dirk-thomas:
> in opposition to the other functions that can raise in `convert_to_py`

Just for the record: they all return error codes - almost none of them raises exceptions.

mikaelarguedas:
> Just for the record: they all return error codes - almost none of them raises exceptions

Yes with a few "exceptions" :stuck_out_tongue_winking_eye: 
I agree only a handful of them can raise all the other ones are "safe"

dirk-thomas:
> I agree only a handful of them can raise all the other ones are "safe"

The problem are not the potential exceptions here. The problem is that some functions return an error code which our code either ignores or `assert`s on.

PR #248, opened 2017-11-21
Homogenize type support header name

mikaelarguedas:
Stumble upon this inconsistency in naming when playing with service type support for parameters.
- First commit renames the file
- Second commit adapt the header guard in the file
- 3rd commit updates include statement of all files including this one

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3587/) (test failure unrelated)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/780/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2927/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3677/)


mikaelarguedas:
> This looks good to me once CI is green. The only question I have is do we want to shorten the other name vs extend this on? I'm fine with either.

Yeah we could go either way. I don't know if originally it was suffixed `_struct` to contrast with the `message_type_support.h` provided by each type support.

Edit: looking a the history, it looks like the renaming was to prevent collision (https://github.com/ros2/rosidl/pull/71) so it seems that the approach of adding the `_struct` to the service one makes more sense

PR #250, opened 2017-11-29
Cpp fix bool init

mikaelarguedas:
When debugging parameters I noticed that boolean were default initialized to true since we refactored the message initialization.
The first commit fixes the issue
The second commit extends the test suite to test all primitive types' initialization

(this is obviously not the proper fix as the code should behave the same if the passed value was the expected bool, my take is that `primitive_value_to_cpp` expects an actual value of the right Python builtin type but is now called in multiple places with the result of `default_cpp_value_from_type` that returns a string of questionable content)

### DIFF
--- a/rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
+++ b/rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
@@ -144,6 +144,22 @@ def msg_type_to_cpp(type_):
 
 
 def value_to_cpp(type_, value):
+    """
+    Convert a python value into a string representing that value in C++.
+
+    This is equivalent to primitive_value_to_cpp but can process arrays values as well
+
+    Warning this still processes only primitive types
+    @param type_: a ROS IDL type
+    @type type_: builtin.str
+    @param value: the value to convert
+    @type value: python builtin (bool, int, float, str or list)
+    @returns: a string containing the C++ representation of the value
+    """
+    # :raises: ValueError on if the provided type_ is unknown
+    # """
+    # TODO(mikaelarguedas) this should raise proper exceptions rather than asserting
+    # without error message
     assert type_.is_primitive_type()
     assert value is not None
 
@@ -163,6 +179,20 @@ def value_to_cpp(type_, value):
 
 
 def primitive_value_to_cpp(type_, value):
+    """
+    Convert a python value into a string representing that value in C++.
+
+    Warning: The value has to be a primitive and not a list
+      (aka this function doesn't work for arrays)
+    @param type_: a ROS IDL type
+    @type type_: builtin.str
+    @param value: the value to convert
+    @type value: python builtin (bool, int, float or str)
+    @returns: a string containing the C++ representation of the value
+    """
+    # :raises: ValueError on if the provided type_ is unknown
+    # TODO(mikaelarguedas) this should raise proper exceptions rather than asserting
+    # without error message
     assert type_.is_primitive_type()
     assert value is not None
 
@@ -196,6 +226,7 @@ def primitive_value_to_cpp(type_, value):
     if type_.type == 'string':
         return '"%s"' % escape_string(value)
 
+    # raise ValueError("unknown primitive type '%s'" % type_.type)
     assert False, "unknown primitive type '%s'" % type_.type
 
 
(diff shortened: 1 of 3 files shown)


mikaelarguedas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3642/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/822/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2975/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3742/)

clalancette:
We discussed this, and @mikaelarguedas is correct; `default_cpp_value_from_type` is confused about which kind of types it should return.  After a closer look, it really should return a python type, and then that should be passed into `primitive_value_to_cpp` (either directly or indirectly through `value_to_cpp`) to get the cpp value.  That should fix the issue correctly.  I've done something like this in 41aa5ff6af4faee93cef334f85056ae8813afe5b; feel free to grab that.

mikaelarguedas:
Added a bit more cleanup and documentation. Tested a few other message files locally to validate the default values.

Now the default values seem fine for all types. I faced a few surprising things regarding memory allocation as well as ZERO init values but I'll open a follow up issue for this as I want to confirm what the expected behavior is before considering a fix.

Ready for review

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
```diff
@@ -144,6 +144,22 @@ def msg_type_to_cpp(type_):
...
+
+    Warning this still processes only primitive types
+    @param type_: a ROS IDL type
+    @type type_: builtin.str
+    @param value: the value to convert
+    @type value: python builtin (bool, int, float, str or list)
+    @returns: a string containing the C++ representation of the value
+    """
+    # :raises: ValueError on if the provided type_ is unknown
+    # """
```
clalancette: Minor nit; it's a little weird to have triple-quotes here and commented-out.  I'd just remove it, as it is syntax.

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
```diff
@@ -144,6 +144,22 @@ def msg_type_to_cpp(type_):
...
+    # TODO(mikaelarguedas) this should raise proper exceptions rather than asserting
+    # without error message
```
clalancette: Yeah, totally agreed :).

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
```diff
@@ -174,31 +204,40 @@ def primitive_value_to_cpp(type_, value):
...
+    if type_.type == 'uint64':
+        return '%sull' % value
+
+    if type_.type == 'float32':
+        return '%sf' % value
+
     if type_.type == 'string':
         return '"%s"' % escape_string(value)
 
+    # raise ValueError("unknown primitive type '%s'" % type_.type)
```
clalancette: I'd remove this commented out code; we already have a TODO at the top of the function.

mikaelarguedas: yeah that's a remnant I started raising but figured we needed a more thorough approach than just raise in a handful of places so I added the TODO on top and commented out my previous docblocks and raise. I'll remove them

mikaelarguedas:
I won't rerun CI as https://github.com/ros2/rosidl/pull/250/commits/7719972dc283762e946cc6511ba12c1379909215 only removed commented out lines

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
```diff
@@ -144,6 +144,20 @@ def msg_type_to_cpp(type_):
...
+
+    Warning this still processes only primitive types
+    @param type_: a ROS IDL type
+    @type type_: builtin.str
+    @param value: the value to convert
+    @type value: python builtin (bool, int, float, str or list)
+    @returns: a string containing the C++ representation of the value
+    """
+    # TODO(mikaelarguedas) this should raise proper exceptions rather than asserting
+    # without error message
```
dirk-thomas: Imo asserting here is appropriate since this should only ensure that the function is being used as intended in the templates. An assertion message would indeed be helpful.

Same below.

PR #252, opened 2017-12-01
enforce membership to rosidl_interface_packages

mikaelarguedas:
This enforces that any package installing interfaces declares membership to the `rosidl_interface_packages` group.

requires https://github.com/ament/ament_cmake/pull/119

### DIFF
--- a/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
+++ b/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
@@ -114,6 +114,16 @@ macro(rosidl_generate_interfaces target)
   )
 
   if(NOT _ARG_SKIP_INSTALL)
+    set(_interface_packages_group_name "rosidl_interface_packages")
+    if(NOT _AMENT_PACKAGE_NAME)
+      ament_package_xml()
+    endif()
+    if(NOT _interface_packages_group_name IN_LIST ${_AMENT_PACKAGE_NAME}_MEMBER_OF_GROUPS)
+      message(FATAL_ERROR
+        "Packages installing interfaces must include \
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
CI failing because of some turtlebot packages not being updated yet (look for "rosidl_interface_packages" in the console output): [Build Status](http://ci.ros2.org/job/ci_turtlebot-demo_linux/106/)

CI only on ros2.repos:
[Build Status](http://ci.ros2.org/job/ci_linux/3668/)

I'l open the PRs necessary to get the turtlebot repos to build before merging this

mikaelarguedas:
line length fixup: https://github.com/ros2/rosidl/commit/69956510f5a57500b2f51fc14484c0ff35060b8d

dirk-thomas:
Is it really justified to fail the build if a package containing interfaces doesn't declare this? Wouldn't a warning (which the user should be able to disable explicitly) be more appropriate?

mikaelarguedas:
That was a concern I had indeed. the risk being that if we don't enforce it likely people will not follow it.
I don't feel very strongly either way. Happy to change it if we think people shouldn't **have to** add it.

wjwwood:
> Is it really justified to fail the build if a package containing interfaces doesn't declare this?

I think so, I don't see a reason for some to install message files and not be part of this group. I'm fine to have a way to disable the check when installing message files (for the unforeseen use case), but I don't think a warning is better than an error, because I bet people will miss it (many people don't see all the build output due to it flying past).

dirk-thomas:
> I'm fine to have a way to disable the check when installing message files (for the unforeseen use case).

That would be fine with me too.

> (many people don't see all the build output due to it flying past)

I consider this to be a flaw of the build tool. I expect that the user gets presented any `stderr` output in the summary of a build (which is what I have already implemented in the next-gen build tool).

PR #264, opened 2018-02-28
update style

dirk-thomas:
Connect to ~~ament/uncrustify#20~~. Connect to ament/uncrustify#21.

The second commit fixes a comment referencing the wrong template name.

Ready for review.

### DIFF
--- a/rosidl_generator_cpp/include/rosidl_generator_cpp/bounded_vector.hpp
+++ b/rosidl_generator_cpp/include/rosidl_generator_cpp/bounded_vector.hpp
@@ -54,7 +54,7 @@ public:
 
   /// Create a %BoundedVector with no elements.
   BoundedVector()
-  noexcept(std::is_nothrow_default_constructible<_Alloc>::value)
+  noexcept (std::is_nothrow_default_constructible<_Alloc>::value)
   : _Base()
   {}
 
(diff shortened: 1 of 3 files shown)


rosidl_generator_cpp/include/rosidl_generator_cpp/bounded_vector.hpp
```diff
@@ -54,7 +54,7 @@ class BoundedVector
 
   /// Create a %BoundedVector with no elements.
   BoundedVector()
-  noexcept(std::is_nothrow_default_constructible<_Alloc>::value)
+  noexcept (std::is_nothrow_default_constructible<_Alloc>::value)
```
wjwwood: I don't think this is right, since it's actually calling `noexcept(...)`: http://en.cppreference.com/w/cpp/language/noexcept

dirk-thomas: This is what it takes to comply with `uncrustify`. Please feel free to fill a ticket upstream. (I just don't think it is worth it.)

dirk-thomas: I anyway took the time to fill a ticket upstream: uncrustify/uncrustify#1649

I don't think we should be waiting for that though.

wjwwood: Sounds good.

Issue #271, opened 2018-04-20
Generate MISRA compatible constants


mjcarroll:
#### Feature description

It has been brought to our attention that the message constants that are generated for ROS2 are not compatible with writing MISRA-compliant code.

Let's consider a ROS2 message file `PointBlock.msg` that apart from other things containing a constant:
```
...
uint32 CAPACITY=256
...
```

One of resulting generated files is `PointBlock.msg point_block__struct.hpp`:
```cpp
...
  // constants
  enum { CAPACITY = 256u };
...
```

Now because of `CAPACITY` is a class-less enumeration and is not signed, we can't use it in assignments to and comparisons with `uint32_t` variables.
```
implicit conversion from 'unsigned int' to 
'enum apex_app_msgs::msg::PointBlock_<std::allocator<void> >::(anonymous at point_block__struct.hpp:113:3)'
changes signedness of underlying type
```

mjcarroll:
@dirk-thomas Do you recall the reason why `rosidl` wraps constants in an enum?

mjcarroll:
I think that the issue stems from the fact that the underlying type of the `enum`, which I believe is at least a signed type (I think width is implementation specific).
Because of this, it makes the implicit conversion have comparison between different signs.

A few alternatives would be:
* Explicitly type the enums (available in  C++11 and up): ` enum: uint32_t { CAPACITY = 256u; }`
* Use `static const` for integral constants, the way we do for everything else.  I noticed that we don't consistently use one way or another.
```
 // constants
  static const bool BOOL_CONST;
  enum { BYTE_CONST = 50 };
  enum { CHAR_CONST = 100 };
  static const float FLOAT32_CONST;
  static const double FLOAT64_CONST;
  enum { INT8_CONST = -50 };
  enum { UINT8_CONST = 200u };
  enum { INT16_CONST = -1000 };
  enum { UINT16_CONST = 2000u };
  enum { INT32_CONST = -30000 };
  enum { UINT32_CONST = 60000u };
  enum { INT64_CONST = -40000000 };
  enum { UINT64_CONST = 50000000u };
  static const std::basic_string<char, std::char_traits<char>, typename ContainerAllocator::template rebind<char>::other> STRING_CONST;
```

dirk-thomas:
> Do you recall the reason why `rosidl` wraps constants in an enum?

This is how it is in ROS 1.

dejanpan:
[...]

PR #274, opened 2018-05-03
Generate structures with MISRA-compliant constants.

mjcarroll:
Because the underlying type of an enum isn't defined (it should
generally be the width of an int), it makes comparisons with the
constants differ in sign, which is against MISRA rules.

Connects to #271

### DIFF
--- a/rosidl_generator_cpp/resource/msg__struct.hpp.em
+++ b/rosidl_generator_cpp/resource/msg__struct.hpp.em
@@ -215,13 +215,7 @@ def generate_zero_string(membset, fill_args):
 
   // constants
 @[for constant in spec.constants]@
-@[  if (constant.type in ['byte', 'int8', 'int16', 'int32', 'int64', 'char'])]@
-  enum { @(constant.name) = @(int(constant.value)) };
-@[  elif (constant.type in ['uint8', 'uint16', 'uint32', 'uint64'])]@
-  enum { @(constant.name) = @(int(constant.value))u };
-@[  else]@
   static const @(MSG_TYPE_TO_CPP[constant.type]) @(constant.name);
-@[  end if]@
 @[end for]@
 
   // pointer types
@@ -286,17 +280,17 @@ using @(spec.base_type.type) =
 
 // constants requiring out of line definition
 @[for c in spec.constants]@
-@[  if c.type not in ['byte', 'int8', 'int16', 'int32', 'int64', 'char', 'uint8', 'uint16', 'uint32', 'uint64']]@
 template<typename ContainerAllocator>
 const @(MSG_TYPE_TO_CPP[c.type])
 @(spec.base_type.type)_<ContainerAllocator>::@(c.name) =
-@[    if c.type == 'string']@
+@[  if c.type == 'string']@
   "@(escape_string(c.value))";
-@[    elif c.type == 'bool']@
+@[  elif (c.type in ['bool', 'byte', 'int8', 'int16', 'int32', 'int64', 'char'])]@
   @(int(c.value));
-@[    else]@
+@[  elif (c.type in ['uint8', 'uint16', 'uint32', 'uint64'])]@
+  @(int(c.value))u;
+@[  else]@
   @(c.value);
-@[    end if]@
 @[  end if]@
 @[end for]@
 
--- a/rosidl_generator_cpp/resource/msg__struct.hpp.em
+++ b/rosidl_generator_cpp/resource/msg__struct.hpp.em
@@ -213,7 +213,7 @@ def generate_zero_string(membset, fill_args):
   }
 @[end for]@
 
-  // constants
+  // constant declarations
 @[for constant in spec.constants]@
   static const @(MSG_TYPE_TO_CPP[constant.type]) @(constant.name);
 @[end for]@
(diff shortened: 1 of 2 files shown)


rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -286,17 +280,17 @@ using @(spec.base_type.type) =
...
 @[for c in spec.constants]@
-@[  if c.type not in ['byte', 'int8', 'int16', 'int32', 'int64', 'char', 'uint8', 'uint16', 'uint32', 'uint64']]@
 template<typename ContainerAllocator>
 const @(MSG_TYPE_TO_CPP[c.type])
 @(spec.base_type.type)_<ContainerAllocator>::@(c.name) =
-@[    if c.type == 'string']@
+@[  if c.type == 'string']@
   "@(escape_string(c.value))";
-@[    elif c.type == 'bool']@
+@[  elif (c.type in ['bool', 'byte', 'int8', 'int16', 'int32', 'int64', 'char'])]@
```
dirk-thomas: Nitpicks: Python conditions shouldn't be wrapped in parenthesis.

Using a tuple rather than a list is preferred when the content isn't modified. Same below.

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -215,13 +215,7 @@ def generate_zero_string(membset, fill_args):
 
   // constants
```
dirk-thomas: Maybe update this comment to:

> constant declarations

mikaelarguedas:
This addresses only C++. Are you planning to do the same for the C messages ?
If yes will it be in this PR ?or a different one ?

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -286,17 +280,17 @@ using @(spec.base_type.type) =
 
 // constants requiring out of line definition
```
dirk-thomas: And this one to:

> constant definitions

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -286,17 +280,17 @@ using @(spec.base_type.type) =
...
 @[for c in spec.constants]@
-@[  if c.type not in ['byte', 'int8', 'int16', 'int32', 'int64', 'char', 'uint8', 'uint16', 'uint32', 'uint64']]@
 template<typename ContainerAllocator>
```
dirk-thomas: Does the container allocator make sense for the numeric types?

mjcarroll:
@mikaelarguedas I didn't think to check `C`, as it wasn't reported.  I don't know if the MISRA C++ standards catch something that the C standards do not, but it would appear that it should be affected by the same issue.  I will include it in this PR.

rosidl_generator_cpp/resource/msg__struct.hpp.em
mjcarroll: I think that this is necessary, the allocator doesn't apply to the member, but the containing object.  If we used `constexpr`, then we could declare and define in a single statement, but otherwise they need to be separate.

mjcarroll:
Started by user Michael Carroll
Building on master in workspace /var/lib/jenkins/jobs/ci_launcher/workspace
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4661/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1550/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3853/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4735/)
Triggering a new build of ci_linux
Triggering a new build of ci_linux-aarch64
Triggering a new build of ci_osx
Triggering a new build of ci_windows
Finished: SUCCESS

mjcarroll:
Opting to leave this at just the C++ constants, will do the C ones in a separate PR, as it has more far-ranging impact, and I'm not sure that it's necessary yet.

serge-nikulin:
What the fate of this issue for C++? Did it get into Bouncy?

mjcarroll:
@serge-nikulin Working on it, found a Windows bug that I'm trying to track down.

mjcarroll:
Windows again:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4791/)

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4731/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1599/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3920/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4791/)

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -284,20 +289,16 @@ def generate_zero_string(membset, fill_args):
...
-@[    if c.type == 'string']@
-  "@(escape_string(c.value))";
-@[    elif c.type == 'bool']@
-  @(int(c.value));
-@[    else]@
-  @(c.value);
-@[    end if]@
-@[  end if]@
+@(spec.base_type.type)_<ContainerAllocator>::@(c.name) = "@(escape_string(c.value))";
+@[ else ]
```
tfoote: Is the unescaped newline here on purpose?

PR #275, opened 2018-05-08
account for null-terminator character

mikaelarguedas:
similar to https://github.com/ros2/system_tests/pull/269

Fixes the following warnings: https://ci.ros2.org/job/ci_packaging_linux/92/warnings23Result/package.464043688/

related to ros2/ros2#481

### DIFF
--- a/rosidl_generator_cpp/test/test_array_generator.hpp
+++ b/rosidl_generator_cpp/test/test_array_generator.hpp
@@ -123,7 +123,7 @@ void test_vector_fill(
   int min, int max,
   int minlength, const int maxlength)
 {
-  if (size > 0 && min != max && minlength != maxlength) {
+  if (size > 0 && min < max && minlength < maxlength) {
     size_t step = (max - min) / size;
     size_t step_length = (maxlength - minlength) / size;
     char * tmpstr = reinterpret_cast<char *>(malloc(maxlength + 1));
--- a/rosidl_generator_cpp/test/test_array_generator.hpp
+++ b/rosidl_generator_cpp/test/test_array_generator.hpp
@@ -126,7 +126,7 @@ void test_vector_fill(
   if (size > 0 && min < max && minlength < maxlength) {
     size_t step = (max - min) / size;
     size_t step_length = (maxlength - minlength) / size;
-    char * tmpstr = reinterpret_cast<char *>(malloc(maxlength + 1));
+    char * tmpstr = static_cast<char *>(malloc(maxlength + 1));
     std::snprintf(tmpstr, minlength + 1, "%*d", minlength, min);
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The doc for [snprintf](http://www.cplusplus.com/reference/cstdio/snprintf/) says:

> Maximum number of bytes to be used in the buffer.
The generated string has a length of at most n-1, leaving space for the additional terminating null character.

So I don't think this patch is correct.

mikaelarguedas:
If you have a suggestion on how to properly handle, I'll happily merge it into this branch.
I was looking for the minimal change to get this to compile without warning. I believe that this PR achieves that goal without making the behavior more wrong than before

dirk-thomas:
Since we do want *n* characters to be written (and not *n-1*) the patch is changing (and in my opinion breaking) the behavior.

If we do rely on truncating the data then the code should explicitly check the return value (in which case I would expect the warning to disappear).

mikaelarguedas:
Thanks @dirk-thomas for the explanation. I modified the patch to actually write all characters provided by the user / test and not *n - 1*

mikaelarguedas:
And CI on our regular jobs/platforms to make sure this doesnt break anything:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4562/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1472/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3749/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4646/)

rosidl_generator_cpp/test/test_array_generator.hpp
```diff
@@ -123,19 +123,19 @@ void test_vector_fill(
   int min, int max,
   int minlength, const int maxlength)
 {
-  if (size > 0 && min != max && minlength != maxlength) {
+  if (size > 0 && min < max && minlength < maxlength) {
     size_t step = (max - min) / size;
     size_t step_length = (maxlength - minlength) / size;
-    char * tmpstr = reinterpret_cast<char *>(malloc(maxlength));
-    std::snprintf(tmpstr, minlength, "%*d", minlength, min);
+    char * tmpstr = reinterpret_cast<char *>(malloc(maxlength + 1));
```
dirk-thomas: Unrelated to the patch: this doesn't need a `reinterpret_cast` - a `static_cast` would be sufficient and "better" here.

mikaelarguedas: :+1: https://github.com/ros2/rosidl/pull/275/commits/fee32650c7346022c4b83ee26a43ed2457eb5839

PR #277, opened 2018-05-15
Generate imports for assert only in debug mode

martins-mozeiko:
Reduce amount of imports are executed at runtime by doing them only when python runs in debug mode (`PYTHONOPTIMIZE=0`). If python runs in optimized mode, assert statements are not executed, but currently there are a lot of imports happening. This patch moves imports inside `if __debug__` condition.

This significantly reduces amount of import statements executed at runtime. We have seen significant performance gains on Raspberry Pi 3 when messages contain arrays - because accessing each array element would trigger many imports.


mikaelarguedas:
Thanks @martins-mozeiko for the patch!

PR #281, opened 2018-06-11
generate __eq__ for Python messages

dirk-thomas:
Treating Python message classes as *data* (same as in C++) and implementing `__eq__` ~~and `__hash__`~~ based on all the fields.

CI for Linux only: [Build Status](https://ci.ros2.org/job/ci_linux/4585/)

### DIFF
--- a/rosidl_generator_py/resource/_msg.py.em
+++ b/rosidl_generator_py/resource/_msg.py.em
@@ -173,13 +173,6 @@ class @(spec.base_type.type)(metaclass=Metaclass):
             return False
 @[end for]@
         return True
-
-    def __hash__(self):
-        return hash((
-@[for field in spec.fields]@
-            self.@(field.name),
-@[end for]@
-        ))
 @[for field in spec.fields]@
 
     @@property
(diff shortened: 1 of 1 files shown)


dhood:
can you clarify the context/motivation for this?

dirk-thomas:
Without two message instances can't easily be compared if they contain the same values. The user would need to compare each and every field (as well as recursive fields) manually. Also if the message later changes and a field is added the existing checks would be incorrect.

tfoote:
I found this article with a good overview: https://hynek.me/articles/hashes-and-equality/

dirk-thomas:
Should we remove the `__hash__` method based on that?

tfoote:
From: https://docs.python.org/3.5/reference/datamodel.html#object.__hash__

> If a class defines mutable objects and implements an __eq__() method, it should not implement __hash__(), since the implementation of hashable collections requires that a key’s hash value is immutable (if the object’s hash value changes, it will be in the wrong hash bucket).

I think it's right to implement `__eq__` but drop the `__hash__` implementation.

dirk-thomas:
> drop the `__hash__` implementation.

Done in 2756c3f4dd4e00ff7e6f9f5d8b1726c491b677b7.

dhood:
FWICT ros1 time msg types are hashable: https://github.com/ros/genpy/blob/b244f2f50813ce0385c8673a899bce9e140a35d8/src/genpy/rostime.py#L127..L131 which might cause migration issues to ros2 for some code (unless times in ros2 are immutable? I don't think so). Not much to change, just an observation that came up from the hashable discussion which might be worth adding to the migration guide?

PR #282, opened 2018-06-14
Use the ament registry to obtain all the language generators automatically

esteve:
Connects to https://github.com/ros2/rosidl_typesupport/pull/24

### DIFF
--- a/rosidl_generator_c/CMakeLists.txt
+++ b/rosidl_generator_c/CMakeLists.txt
@@ -152,3 +152,7 @@ install(
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin
 )
+
+ament_index_register_resource(
+  "rosidl_interface_generator" CONTENT "${PROJECT_NAME}"
+)
--- a/rosidl_generator_c/package.xml
+++ b/rosidl_generator_c/package.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
-<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
-<package format="2">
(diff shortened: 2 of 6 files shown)


dirk-thomas:
This seems to be similar to #283. I think the other PR "better" since it covers both generators as well as runtime packages.

esteve:
No worries, closing this in favor of yours. Please update https://answers.ros.org/question/291808/how-to-hook-a-new-idl-generator/ when you have a moment so it's no longer inaccurate.

dirk-thomas:
Since you have a use case at hand maybe you can give the PRs a shot and confirm it works. It would be great to get those in for Bouncy (even though freeze is supposed to be tomorrow :worried:).

esteve:
Sure, I can give it a try. I tried mine locally and they work, and since yours cover more cases, I believe they should work as well. Let me spin up a build locally with your branches and see how it goes.

esteve:
In any case, it'd be fine to skip this and leave it for Bouncy+1, the use case is rather specific and won't affect most of the users (not many people use ros2_java... yet :wink: )

dirk-thomas:
> Let me spin up a build locally with your branches and see how it goes.

It looks like there is a problem with CI atm: https://ci.ros2.org/job/ci_linux/4659/ Will look into it and put them back in review when CI passes cleanly.

esteve:
I couldn't get past `example_interfaces` with your branches, I don't really know where the error is. I'm restoring this branch and reopening the PR and https://github.com/ros2/rosidl_typesupport/pull/24 in case you want to give it a look.

dirk-thomas:
I fixed the problem in [this](https://github.com/ros2/rosidl/pull/283/commits/880eb608dc97d8b07ac65f6e5170a395fde3ac89) and [this](https://github.com/ros2/rosidl_typesupport/pull/25/commits/995e6927748f677502dbdd955b6a1fbc733f106f) commit. A new round of CI builds are in flight.

It would be great if you could give it another try.

PR #283, opened 2018-06-14
add groups for generator and runtime packages

dirk-thomas:
The rosidl generator | runtime packages declare being a member of a group as well as register themselves at the resource index.


dirk-thomas:
@esteve I will merge this change for now hoping it addresses your use case. Please comment after the fact so that we know it does (or doesn't :worried: and iterate on it).

PR #287, opened 2018-06-25
use pytest instead of nose

dirk-thomas:
The package doesn't declare a test dependency on `nose` and therefore the tests fail in the devel jobs: http://build.ros2.org/view/Bdev/job/Bdev__rosidl__ubuntu_bionic_amd64/2/testReport/

Instead of adding that test dependency the patch switched the tests to use `pytest` instead and adds an explicit test dependency on it.

* Linux only: [Build Status](https://ci.ros2.org/job/ci_linux/4818/)

### DIFF
--- a/rosidl_parser/package.xml
+++ b/rosidl_parser/package.xml
@@ -12,6 +12,7 @@
   <test_depend>ament_cmake_pytest</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
+  <test_depend>pytest</test_depend>
 
   <export>
     <build_type>ament_cmake</build_type>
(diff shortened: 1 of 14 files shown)


rosidl_parser/package.xml
```diff
@@ -12,6 +12,7 @@
   <test_depend>ament_cmake_pytest</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
+  <test_depend>pytest</test_depend>
```
dirk-thomas: This rosdep key was wrong. Fixed in follow up commit c9e206ef47a65979a377d01140b70270042715cd.

PR #288, opened 2018-06-29
Error on all msg/srv names with underscores

dhood:
While discussing https://github.com/ros2/rmw_opensplice/issues/239 we noticed that, while msg/srv filenames with underscores don't work in general, an msg file that starts with `Sample_` can pass the [`is_valid_message_name` check](https://github.com/ros2/rosidl/blob/6549399b496ee394cda964cede24718e0a38af9c/rosidl_parser/rosidl_parser/__init__.py#L106) because of the `Sample_` prefix stripping that is in place for srv support for opensplice.

At the moment this PR just includes a test that highlights it.

### DIFF
--- a/rosidl_parser/test/test_valid_names.py
+++ b/rosidl_parser/test/test_valid_names.py
@@ -48,7 +48,7 @@ def test_is_valid_message_name():
             'Foo', 'FooBar']:
         assert is_valid_message_name(valid_message_name)
     for invalid_message_name in [
-            '0foo', '_Foo', 'Foo_', 'Foo_Bar']:
+            '0foo', '_Foo', 'Foo_', 'Foo_Bar', 'Sample_Foo']:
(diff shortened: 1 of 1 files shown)


dhood:
I poked around and determined that this isn't a simple thing to fix because the list of msg files is passed without any context of whether or not the files were user-written or output from a .srv file. I'm sure that the context could be added somehow, but I didn't think it was worth the time. Instead, I opened https://github.com/ros2/design/pull/192 to document that exceptions to the rule exist but that they are not intended for user-written messages.

dhood:
I am going to close this PR because I don't think the work to error on message files that pass because they get caught in the exceptions is something that will be looked at for the reasons above. It also may be that the exceptions can be removed in the future from discussion in https://github.com/ros2/design/pull/192#issuecomment-422993027

PR #290, opened 2018-07-10
fix using uninitialized CMake variables

dirk-thomas:
Otherwise being flagged when invoking with `--warn-uninitialized`.

### DIFF
--- a/rosidl_generator_c/CMakeLists.txt
+++ b/rosidl_generator_c/CMakeLists.txt
@@ -42,6 +42,12 @@ ament_index_register_resource("rosidl_generator_packages")
 
 ament_python_install_package(${PROJECT_NAME})
 
+if(BUILD_SHARED_LIBS)
+  set(${PROJECT_NAME}_LIBRARY_TYPE "SHARED")
+else()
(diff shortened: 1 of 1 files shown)


rosidl_generator_c/CMakeLists.txt
```diff
@@ -83,6 +83,11 @@ if(BUILD_TESTING)
     "msg/Wire.msg"
   )
 
+  if(BUILD_SHARED_LIBS)
+    set(${PROJECT_NAME}_LIBRARY_TYPE "SHARED")
+  else()
+    set(${PROJECT_NAME}_LIBRARY_TYPE "STATIC")
+  endif()
```
mikaelarguedas: Would moving the block from line 134 before `if(BUILD_TESTING)` fix the issue and avoid duplication of the variable assignment ?

dirk-thomas: Absolutely, I just didn't check "that far down" in the CMake file: 7ba0b8c8960af718a6d2232fa8889d7e2e5c5d0e

PR #291, opened 2018-07-11
Change msg__traits.hpp.em to support returning message datatype

sagniknitr:
PR for issue #258 . Kindly review and let me know if any further changes required.

PR #292, opened 2018-07-16
fix comparison

dirk-thomas:
Follow up of #291. @sagniknitr FYI.

Fixes the comparison of the string in the test: https://ci.ros2.org/view/nightly/job/nightly_win_deb/943/testReport/junit/rosidl_generator_cpp/Test_rosidl_generator_traits/check_data_type/

After: https://ci.ros2.org/job/ci_windows/4932/testReport/rosidl_generator_cpp/Test_rosidl_generator_traits/check_data_type/

### DIFF
--- a/rosidl_generator_cpp/test/test_msg_datatype.cpp
+++ b/rosidl_generator_cpp/test/test_msg_datatype.cpp
@@ -18,8 +18,10 @@
 
 
 TEST(Test_rosidl_generator_traits, check_data_type) {
-  ASSERT_EQ(rosidl_generator_traits::data_type<rosidl_generator_cpp::msg::String>(),
-    "rosidl_generator_cpp::msg::String");
-  ASSERT_EQ(rosidl_generator_traits::data_type<rosidl_generator_cpp::msg::Empty>(),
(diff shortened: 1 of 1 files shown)


rosidl_generator_cpp/test/test_msg_datatype.cpp
```diff
@@ -18,8 +18,8 @@
 
 
 TEST(Test_rosidl_generator_traits, check_data_type) {
-  ASSERT_EQ(rosidl_generator_traits::data_type<rosidl_generator_cpp::msg::String>(),
+  ASSERT_STREQ(rosidl_generator_traits::data_type<rosidl_generator_cpp::msg::String>(),
```
mikaelarguedas: Can you please also swap the arguments (string litteral first) as the "expected" result should be the first one and "actual" the second one

dirk-thomas: Swapped: 90749e68ca7b3759c081dcbe7660358209273b3e

sagniknitr:
Thanks for the information.I forgot that ASSERT_EQ() compares pointers.Will take care from next time.

PR #294, opened 2018-08-15
relax the regex splitting service files

dirk-thomas:
The change allows the service files to contain Windows newlines (`\r\n`) as well as service files which EOF immediately after the separator `---`.


mikaelarguedas:
> I did find another edge case which is if there's no divider the request and response are both the full content instead of an error.

Yeah I faced that edge case before (https://github.com/ros-planning/navigation_msgs/pull/5) and it was unclear to me in what way we should fail for this

dirk-thomas:
> I did find another edge case which is if there's no divider ...

A `.srv` file withoout a `---` line isn't valid and should be catched by the parser.

PR #300, opened 2018-10-08
Fix a double free problem in the generated C code.

clalancette:
When the generated C code is doing an array initialization,
the initialization of one of the array members may fail.  If
that happens with the current code, then it will call "__destroy"
on each of the already-initialized members.  In turn, the
"__destroy" method will uninitialize each of the members of
the structure, then free the whole structure.  The final
freeing is incorrect; the freeing of the memory should be
done when the whole array is freed, not on a member-by-member
basis.  The solution here is to call the "__fini" method during
a failure, which will uninitialize members, but not free the
entire structure.  This should get rid of double-free warnings
that come from clang static analysis.

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5273/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2029/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4389/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5252/)

### DIFF
--- a/rosidl_generator_c/resource/msg__functions.c.em
+++ b/rosidl_generator_c/resource/msg__functions.c.em
@@ -82,7 +82,7 @@ for field in spec.fields:
         if field.type.is_primitive_type():
             if field.type.type == 'string':
                 lines.append('if (!rosidl_generator_c__String__init(&msg->%s)) {' % field.name)
-                lines.append('  %s__destroy(msg);' % msg_typename)
+                lines.append('  %s__fini(msg);' % msg_typename)
                 lines.append('  return false;')
                 lines.append('}')
                 if field.default_value is not None:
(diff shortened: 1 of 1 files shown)


clalancette:
> Makes sense, but a test would be nice, since just reading the diff is hard (since the "second free" isn't near the diffs if I understand correctly. On the other hand a test might be really hard to implement since we're not passing an allocator to our message constructors.

Yeah, I'm honestly not sure how to do a test, other than writing some "pattern matching" to look for the right sequence in the generated code.  With your approval, I'm going to merge this for now.

allenh1:
@clancette maybe the next incrementation of rosidl generator should use an rcutils allocator? That way, a test could be made with an allocator that just always fails and we could then test it?

I've also thought it could be a good idea to add `-fsanitize=address` to the compiler options for this package during CI.

clalancette:
> @clalancette maybe the next incrementation of rosidl generator should use an rcutils allocator? That way, a test could be made with an allocator that just always fails and we could then test it?

Oh, interesting idea, not exactly sure how to make that work.  I'll open an issue about it, though, so we can investigate in the future.
> 
> I've also thought it could be a good idea to add `-fsanitize=address` to the compiler options for this package during CI.

Yeah, I've been wanting to add some static analysis jobs to CI for a while, I haven't (yet) had time for it.  We definitely want a separate job for `-fsanitize=XXX`, since it has a non-trivial impact on performance.

PR #307, opened 2018-10-23
Add static cast when assigning values to arrays or vectors

jacobperron:
This resolves bugs due to implicit casts.

Specifically, resolves ros2/rcl_interfaces#42

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5328/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2064/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4427/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5295/)

### DIFF
--- a/rosidl_generator_cpp/resource/msg__struct.hpp.em
+++ b/rosidl_generator_cpp/resource/msg__struct.hpp.em
@@ -100,10 +100,10 @@ def generate_default_string(membset):
                 strlist.append('this->%s.resize(%d);' % (member.name, member.num_prealloc))
             if isinstance(member.default_value, list):
                 if all(v == member.default_value[0] for v in member.default_value):
-                    strlist.append('std::fill(this->%s.begin(), this->%s.end(), %s);' % (member.name, member.name, member.default_value[0]))
+                    strlist.append('std::fill(this->%s.begin(), this->%s.end(), static_cast<typename decltype(this->%s)::value_type>(%s));' % (member.name, member.name, member.name, member.default_value[0]))
                 else:
                     for index, val in enumerate(member.default_value):
-                        strlist.append('this->%s[%d] = %s;' % (member.name, index, val))
+                        strlist.append('this->%s[%d] = static_cast<typename decltype(this->%s)::value_type>(%s);' % (member.name, index, member.name, val))
             else:
                 strlist.append('this->%s = %s;' % (member.name, member.default_value))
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Instead of using a cast couldn't the generated code just use the appropriate type there?

sloretz:
> Instead of using a cast couldn't the generated code just use the appropriate type there?

@dirk-thomas IIUC the problem is the default value is `0` which defaults to `const int`. The template for `std::fill` on windows picks that up and complains that an array of `uint8_t` is being filled with the type `const int`. 
The cast just tells the compiler what type to use for the constant for std::fill.

Are you saying fill should be called with the types instead of the cast?

```c++
std::fill<std::array<uint8_t>::iterator, std::array<uint8_t>::value_type>(member.begin(), member.end(), 0)
```
```

**Comment by dirk-thomas** [comment]
Yes, kind of. Since we do know the type at generation time of that file I would lean towards using that instead of `typename decltype(this->%s)::value_type`. Just a thought so - not a requirement.

**Comment by jacobperron** [comment]
> Instead of using a cast couldn't the generated code just use the appropriate type there?

@dirk-thomas @sloretz 
I'm not sure I follow. Are we talking about changing the generator code to be explicit about the `std::fill` template? If so, the line will look like

```
std::fill<typename decltype(this->%s)::iterator, typename decltype(this->%s)::value_type>(this->%s.begin(), this->%s.end(), %s);
```

vs.

```
std::fill(this->%s.begin(), this->%s.end(), static_cast<typename decltype(this->%s)::value_type>(%s));
```

**Comment by sloretz** [comment]
> Are we talking about changing the generator code to be explicit about the std::fill template?

I think so.

If `member.type` is available then the `decltype()` might not be necessary

```python
strlist.append(('std::fill<%s::iterator, %s::value_type>(this->%s.begin(), this->%s.end(), %s);' % (member.type, member.type, member.name, member.name, member.default_value[0]))
```

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -100,7 +100,10 @@ def generate_default_string(membset):
                 strlist.append('this->%s.resize(%d);' % (member.name, member.num_prealloc))
             if isinstance(member.default_value, list):
                 if all(v == member.default_value[0] for v in member.default_value):
-                    strlist.append('std::fill(this->%s.begin(), this->%s.end(), %s);' % (member.name, member.name, member.default_value[0]))
+                    # Specifying type for std::fill because of MSBuild warning about cast (C4244)
```
clalancette: I think this comment (and the one below) should probably be more specific; that is, it should specify which version of MSVC we are having problems with, and it should specify that it is only a problem for types that are *smaller* than an int.  That way we don't lose information that we've already discovered :).

jacobperron: Ok. Am I correct that we are using MSVC 2017?

jacobperron: nvm, I think I got it from the log: **MSVC 14.12**

jacobperron:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5349/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2079/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4442/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5312/)

PR #310, opened 2018-10-31
Generate messages and services from actions

sloretz:
This is part of the cmake plumbing to generate code for actions.  It converts `.action` files passed to `rosidl_generate_interfaces()` to `.msg` and `.srv`, then passes those through the normal generation pipeline with the other `.msg` and `.srv` passed.

It does not generate typesupport files for the `.action` themselves. A proposal for those is hardcoded in ros2/rcl_interfaces#47.

Requires ros2/rosidl#311

connects to ros2/rcl_interfaces#47

Issue #303, opened 2018-10-18
Generate C++ typesupport for actions


sloretz:
## Feature request

There should be a c++ file importable by `#include <package_name/action/actionname.hpp>` which contains everything needed to use an action type in c++.

#### Feature description

It needs to be possible to generate a header file that describes an action, and has all the information needed to create topics and services. Since services already are an example of a typesupport that needs to include messages, actions should use the same strategy.

##### Output files

Services have a top level header which includes two files: `traits` and `struct`. Traits indicates whether either the request or response is bounded or unbounded in size. I don't think there needs to be a traits file for the whole action.
https://github.com/ros2/rosidl/blob/206dd2bbea68ece80d5e811236cb4aecda495d6c/rosidl_generator_cpp/resource/srv.hpp.em#L22-L23

The struct file includes the request and response message, as well as defines a struct with a `using` directive that indicates which message is which

https://github.com/ros2/rosidl/blob/206dd2bbea68ece80d5e811236cb4aecda495d6c/rosidl_generator_cpp/resource/srv__struct.hpp.em#L31-L35

The action struct template could look something like this

```c++

#include "@(spec.pkg_name)/action/@(get_header_filename_from_msg_name(spec.action_name))__sendgoal.hpp"
#include "@(spec.pkg_name)/action/@(get_header_filename_from_msg_name(spec.action_name))__get_result.hpp"
#include "@(spec.pkg_name)/action/@(get_header_filename_from_msg_name(spec.action_name))__feedback.hpp"
// Convenience include for common service and message definition
#include <rcl_interfaces/srv/cancel_action.hpp>
#include <rcl_interfaces/srv/goal_status.hpp>

 struct @(spec.action_name) 
 { 
   using SendGoalService = @(spec.pkg_name)::action::@(spec.action_name)_SendGoal; 
   using GetResultService = @(spec.pkg_name)::action::@(spec.action_name)_GetResult; 
   using FeedbacMessage = @(spec.pkg_name)::action::@(spec.action_name)_Feedback; 
};
```

##### CMake targets

The CMake target needs to be aware of the generated output files so that it can rebuild stuff downstream when they change. Awareness of the files generated by `.action` needs to be included here

https://github.com/ros2/rosidl/blob/206dd2bbea68ece80d5e811236cb4aecda495d6c/rosidl_generator_cpp/cmake/rosidl_generator_cpp_generate_interfaces.cmake#L25-L36

connects to ros2/ros2#583

[...]

Issue #302, opened 2018-10-18
Generate C typesupport for Actions


sloretz:
## Feature request

There should be a C file importable as `#include <package_name/action/actionname.h>` for an action. That file should include the service and message definitions that make up an action.

#### Feature description

There should be one C file importable that describes the services and messages needed to use an action. These services and messages are discoverable based on the name of their types.

#### Implementation considerations
Services include message type supports, so actions should use a similar strategy.

Services have one top level file including a `request` and `response` typesupport for messages.
https://github.com/ros2/rosidl/blob/206dd2bbea68ece80d5e811236cb4aecda495d6c/rosidl_generator_c/resource/srv.h.em#L23-L24

Actions should have a top level file that includes the "Send Goal " service type support, "Get result" service type support, and "Feedback" message type support. It could also include a cancel goal service and status message from rcl_interfaces

TODO what is `ROSIDL_TYPESUPPORT_INTERFACE__SERVICE_SYMBOL_NAME` doing, and does there need to be a matching function for actions?
https://github.com/ros2/rosidl/blob/206dd2bbea68ece80d5e811236cb4aecda495d6c/rosidl_typesupport_interface/include/rosidl_typesupport_interface/macros.h#L29

blocked by ros2/rcl_interfaces#40
connects to ros2/ros2#583

sloretz:
Fixed by #312 and ros2/rosidl_typesupport#36

PR #311, opened 2018-10-31
Action file support

### LINKED ISSUE #305: Parse .action files
## Feature request
It needs to be possible to parse `.action` files into an intermediate representation.

#### Feature description
`rosidl_parser` needs methods `parse_action_file` and `parse_action_string` which read a `.action` file and generate an intermediate representation that can be used by downstream generators.

The intermediate representation should have service definitions for the `Send Goal` and `Get Result` services, as well as a message definition for the `Feedback` topic. 

#### Implementation considerations

`parse_service_string` Splits the service file by `---` and calls `parse_message_string` on the individual parts. The code for actions should create separate strings for the services and message and then call `parse_service_string` and `parse_message_string` to create a description of an action.

https://github.com/ros2/rosidl/blob/206dd2bbea68ece80d5e811236cb4aecda495d6c/rosidl_parser/rosidl_parser/__init__.py#L686-L705

connects to ros2/ros2#583

sservulo:
This PR closes #305. Added `.action` file parsing to rosidl_parser.
CMake preprocessing of .action files into  services and messages is coming to this branch from `sloretz/speficication-to-str`.


rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -705,3 +709,75 @@ def parse_service_string(pkg_name, srv_name, message_string):
...
+    basename = os.path.basename(interface_filename)
+    action_name = os.path.splitext(basename)[0]
+    with open(interface_filename, 'r') as h:
+        return parse_action_string(pkg_name, action_name, h.read())
+
+
+def parse_action_string(pkg_name, action_name, message_string):
+    lines = message_string.splitlines()
+    separator_indices = [
+        index for index, line in enumerate(lines) if line == SERVICE_REQUEST_RESPONSE_SEPARATOR]
```
hidmic: @sservulo consider having an `ACTION_GOAL_FEEDBACK_RESULT_SEPARATOR` constant for better readability.

rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -705,3 +709,75 @@ def parse_service_string(pkg_name, srv_name, message_string):
...
+    if not separator_indices:
+        raise InvalidServiceSpecification(
+            "Could not find separator '%s' between action goal, result and feedback services" %
+            SERVICE_REQUEST_RESPONSE_SEPARATOR)
+    if len(separator_indices) != 3:
+        raise InvalidServiceSpecification(
+            "Could not find unique separator '%s' between action goal, result and \
+            feedback services" %
+            SERVICE_REQUEST_RESPONSE_SEPARATOR)
```
hidmic: @sservulo nit: we could also check directly for the amount of separators (or sections) and complain about bad action definition. Also, I'd wouldn't phrase exception messages in terms of services, it's somewhat confusing.

rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -705,3 +709,75 @@ def parse_service_string(pkg_name, srv_name, message_string):
...
+    #----------------------------------------------------------------------------------------------
+
+    #----------------------------------------------------------------------------------------------
+    ## Feedback message
+    implicit_input = ["uint8[16] uuid"]
+    message_string = '\n'.join(
+        lines[separator_indices[1] + 1:] + implicit_output)
+    feedback_msg = parse_message_string(
+        pkg_name, action_name + ACTION_FEEDBACK_MESSAGE_SUFFIX, message_string)
+    #----------------------------------------------------------------------------------------------
```
hidmic: @sservulo Hmm, implicit fields may clash with explicitly given ones. Action generation in ROS1 would nest message definition as a field of a completely implicit message, see [genaction.py](docs.ros.org/melodic/api/actionlib_msgs/html/genaction_8py_source.html) sources. Alternatively, we could check for the presence of duplicate entries. @sloretz @jacobperron thoughts?

rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -705,3 +709,75 @@ def parse_service_string(pkg_name, srv_name, message_string):
...
     return ServiceSpecification(pkg_name, srv_name, request_message, response_message)
+
+def parse_action_file(pkg_name, interface_filename):
+    basename = os.path.basename(interface_filename)
+    action_name = os.path.splitext(basename)[0]
+    with open(interface_filename, 'r') as h:
+        return parse_action_string(pkg_name, action_name, h.read())
+
+
+def parse_action_string(pkg_name, action_name, message_string):
```
hidmic: @sservulo consider `message_string` -> `action_string`.

rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -705,3 +709,75 @@ def parse_service_string(pkg_name, srv_name, message_string):
...
+    lines = message_string.splitlines()
+    separator_indices = [
```
hidmic: @sservulo nit: another option would be to split the whole definition into sections separated by "---" lines and then work with those. You wouldn't have to play with slices then.

rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -705,3 +709,75 @@ def parse_service_string(pkg_name, srv_name, message_string):
...
+        response_message_string)
+
+    services.append(ServiceSpecification(pkg_name, action_name, request_message, response_message))
+    #----------------------------------------------------------------------------------------------
+
+    #----------------------------------------------------------------------------------------------
+    ## Feedback message
+    implicit_input = ["uint8[16] uuid"]
+    message_string = '\n'.join(
+        lines[separator_indices[1] + 1:] + implicit_output)
```
hidmic: @sservulo implicit_*input* maybe?

rosidl_parser/rosidl_parser/__init__.py
sloretz: That's a good point. +1 to nesting the user defined message

rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -705,3 +709,75 @@ def parse_service_string(pkg_name, srv_name, message_string):
...
+
+def parse_action_string(pkg_name, action_name, message_string):
+    lines = message_string.splitlines()
+    separator_indices = [
+        index for index, line in enumerate(lines) if line == SERVICE_REQUEST_RESPONSE_SEPARATOR]
+    if not separator_indices:
+        raise InvalidServiceSpecification(
+            "Could not find separator '%s' between action goal, result and feedback services" %
+            SERVICE_REQUEST_RESPONSE_SEPARATOR)
+    if len(separator_indices) != 3:
```
sloretz: 3 -> 2

rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -705,3 +709,75 @@ def parse_service_string(pkg_name, srv_name, message_string):
...
+        request_message_string)
+
+    implicit_output = ["bool accepted", "builtin_interfaces/Time stamp"]
+    response_message_string = '\n'.join(implicit_output)
+    response_message = parse_message_string(
+        pkg_name,
+        action_name + ACTION_GOAL_MESSAGE_SUFFIX + SERVICE_RESPONSE_MESSAGE_SUFFIX,
+        response_message_string)
+
+    services.append(ServiceSpecification(pkg_name, action_name, request_message, response_message))
```
sloretz: service name `action_name + ACTION_GOAL_MESSAGE_SUFFIX`

rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -705,3 +709,75 @@ def parse_service_string(pkg_name, srv_name, message_string):
...
+
+    implicit_output = ["int8 status"]
+    response_message_string = '\n'.join(
+        lines[separator_indices[0] + 1 : separator_indices[1]] + implicit_output)
+    response_message = parse_message_string(
+        pkg_name,
+        action_name + ACTION_RESULT_MESSAGE_SUFFIX + SERVICE_RESPONSE_MESSAGE_SUFFIX,
+        response_message_string)
+
+    services.append(ServiceSpecification(pkg_name, action_name, request_message, response_message))
```
sloretz: Service name should be `action_name + ACTION_RESULT_MESSAGE_SUFFIX`

rosidl_parser/rosidl_parser/__init__.py
sservulo: Other issues were already adressed. @sloretz, this one will probably delay a bit cmake header typesupport on #302 and #303 due to the need to update location assumptions of nested msgs.

sservulo:
Addressed review comments.

rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -717,25 +718,20 @@ def parse_action_file(pkg_name, interface_filename):
...
-    separator_indices = [
-        index for index, line in enumerate(lines) if line == SERVICE_REQUEST_RESPONSE_SEPARATOR]
-    if not separator_indices:
-        raise InvalidServiceSpecification(
-            "Could not find separator '%s' between action goal, result and feedback services" %
-            SERVICE_REQUEST_RESPONSE_SEPARATOR)
-    if len(separator_indices) != 3:
+def parse_action_string(pkg_name, action_name, action_string):
+    action_blocks = action_string.split(ACTION_REQUEST_RESPONSE_SEPARATOR)
+    if len(action_blocks) != 2:
```
sloretz: Before it was counting the number of `---` in an action file which should be 2. Now it's counting the sections divided by `---` which is 3.

sservulo: @sloretz I know, it was fixed in `3b7a695`, latest commit.

sloretz: Ah oops, I forgot I clicked on a single commit.

rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -705,3 +722,76 @@ def parse_service_string(pkg_name, srv_name, message_string):
...
+    #----------------------------------------------------------------------------------------------
+    ## Send goal
+    implicit_input = ["uint8[16] uuid"]
+    request_message_string = goal_service_string + '\n' + '\n'.join(implicit_input)
+    request_message = parse_message_string(
+        pkg_name,
+        action_name + ACTION_GOAL_MESSAGE_SUFFIX + SERVICE_REQUEST_MESSAGE_SUFFIX,
+        request_message_string)
+
+    implicit_output = ["bool accepted", "builtin_interfaces/Time stamp"]
```
sloretz: @jacobperron @hidmic Will `rcl_actions` associate the service response by mapping goal ID to the request sequence number, or by having the higher level client library get the goal ID out of the message? If the former then these fields should be OK, if the latter then the stamp field should be `action_msgs/GoalInfo` so it includes a goal id too.

rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -705,3 +722,76 @@ def parse_service_string(pkg_name, srv_name, message_string):
...
+    #----------------------------------------------------------------------------------------------
+    ## Get result
+    implicit_input = ["uint8[16] uuid"]
+    request_message_string = '\n'.join(implicit_input)
+    request_message = parse_message_string(
+        pkg_name,
+        action_name + ACTION_RESULT_MESSAGE_SUFFIX + SERVICE_REQUEST_MESSAGE_SUFFIX,
+        request_message_string)
+
+    implicit_output = ["int8 status"]
```
sloretz: What's the status field for?

sloretz:
Besides a couple questions about the implicit fields, I think with a couple unit tests for parsing actions I think this could be merged, and then #310 could be re-targeted at master.

rosidl_parser/rosidl_parser/__init__.py
sservulo: It's the goal status as defined in the [proposal](https://github.com/ros2/design/blob/actions_proposal/articles/actions.md#get-result-service).

sloretz: Whoops, I completely forgot about that

sloretz: @jacobperron same question about `rcl_actions` using the sequence number vs goal id. If the latter then using [action_msgs/GoalStatus](https://github.com/ros2/rcl_interfaces/blob/master/action_msgs/msg/GoalStatus.msg) here would get that.

rosidl_parser/rosidl_parser/__init__.py
hidmic: I think that'd be currently the case (using goal IDs I mean).

rosidl_parser/rosidl_parser/__init__.py
hidmic: Same as above.

sservulo:
Added tests to Action parser and included some fixes in implicit field collision logic. Note that the Action messages suffixes were changed in `3c4df0a` to be compliant with message name regex.

hidmic:
@sloretz @jacobperron I'd say we get this one in to unblock the rest. If the current set of implicit fields is later unfit (which I'm inclined to think but we may find out that it's not), we can change them.

rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -431,6 +447,19 @@ def parse_message_string(pkg_name, msg_name, message_string):
                     line=line, pkg=pkg_name, msg=msg_name, err=err,
                 ))
                 raise
+
+            # check for field name collision with action implicit parameters
+            # (e.g. 2 or more occurrences of a field_name contained in ACTION_IMPLICIT_FIELDS,
+            # including the implicit one)
+            if field_name in ACTION_IMPLICIT_FIELDS and action_fields[field_name] >= 1:
```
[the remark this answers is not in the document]
sservulo: @jacobperron It will trigger only if a field named with the same name as one implicit field is in the spec and there are at least two of them (replacing a `ValueType` error in this case), to make sure it triggers only when there's duplication, so if you have a field named `uuid` used in your msg definition it will not trigger. Should I go with the invalid fields route, then?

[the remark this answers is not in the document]
jacobperron: Thanks for the clarification. I think it is fine to leave it as-is.

jacobperron:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5475/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2155/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4527/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5389/)

hidmic:
@sservulo this one's ready to go!

rosidl_parser/rosidl_parser/__init__.py
```diff
@@ -705,3 +739,77 @@ def parse_service_string(pkg_name, srv_name, message_string):
...
+
+def parse_action_file(pkg_name, interface_filename):
+    basename = os.path.basename(interface_filename)
+    action_name = os.path.splitext(basename)[0]
+    with open(interface_filename, 'r') as h:
+        return parse_action_string(pkg_name, action_name, h.read())
+
+
+def parse_action_string(pkg_name, action_name, action_string):
+    action_blocks = action_string.split(ACTION_REQUEST_RESPONSE_SEPARATOR)
```
dirk-thomas: This logic is incorrect. It e.g. even matches a `---` in a comment. Please use the same logic as for splitting the request and response in a service.

hidmic: You're right. I had not foreseen the possibility of a separator in comments. So it's either above's logic using slices or `re.split('^---$', a, flags=re.MULTILINE)`.

sservulo: Fixed.

PR #319, opened 2018-11-12
Revert "Registers IDL interfaces' subfolders with Ament."

mjcarroll:
This reverts commit 7d791728f36b99f08242aedd33aac05f95c64619.

This should address recent failures in the build farm with regards to `ros1_bridge`, where the dynamic bridge could not match messages due to the change in the way they were registered.

For example: https://ci.ros2.org/view/packaging/job/packaging_linux/1254/


mjcarroll:
Linux: https://ci.ros2.org/buildStatus/icon?job=ci_packaging_linux&build=143
Windows: https://ci.ros2.org/buildStatus/icon?job=ci_packaging_windows&build=81

hidmic:
@mjcarroll there's just one small issue with this: tests in the `system_tests/test_communication` package will run for the messages and services that underpin actions, which is not intended and not supported as it was stated [here](https://github.com/ros2/system_tests/pull/303#pullrequestreview-169578958).

hidmic:
I wonder how big's the gap to fill in order to make `ros1_bridge` pass again.

sloretz:
Just remembered ros2/ros2cli#153 would also need a PR if this is reverted. I imagine the inverse, updating ros1_bridge, would be close to that PR.

mjcarroll:
@dirk-thomas Do you have any objection to just fixing up ros1_bridge in this case?

dirk-thomas:
I don't mind how it gets fixed for now.

mjcarroll:
We opted to fix this in ros1_bridge: https://github.com/ros2/ros1_bridge/pull/143

PR #324, opened 2018-11-15
conversion functions from .msg / .srv to .idl

dirk-thomas:
This is the third PR integrating #298 step-by-step.

Builds on top of #323.

Since the functionality provided by this patch isn't actively being used yet there isn't a point in running CI.

The command line script `msg2idl.py` and `srv2idl.py` can be used to manually convert `.msg` / `.srv` files.

PR #325, opened 2018-11-15
add lark-based grammar

dirk-thomas:
This is the fourth PR integrating #298 step-by-step.

Builds on top of #324.

The grammar covers a subset of the IDL spec.

Linux CI: [Build Status](https://ci.ros2.org/job/ci_linux/5549/)

### DIFF
--- a/rosidl_parser/rosidl_parser/grammar.lark
+++ b/rosidl_parser/rosidl_parser/grammar.lark
@@ -40,11 +40,13 @@ hexadecimal_literal: "0x"i HEXDIGIT+
 character_literal: "'" CHAR "'"
 wide_character_literal: "L'" CHAR "'"
 
-CHAR: _CHARACTERS
+CHAR: CHAR_SPACE
+  | _CHARACTERS
   | DIGIT
   | _FORMATTING_CHARACTERS
   | _ESCAPE_SEQUENCES
   | _GRAPHIC_CHAR
+CHAR_SPACE: " "
 
 // Table 7-2: Characters
 _CHARACTERS: LETTER | /[ÃãÄäÅåÆæÇçÈèÉéÊêËëÌìÍíÎîÏïÑñÒòÓóÔôÕõÖöØøÙùÚúÛûÜüßÿ]/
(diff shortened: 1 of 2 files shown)


rosidl_parser/rosidl_parser/grammar.lark
```diff
@@ -0,0 +1,518 @@
...
+%import common.HEXDIGIT
+%import common.LETTER
+%import common.WS
+%ignore WS
+
+
+// 7.2.2 Comments
+COMMENT: "//" /[^\n]/*
+  | "/*" /(.|\n)+/ "*/"
+%ignore COMMENT
```
sloretz: Is ignoring comments intentional? I would have expected them to be kept after #323

rosidl_parser/rosidl_parser/grammar.lark
```diff
@@ -0,0 +1,518 @@
...
+
+// (48)
+struct_forward_dcl: "struct" IDENTIFIER
+
+// (49)
+union_dcl: union_def
+  | union_forward_dcl
+
+// (50)
+union_def: "union" IDENTIFIER "switch" "(" switch_type_spec ")" "{" switch_body "}"
```
sloretz: I assumed unions wouldn't be supported. Will the generators generate code for them?

dirk-thomas:
> any literal can be assigned to any type.

Correct, that kind of check is out of scope for the grammar.

rosidl_parser/rosidl_parser/grammar.lark
dirk-thomas: Yes, ignoring comments is intentional. Comments in `.idl` are not extracted. The comments from `.msg` files are bring transformed into annotations in IDL and those are being extracted.

rosidl_parser/rosidl_parser/grammar.lark
dirk-thomas: Correct assumption, no the pipeline doesn't support them.

I should check if I can remove those from the grammar instead.

sloretz: Roger. I think it can be in the grammar as long as something above warn's or errors when it sees a union.

dirk-thomas:
I disabled the `union_dcl` as well as the `native_dcl` which we both don't support atm: 868bace4b03bd5c638d2d51a946d465048697c4d

PR #326, opened 2018-11-15
add object representation and conversion from ast

dirk-thomas:
This is the fifth PR integrating #298 step-by-step.

Builds on top of #325. Needs ros2/ci#220.

After lark parses any input file the AST is converted to an object representation (which will later be passed into the message generator templates).

Linux CI: [Build Status](https://ci.ros2.org/job/ci_linux/5560/)


rosidl_parser/rosidl_parser/definition.py
```diff
@@ -0,0 +1,343 @@
...
+        return super().__eq__(other) and self.name == other.name
+
+
+class NamespacedType(BaseType):
+
+    __slots__ = ('namespaces', 'name')
+
+    def __init__(self, namespaces, name):
+        super().__init__()
+        self.namespaces = namespaces
```
sloretz: Nit, my first thought was that the plural name meant a type could have multiple fully qualified namespaces, but now I see that it's a list of the different names making up a single namespace. Would you mind adding a comment to this type to describe it?

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -0,0 +1,343 @@
...
+
+class Sequence(NestedType):
+
+    __slots__ = set()
+
+    def __init__(self, basetype):
+        super().__init__(basetype)
+
+
+class UnboundedSequence(Sequence):
```
sloretz: What's the difference between Sequence and UnboundedSequence?

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -0,0 +1,343 @@
...
+        super().__init__(basetype)
+
+
+class BoundedSequence(Sequence):
+
+    __slots__ = ('upper_bound', )
+
+    def __init__(self, basetype, upper_bound):
+        super().__init__(basetype)
+        assert isinstance(upper_bound, int)
```
sloretz: Converting to `self.upper_bound = int(upper_bound)` would raise an exception of the type isn't convertible even when `PYTHONOPTIMIZE` is set. Maybe that's not important though if this assert is calling out an assumption rather than enforcing a restriction.

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -0,0 +1,343 @@
...
+
+
+class Annotation:
+
+    __slots__ = ('name', 'value')
+
+    def __init__(self, name, value):
+        assert isinstance(name, str)
+        self.name = name
+        self.value = value
```
sloretz: Is this an annotation declaration or application? If the latter and an annotation has multiple parameters like `@range(min=10, max=20)`, what is stored in value?

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -0,0 +1,343 @@
...
+        self.request_message = request
+
+        assert isinstance(response, Message)
+        assert response.structure.type.namespaces == type_.namespaces
+        assert response.structure.type.name == type_.name + \
+            SERVICE_RESPONSE_MESSAGE_SUFFIX
+        self.response_message = response
+
+
+class IdlIdentifier:
```
sloretz: This doesn't appear to be used. What's it for?

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -0,0 +1,343 @@
...
+
+class AbstractType:
+
+    __slots__ = ()
+
+    def __eq__(self, other):
+        return type(self) == type(other)
+
+
+class BaseType(AbstractType):
```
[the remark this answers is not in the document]
dirk-thomas: `AbstractType` is the root type. All non-nested types inherit from `BaseType` but arrays and sequences do not.

rosidl_parser/rosidl_parser/definition.py
dirk-thomas: `Sequence` is the base class of `BoundedSequence` and `UnboundedSequence`.

rosidl_parser/rosidl_parser/definition.py
dirk-thomas: That would still allow to pass e.g. a string to the API which I would rather restrict to only accept `int`.

rosidl_parser/rosidl_parser/definition.py
dirk-thomas: The value is annotation specific. For `range` it would be a dictionary: `{'min': 10, 'max': 20}`

rosidl_parser/rosidl_parser/definition.py
dirk-thomas: Indeed, this is now represented by two separate variables - the package name and the relative path. Removed in f644436528c3469af04e4f325b9b6b65892de6c2.

rosidl_parser/rosidl_parser/definition.py
sloretz: Mind adding a comment describing value in the code?

rosidl_parser/rosidl_parser/definition.py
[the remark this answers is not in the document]
sloretz: Why don't arrays and sequences inherit from `BaseType`? Is it because there is a place where arrays and sequences are forbidden, but all other types are allowed?

rosidl_parser/test/test_parser.py
```diff
@@ -0,0 +1,174 @@
...
+    assert len(constants) == 5
+
+    constant = [c for c in constants if c.name == 'SHORT_CONSTANT']
+    assert len(constant) == 1
+    constant = constant[0]
+    assert isinstance(constant.type, BasicType)
+    assert constant.type.type == 'int16'
+    assert constant.value == -23
+
+    # assert 'UNSIGNED_LONG_CONSTANT' in msg.constants
```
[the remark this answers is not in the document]
dirk-thomas: Good point. Reenabled all of them in 791dd7ac7a73cb02c59468ad0bd2e6c3b39fb215.

rosidl_parser/rosidl_parser/definition.py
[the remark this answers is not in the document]
dirk-thomas: > Is it because there is a place where arrays and sequences are forbidden ...

Yes, an array or sequence con only contain `BaseType` elements.

rosidl_parser/rosidl_parser/definition.py
dirk-thomas: Added docblock in 7647c6b042bd4c0d4135baf513c34948ec82e729.

rosidl_parser/rosidl_parser/definition.py
dirk-thomas: Added docblock in 7647c6b042bd4c0d4135baf513c34948ec82e729.

rosidl_parser/rosidl_parser/parser.py
```diff
@@ -0,0 +1,488 @@
...
+from rosidl_parser.definition import SERVICE_RESPONSE_MESSAGE_SUFFIX
+from rosidl_parser.definition import String
+from rosidl_parser.definition import Structure
+from rosidl_parser.definition import UnboundedSequence
+from rosidl_parser.definition import WString
+
+with open(os.path.join(os.path.dirname(__file__), 'grammar.lark'), 'r') as h:
+    grammar = h.read()
+
+parser = Lark(grammar, start='specification')
```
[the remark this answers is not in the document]
dirk-thomas: Afaik the import is relative to the `grammar.lark` file.

rosidl_parser/rosidl_parser/parser.py
```diff
@@ -0,0 +1,488 @@
...
+            os.makedirs(os.path.dirname(png_file), exist_ok=True)
+            pydot__tree_to_png(tree, png_file)
+
+    return extract_content_from_ast(tree)
+
+
+def extract_content_from_ast(tree):
+    content = IdlContent()
+
+    include_directives = tree.find_data('include_directive')
```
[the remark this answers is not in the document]
dirk-thomas: Correct. This is not checked anywhere explicitly atm.

rosidl_parser/rosidl_parser/parser.py
```diff
@@ -0,0 +1,488 @@
...
+    for include_directive in include_directives:
+        assert len(include_directive.children) == 1
+        child = include_directive.children[0]
+        assert child.data in ('h_char_sequence', 'q_char_sequence')
+        include_token = next(child.scan_values(_find_tokens(None)))
+        content.elements.append(Include(include_token.value))
+
+    const_dcls = tree.find_data('const_dcl')
```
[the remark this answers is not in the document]
dirk-thomas: Correct. Another simplification atm.

dirk-thomas:
Just another CI after the changes: <a href='https://ci.ros2.org/job/ci_linux/5572/'></a>

rosidl_parser/rosidl_parser/parser.py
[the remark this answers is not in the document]
sloretz: It looks like `Lark` is given the content of `grammar.lark`. How does it know the path to the file?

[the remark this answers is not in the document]
dirk-thomas: Maybe it is not using the file in this patch but the one bundled in the `lark-parser` package? I will try to remove it from this...

rosidl_parser/test/test_parser.py
```diff
@@ -0,0 +1,182 @@
...
+    assert isinstance(constant.type, BasicType)
+    assert constant.type.type == 'uint32'
+    assert constant.value == 42
+
+    constant = [c for c in constants if c.name == 'FLOAT_CONSTANT']
+    assert len(constant) == 1
+    constant = constant[0]
+    assert isinstance(constant.type, BasicType)
+    assert constant.type.type == 'float'
+    assert constant.value == 1.25
```
sloretz: It's probably safe to assume assigning the value `1.25` is the same code path as [`float('1.25')`](https://github.com/ros2/rosidl/blob/7647c6b042bd4c0d4135baf513c34948ec82e729/rosidl_parser/rosidl_parser/parser.py#L478), but [`pytest.approx()`](https://docs.pytest.org/en/latest/reference.html#pytest-approx) is a way to remove that assumption.

dirk-thomas:
Sill good with the `common.lark` file removed: <a href='https://ci.ros2.org/job/ci_linux/5576/'></a>

rosidl_parser/test/test_parser.py
dirk-thomas: For `1.25` that is not necessary since it is a number which can be exactly represented as a floating point number (instead of e.g. `1.2`).

dirk-thomas:
Thanks. I created #329 for the followup improvements.

Squashing the commits before the merge...

dirk-thomas:
> green check mark

That is very suspicious because the PR job has no way of pulling the `lark-parser` until we have a Debian package for it... :thinking:

dirk-thomas:
Full build: [Build Status](https://ci.ros2.org/job/ci_linux/5577/)

dirk-thomas:
@ros-pull-request-builder retest this please

dirk-thomas:
Waiting for the `python3-lark-parser` Debian package...

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5600/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2233/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4609/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5478/)
  * [test failure](https://ci.ros2.org/job/ci_windows/5478/testReport/junit/(root)/rosidl_parser/test_test_parser/) due to encoding: 
  * fixed in 49d24f2506b29b269a7a778adeea828f8078fa43
  * rerun of the specific test only: [Build Status](https://ci.ros2.org/job/ci_windows/5480/)

PR #328, opened 2018-11-19
enable access functions for primitive types

Karsten1987:
for visibility only. I would like to have this merged, but currently there is no nice way (that I know of) to access `std::vector<bool>` as this is a specialization of vector.
https://en.cppreference.com/w/cpp/container/vector_bool

It does not simply return a reference to it's element, `std::vector<bool>` defines a separate class `std::vector<bool>::reference` for it, which is returned by value.

In order to make this patch compile, I explicitly disable these functions for bool.

### DIFF
--- a/rosidl_typesupport_introspection_cpp/resource/msg__type_support.cpp.em
+++ b/rosidl_typesupport_introspection_cpp/resource/msg__type_support.cpp.em
@@ -40,15 +40,22 @@ namespace rosidl_typesupport_introspection_cpp
 
 @[if spec.fields]@
 @[  for field in spec.fields]@
-@[    if not field.type.is_primitive_type() and field.type.is_array]@
+@[    if field.type.is_array and not field.type.type == 'bool']@
+@{
+# from rosidl_generator_cpp import msg_type_only_to_cpp
+# type = msg_type_only_to_cpp(field.type)
+default_type = str(field.type.pkg_name) + '::msg::' + field.type.type
+type = cpp_primitives.get(field.type.type, default_type)
+if field.type.type == 'string':
+    type = 'std::string'
+}@
 size_t size_function__@(spec.base_type.type)__@(field.name)(const void * untyped_member)
 {
 @[      if field.type.array_size and not field.type.is_upper_bound]@
   (void)untyped_member;
   return @(field.type.array_size);
 @[      else]@
-  const auto * member =
-    reinterpret_cast<const std::vector<@(field.type.pkg_name)::msg::@(field.type.type)> *>(untyped_member);
+  const auto * member = reinterpret_cast<const std::vector<@(type)> *>(untyped_member);
   return member->size();
 @[      end if]@
 }
@@ -57,10 +64,10 @@ const void * get_const_function__@(spec.base_type.type)__@(field.name)(const voi
 {
 @[      if field.type.array_size and not field.type.is_upper_bound]@
   const auto & member =
-    *reinterpret_cast<const std::array<@(field.type.pkg_name)::msg::@(field.type.type), @(field.type.array_size)> *>(untyped_member);
+    *reinterpret_cast<const std::array<@(type), @(field.type.array_size)> *>(untyped_member);
 @[      else]@
   const auto & member =
-    *reinterpret_cast<const std::vector<@(field.type.pkg_name)::msg::@(field.type.type)> *>(untyped_member);
+    *reinterpret_cast<const std::vector<@(type)> *>(untyped_member);
 @[      end if]@
   return &member[index];
 }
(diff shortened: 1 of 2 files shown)


rosidl_typesupport_introspection_cpp/resource/msg__type_support.cpp.em
```diff
@@ -69,10 +76,10 @@ void * get_function__@(spec.base_type.type)__@(field.name)(void * untyped_member
...
 @[      if field.type.array_size and not field.type.is_upper_bound]@
   auto & member =
-    *reinterpret_cast<std::array<@(field.type.pkg_name)::msg::@(field.type.type), @(field.type.array_size)> *>(untyped_member);
+    *reinterpret_cast<std::array<@(type), @(field.type.array_size)> *>(untyped_member);
 @[      else]@
   auto & member =
-    *reinterpret_cast<std::vector<@(field.type.pkg_name)::msg::@(field.type.type)> *>(untyped_member);
+    *reinterpret_cast<std::vector<@(type)> *>(untyped_member);
 @[      end if]@
   return &member[index];
```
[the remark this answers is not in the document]
anhosi: Do you need to return a reference to the `bool` inside?
Is `untyped_member` properly initialized or is it zero-initiallized memory that is being `reinterpret_cast`ed? The second case will not work (at least on all platforms) and you need to initialize the memory by doing a placement new.

wjwwood:
> It does not simply return a reference to it's element, std::vector<bool> defines a separate class std::vector<bool>::reference for it, which is returned by value.

Why do you think that? As I read it, `std::vector<bool>::reference` is a typedef to `value_type&` (where `value_type` is just `T`), at least since C++11:

https://en.cppreference.com/w/cpp/container/vector#Member_types

wjwwood:
Oh I see, I didn't understand your description, https://en.cppreference.com/w/cpp/container/vector_bool is a special case.

wjwwood:
Seems like the recommendation is to use something other than bool for the vector of bools (e.g. like `char`) and do conversions:

https://stackoverflow.com/questions/14384880/how-to-prevent-specialization-of-stdvectorbool

I don't see anther way to support array of bool correctly in C++.

/cc @dirk-thomas since you're working on IDL stuff and this would be a change for how we store things.

Karsten1987:
together with @anhosi we decided to explicitly disable the `std::vector<bool>` as for now. Fortunately, the `std::array<bool>` doesn't have the extra API and can thus be considered like every other primitive data type. That's essential for the work we do for rosbag2.

I put this in review and advocate for merging it wo/ support for `std::vector<bool>` even though that might not be the most elegant solution. The rationale is that this code is genuinely enhancing `rosidl_typesupport_introspection_cpp`, even if only partial. 

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5587/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2221/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4596/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5466/)

Karsten1987:
new round of CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5590/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2224/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4600/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5469/)

Karsten1987:
this is currently blocking the work done in https://github.com/ros2/rosbag2/pull/56
@dirk-thomas, you might have the most context maybe for a review. As described above, we enable the functions for all primitives except `std::vector<bool>`

rosidl_typesupport_introspection_cpp/resource/msg__type_support.cpp.em
```diff
@@ -40,15 +41,28 @@ namespace rosidl_typesupport_introspection_cpp
 
 @[if spec.fields]@
 @[  for field in spec.fields]@
-@[    if not field.type.is_primitive_type() and field.type.is_array]@
+@{
+def is_vector_bool(field):
+  return field.type.type == 'bool' and not (field.type.array_size and not field.type.is_upper_bound)
+
```
dirk-thomas: Please remove this empty line.

rosidl_typesupport_introspection_cpp/resource/msg__type_support.cpp.em
```diff
@@ -40,15 +41,28 @@ namespace rosidl_typesupport_introspection_cpp
...
+}@
+@# exclude std::vector<bool> because of specialization in their API
+@[    if field.type.is_array and not is_vector_bool(field)]@
+@{
+# from rosidl_generator_cpp import msg_type_only_to_cpp
+# type = msg_type_only_to_cpp(field.type)
+default_type = str(field.type.pkg_name) + '::msg::' + field.type.type
+type = cpp_primitives.get(field.type.type, default_type)
```
dirk-thomas: Since `type` is a Python builtin use a different name, e.g. `type_`.

PR #330, opened 2018-11-19
fix comment regex in grammar to not be greedy

dirk-thomas:
Otherwise the content between multiple multi line comments is being ignored.


sloretz:
One edge not changed by this PR: I would expect `/**/` to be a legal comment, but this grammar requires at least 1 character in the middle.

dirk-thomas:
> this grammar requires at least 1 character in the middle.

I will take care of this after the round of PRs is in. :memo:

PR #333, opened 2018-11-21
Rmw preallocate

mjcarroll:
Connects to ros2/rmw#159


dirk-thomas:
Please only merge this patch until after the IDL PRs have landed since they refactor the touched code significantly.

mjcarroll:
Will do, I don't think that this will be ready in totality before IDL is merged anyway, but thanks for the heads up.

rosidl_generator_c/src/message_bounds.c
```diff
@@ -0,0 +1,41 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "rosidl_generator_c/message_bounds_struct.h"
+
+#include <assert.h>
+#include <stdio.h>
```
clalancette: Nit: I don't think we need to include stdio.h in here.

rosidl_generator_c/src/message_bounds.c
```diff
@@ -0,0 +1,41 @@
...
+#include <string.h>
+
+const rosidl_message_bounds_t * get_message_bounds_handle(
+  const rosidl_message_bounds_t * handle, const char * identifier)
+{
+  assert(handle);
```
clalancette: I know this is following a pattern from the rest of the code, but it seems like this is encoding an error policy down in this library.  I won't insist that you fix it here, but I would appreciate opening an issue to revisit the use of `assert` in this entire library.

rosidl_generator_c/include/rosidl_generator_c/message_bounds_struct.h
```diff
@@ -0,0 +1,54 @@
...
+
+ROSIDL_GENERATOR_C_PUBLIC
+const rosidl_message_bounds_t * get_message_bounds_handle(
+  const rosidl_message_bounds_t * handle, const char * identifier);
+
+ROSIDL_GENERATOR_C_PUBLIC
+const rosidl_message_bounds_t * get_message_bounds_handle_function(
+  const rosidl_message_bounds_t * handle, const char * identifier);
+
+#define ROSIDL_GET_MSG_BOUNDS(PkgName, MsgSubfolder, MsgName) \
```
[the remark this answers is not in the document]
mjcarroll: Currently not used, this would be in use once the message bounds structures are actually generated in the IDL.  That's now out of scope for this PR and will be added at a later date.

rosidl_generator_c/src/message_bounds.c
mjcarroll: This is a direct copy of what is currently happening in `message_type_support.c`.  I can open an issue to address both cases.

rosidl_generator_c/include/rosidl_generator_c/message_bounds_struct.h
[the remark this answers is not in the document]
clalancette: All right, that's fine, we can leave it.

PR #337, opened 2018-12-04
Adds includes for action wrapper IDL messages fields.

hidmic:
Small pull request, fixing missing includes for action wrapping messages.

### DIFF
--- a/rosidl_parser/rosidl_parser/definition.py
+++ b/rosidl_parser/rosidl_parser/definition.py
@@ -457,7 +457,7 @@ class Action:
 
     __slots__ = (
         'structure_type', 'goal_request', 'result_response', 'feedback',
-        'goal_service', 'result_service', 'feedback_message')
+        'goal_service', 'result_service', 'feedback_message', 'includes')
 
     def __init__(self, type_, goal_request, result_response, feedback):
         """
@@ -550,6 +550,10 @@ class Action:
                 Member(Array(BasicType('uint8'), 16), 'uuid'),
                 Member(feedback.structure.type, 'feedback')]
         ))
+
+        self.includes = [
+            Include('builtin_interfaces/msg/Time.idl')
+        ]
 
 
 class IdlLocator:
--- a/rosidl_parser/rosidl_parser/definition.py
+++ b/rosidl_parser/rosidl_parser/definition.py
@@ -457,7 +457,8 @@ class Action:
 
     __slots__ = (
         'structure_type', 'goal_request', 'result_response', 'feedback',
-        'goal_service', 'result_service', 'feedback_message', 'includes')
+        'goal_service', 'result_service', 'feedback_message',
+        'implicit_includes')
 
     def __init__(self, type_, goal_request, result_response, feedback):
         """
@@ -493,6 +494,11 @@ class Action:
         assert feedback.structure.type.name == type_.name + \
             ACTION_FEEDBACK_MESSAGE_SUFFIX
         self.feedback = feedback
+
+        # necessary include for injected timestamp member
+        self.implicit_includes = [
+            Include('builtin_interfaces/msg/Time.idl')
+        ]
 
         # derived types
         goal_service_name = type_.name + ACTION_WRAPPER_TYPE_SUFFIX + \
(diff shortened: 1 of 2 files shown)


rosidl_parser/rosidl_parser/definition.py
```diff
@@ -551,6 +551,10 @@ def __init__(self, type_, goal_request, result_response, feedback):
                 Member(feedback.structure.type, 'feedback')]
         ))
 
+        self.includes = [
+            Include('builtin_interfaces/msg/Time.idl')
+        ]
+
```
dirk-thomas: Includes are not specific to an `Action` but are available through the `IdlContent` instance. Therefore I don't think this should be added to the `Action` class.

rosidl_parser/rosidl_parser/parser.py
```diff
@@ -235,6 +235,7 @@ def extract_content_from_ast(tree):
                 name=goal_request_basename),
             goal_request, result_response, feedback_message)
         content.elements.append(action)
+        content.elements.extend(action.includes)
```
dirk-thomas: This could add an explicit `Include` instead.

rosidl_parser/rosidl_parser/definition.py
hidmic: Since it is the `Action` that's adding the field, I thought it was cleaner to keep it as an implementation detail.

rosidl_parser/rosidl_parser/parser.py
hidmic: Sure, that's the other option, assuming we keep both parser and `Action` specific fields in sync.

rosidl_parser/rosidl_parser/parser.py
```diff
@@ -234,6 +234,7 @@ def extract_content_from_ast(tree):
                 namespaces=goal_request.structure.type.namespaces,
                 name=goal_request_basename),
             goal_request, result_response, feedback_message)
+        content.elements += action.implicit_includes
```
dirk-thomas: This might result in a duplicate include though.

dirk-thomas: Why not add the deduplication here (rather than in potentially every location which uses this API)?

rosidl_parser/rosidl_parser/parser.py
```diff
@@ -234,7 +234,13 @@ def extract_content_from_ast(tree):
                 namespaces=goal_request.structure.type.namespaces,
                 name=goal_request_basename),
             goal_request, result_response, feedback_message)
-        content.elements.append(action)
+
+        all_includes = content.get_elements_of_type(Include)
+        unique_include_locators = set([
+            include.locator for include in all_includes])
```
dirk-thomas: I would expect the linter to warn about this. This likely needs to be a set comprehension instead.

dirk-thomas: @hidmic Once you have checked this (and updated if necessary) please squash merge this PR into `idl-stage-7`.

jacobperron:
Running CI with this branch seems to resolve the current build error, moving onto another:

[Build Status](https://ci.ros2.org/job/ci_linux/5772/)

```
04:26:19 /home/rosbuild/ci_scripts/ws/src/ros2/rcl_interfaces/test_msgs/test/test_cpp_type_support.cpp: In member function ‘virtual void ActionCppTypeSupport_can_get_type_support_Test::TestBody()’:
04:26:19 /home/rosbuild/ci_scripts/ws/src/ros2/rcl_interfaces/test_msgs/test/test_cpp_type_support.cpp:24:60: error: ‘test_msgs’ was not declared in this scope
04:26:19      rosidl_typesupport_cpp::get_action_type_support_handle<test_msgs::action::Fibonacci>();
```

dirk-thomas:
I am going ahead and merge this in order to be able to build and test from a single branch.

rosidl_parser/rosidl_parser/parser.py
```diff
@@ -234,7 +234,13 @@ def extract_content_from_ast(tree):
                 namespaces=goal_request.structure.type.namespaces,
                 name=goal_request_basename),
             goal_request, result_response, feedback_message)
-        content.elements.append(action)
```
dirk-thomas: The patch remove the appending of the action.

PR #339, opened 2018-12-07
Reduces action interfaces CMake target name length to please Windows.

hidmic:
Fixes ros2/build_cop#154. Had to reduce the name length for action targets a bit. I did so everywhere for consistency.


hidmic:
Thanks @sloretz ! Merging.

PR #355, opened 2019-03-21
Add message namespace to MessageMembers struct

jacobperron:
The message namespace is useful in rmw implementations for reconstructing the full name of message types.

Connects to ros2/ros2#677


rosidl_typesupport_introspection_c/include/rosidl_typesupport_introspection_c/message_introspection.h
```diff
@@ -43,6 +43,7 @@ typedef struct rosidl_typesupport_introspection_c__MessageMember
 typedef struct rosidl_typesupport_introspection_c__MessageMembers
 {
   const char * package_name_;
+  const char * message_namespace_;
```
dirk-thomas: Shouldn't the `package_name_` field be removed?

jacobperron: Yes, I thought I did that. Oops.
Done, also updated service structs: f5a5db0

PR #360, opened 2019-04-17
rename API of type objects

dirk-thomas:
The current state is not ready to be merged yet but I am looking for a review of the proposed changes. Once this is reviewed / approved I will start working on updating all message generators.

I will add some inline comments describing why I made some of the changes.

### DIFF
--- a/rosidl_parser/rosidl_parser/definition.py
+++ b/rosidl_parser/rosidl_parser/definition.py
@@ -12,34 +12,78 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from collections import OrderedDict
 import pathlib
+from typing import Iterable
+from typing import Tuple
 
+# Basic types as defined by the IDL specification
 
-"""Basic types as defined by the IDL specification."""
-BASIC_TYPES = [
-    'short',
-    'unsigned short',
-    'long',
-    'unsigned long',
-    'long long',
-    'unsigned long long',
+# 7.4.1.4.4.2 Basic Types
+SIGNED_NONEXPLICIT_INTEGER_TYPES = (  # rules (26)
+    'short',  # rule (27)
+    'long',  # rule (28)
+    'long long',  # rule (29)
+)
+UNSIGNED_NONEXPLICIT_INTEGER_TYPES = (  # rules (30)
+    'unsigned short',  # rule (31)
+    'unsigned long',  # rule (32)
+    'unsigned long long',  # rule (33)
+)
+NONEXPLICIT_INTEGER_TYPES = (
+    *SIGNED_NONEXPLICIT_INTEGER_TYPES,
+    *UNSIGNED_NONEXPLICIT_INTEGER_TYPES,
+)
+FLOATING_POINT_TYPES = (  # rule (24)
     'float',
     'double',
     'long double',
-    'char',
-    'wchar',
-    'boolean',
-    'octet',
-    'int8',
-    'uint8',
-    'int16',
-    'uint16',
-    'int32',
-    'uint32',
-    'int64',
-    'uint64',
-]
+)
+CHARACTER_TYPES = (
+    'char',  # rule (34)
+    'wchar',  # rule (35)
+)
+BOOLEAN_TYPE = 'boolean'  # rule (36)
+OCTET_TYPE = 'octet'  # rule (37)
+
+# 7.4.13.4.4 Integers restricted to holding 8-bits of information
+# 7.4.13.4.5 Explicitly-named Integer Types
+SIGNED_EXPLICIT_INTEGER_TYPES = (
+    'int8',  # rule (208)
+    'int16',  # rule (210)
+    'int32',  # rule (211)
+    'int64',  # rule (212)
+)
+UNSIGNED_EXPLICIT_INTEGER_TYPES = (
+    'uint8',  # rule (209)
+    'uint16',  # rule (213)
+    'uint32',  # rule (214)
+    'uint64',  # rule (215)
+)
+EXPLICIT_INTEGER_TYPES = (
+    *SIGNED_EXPLICIT_INTEGER_TYPES,
+    *UNSIGNED_EXPLICIT_INTEGER_TYPES,
+)
+SIGNED_INTEGER_TYPES = (   # rules (26) + (208) + (210-212)
+    *SIGNED_NONEXPLICIT_INTEGER_TYPES,
+    *SIGNED_EXPLICIT_INTEGER_TYPES,
+)
+UNSIGNED_INTEGER_TYPES = (   # rules (30) + (209) + (213-215)
+    *UNSIGNED_NONEXPLICIT_INTEGER_TYPES,
+    *UNSIGNED_EXPLICIT_INTEGER_TYPES,
+)
+INTEGER_TYPES = (   # rules (25) + (206-207) + (210-215)
+    *SIGNED_INTEGER_TYPES,
+    *UNSIGNED_INTEGER_TYPES,
+)
+
+"""All basic types as defined by the IDL specification."""
+BASIC_TYPES = (
+    *INTEGER_TYPES,
+    *FLOATING_POINT_TYPES,
+    *CHARACTER_TYPES,
+    BOOLEAN_TYPE,
+    OCTET_TYPE,
+)
 
 CONSTANT_MODULE_SUFFIX = '_Constants'
 
@@ -56,7 +100,7 @@ ACTION_FEEDBACK_MESSAGE_SUFFIX = '_FeedbackMessage'
 
 
 class AbstractType:
-    """The base class for all types."""
+    """The abstract base class for all types."""
 
     __slots__ = ()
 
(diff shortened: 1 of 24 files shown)


rosidl_parser/rosidl_parser/definition.py
```diff
@@ -14,32 +14,77 @@
...
+)
+
+"""All basic types as defined by the IDL specification."""
+BASIC_TYPES = (
+    *INTEGER_TYPES,
+    *FLOATING_POINT_TYPES,
+    *CHARACTER_TYPES,
+    BOOLEAN_TYPE,
+    OCTET_TYPE,
+)
```
dirk-thomas: Providing constants for the various groups of types allows message generators to refer to those (e.g. `INTEGER_TYPES`) instead of needing to enumerate them explicitly.

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -56,49 +101,64 @@
...
 
     __slots__ = ()
 
     def __eq__(self, other):
         return type(self) == type(other)
 
 
-class BaseType(AbstractType):
-    """The base class for types which can be used inside nested types."""
+class AbstractNestableType(AbstractType):
```
dirk-thomas: Any class which is not supposed to instantiated directly starts with `Abstract` now.

dirk-thomas: Instead of using the fuzzy name `BaseType` the term `Nestable` should be clearer. Also the docblock elaborates on nested and nestable types.

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -56,49 +101,64 @@
...
+class BasicType(AbstractNestableType):
+    """
+    A basic type according to the IDL specification.
+
+    Its `typename` attribute defines the specific numeric, character, boolean
+    and octet type.
+    """
 
-    __slots__ = ('type', )
+    __slots__ = ('typename', )
```
dirk-thomas: Instead of using the generic name `type` for multiple slots I changed the slot names to something more specific. Instead of accessing something like `foo.type.type.type` it will be more readable like `foo.structure.namespaced_type.typename` (just a made up example).

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -107,170 +167,231 @@ def __eq__(self, other):
...
+        :param namespaces: the names of nested namespaces identifying a
+          specific scope
+        :param name: the name of the type within that scope
         """
         super().__init__()
         self.namespaces = namespaces
         self.name = name
 
+    def namespaced_name(self) -> Tuple[str, ...]:
+        return (*self.namespaces, self.name)
```
dirk-thomas: Since this is used together in a lot of places providing the combination in a method makes sense.

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -107,170 +167,231 @@ def __eq__(self, other):
...
 
-class BaseString(BaseType):
-    """The base class of string types."""
+class AbstractGenericString(AbstractNestableType):
+    """The abstract base class of all string types."""
+
+    __slots__ = ()
+
+    def has_maximum_size(self):
+        raise NotImplementedError('Only implemented in subclasses')
```
dirk-thomas: The type hierarchy can only provide one dimension to differentiate types. But in some cases the character type of a string (8-bit vs. 16-bit) isn't relevant but the logic wants to treat bounded and unbounded strings separately. While it could check for multiple types this method provides an easy way to distinguish this second dimension.

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -107,170 +167,231 @@ def __eq__(self, other):
...
+
+
+class AbstractString(AbstractGenericString):
```
dirk-thomas: An in-between abstract class for 8-bit strings (bounded and unbounded). There a similar one for 16-bit strings below. It just ease the logic instead of having to check for two concrete types each time.

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -107,170 +167,231 @@ def __eq__(self, other):
...
+    """
+    The abstract base class of nested types.
 
-    __slots__ = ('basetype', )
+    Arrays and sequences are nested types which contain elements of nestable
+    types.
+    """
 
-    def __init__(self, basetype):
+    __slots__ = ('value_type', )
```
dirk-thomas: The term `value_type` is inspired by the naming in STL containers.

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -410,46 +530,49 @@ class Message:
...
         self.structure = structure
-        self.constants = OrderedDict()
+        self.constants = []
 
 
 class Service:
     """A namespaced type containing a request and response message."""
 
-    __slots__ = ('structure_type', 'request_message', 'response_message')
+    __slots__ = ('namespaced_type', 'request_message', 'response_message')
```
dirk-thomas: Similar to the comment above `namespaced_type` better conveys what is contained in this slot.

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -410,46 +530,49 @@ class Message:
...
         Constructor.
 
-        :param Structure structure: the structure of the message
+        :param structure: the structure of the message
         """
         super().__init__()
         assert isinstance(structure, Structure)
         self.structure = structure
-        self.constants = OrderedDict()
+        self.constants = []
```
dirk-thomas: Everywhere this is used only `.values()` is being access. Therefore changing the data structure accordingly. The parser already ensures that e.g. no duplicates are being added.

rosidl_parser/rosidl_parser/definition.py
jacobperron: Depending on how this is used, consider making this a property, adding a little sugar.

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -14,32 +14,77 @@
...
+)
+SIGNED_INTEGER_TYPES = (   # rules (26) + (208) + (210-212)
+    *SIGNED_NONEXPLICIT_INTEGER_TYPES,
+    *SIGNED_EXPLICIT_INTEGER_TYPES,
+)
+UNSIGNED_INTEGER_TYPES = (   # rules (30) + (209) + (213-215)
+    *UNSIGNED_NONEXPLICIT_INTEGER_TYPES,
+    *UNSIGNED_EXPLICIT_INTEGER_TYPES,
+)
+INTEGER_TYPES = (   # rules (25) + ()
```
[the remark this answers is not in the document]
dirk-thomas: Good catch - should be:

`# rules (25) + (206-207) + (210-215)`

I addressed it locally (just not in a state where I can push it :wink:).

rosidl_parser/rosidl_parser/definition.py
dirk-thomas: While it saves a little bit of code the downside is that the calling code needs to explicitly import the function. In favor of keeping the calling code simpler I would prefer spending the extra lines in this code.

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -56,49 +101,64 @@
...
-    """The base class for types which can be used inside nested types."""
+class AbstractNestableType(AbstractType):
+    """
+    The abstract base class for types which can be used inside nested types.
+
+    Arrays and sequences are nested types which contain elements of nestable
+    types.
+    Nestable types are:
+    - BasicType like numerics, character types, boolean and octet
+    - NamedType identified by an name which hasn't been resolved yet
```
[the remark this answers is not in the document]
dirk-thomas: Thanks - addressed locally.

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -56,49 +101,64 @@
...
+    - NamespacedType which describe another Structure
```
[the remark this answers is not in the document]
dirk-thomas: Thanks - addressed locally.

rosidl_parser/rosidl_parser/definition.py
dirk-thomas: I would lean towards keeping it a method so the caller sees what is actually an attribute and what is derived data. If others too think it should be a property I am happy to change it though.

dirk-thomas:
I (force) pushed a commit with all the feedback addressed. I squashed the changes to keep them separate from the next ones updating the code to use that modified API.

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -107,170 +166,233 @@ def __eq__(self, other):
...
-        :param int maximum_size: the maximum length of the string in
-          characters, or None if unlimited
+        :param maximum_size: the maximum length of the string in characters
+          (must be greater than zero)
         """
-        super().__init__(maximum_size=maximum_size)
+        super().__init__()
+        # TODO(dirk-thomas) can't be enforced yet since the parser might pass a
+        # constant name
+        # assert maximum_size > 0
```
dirk-thomas: While I wanted to add this `assert` it turns out the value is currently also being used to temporary store a constant name during the parsing which gets replaced with the constant value shortly after.

I won't change this during this renaming but added a note to be aware of it.

dirk-thomas:
Preliminary builds passed already so this is a set of lengthy builds with all three RMW impl. enabled:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6740/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3048/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5537/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6564/)

All the referenced PRs only update the naming and don't contain any other changes. I will create separate PRs to actually simplify the logic after this is merged.

dirk-thomas:
Please also review the changes of the second commit as well as the connected PRs which update the code base to use the renamed API.

dirk-thomas:
As green(ish) as it gets these days (with a tint of yellow known to flaky or failing) :wink:

PR #363, opened 2019-04-19
Refactor message definitions for consistency

jacobperron:
- Rename PrimitivesStatic.msg -> Primitives.msg
Previously a misnomer as it contains primitives instead of static arrays.
- Rename PrimitivesStaticArrays.msg -> PrimitivesStatic.msg
- Rename message fields to reflect that they are arrays.


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6741/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3049/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5538/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6565/)

dirk-thomas:
* Please use separate commits for refacing files without changes to them.
* Message field were using plural (`_values`) to indicate that they are arrays / sequences. I am not sure if they need to be renamed.
* Also the term `array` isn't good when the size is not static. Those are being called sequences in the IDL spec. Therefore I would suggest to use that terminology (as in the current design docs).

jacobperron:
> Please use separate commits for refacing files without changes to them.

Will do.

>Message field were using plural (_values) to indicate that they are arrays / sequences. I am not sure if they need to be renamed.

Yeah, I'll revert that.

> Those are being called sequences in the IDL spec. Therefore I would suggest to use that terminology (as in the current design docs).

That makes sense. Perhaps we should also update the message names accordingly, e.g.
- `PrimitivesStatic` -> `PrimitivesArray`
- `PrimitivesBounded` -> `PrimitivesBoundedArray`
- `PrimitivesUnbounded` -> `PrimitivesSequence`

And the nested messages too.

If you agree, I can make the changes.

jacobperron:
I think I got that slightly wrong, it should be `PrimitivesBoundedSequence` (instead of `BoundedArray`).

dirk-thomas:
The last one sounds good.

jacobperron:
Replaced by #367

PR #367, opened 2019-04-24
[rosidl_generator_cpp] Use message definitions from the package test_interface_files

jacobperron:
Connects to ros2/rcl_interfaces#58

This is part of an effort to keep all test messages in a single location.
I've left TODOs for missing interfaces that we should consider adding to test_interface_files.

Signed-off-by: Jacob Perron <jacob@openrobotics.org>


rosidl_generator_cpp/test/test_interfaces.cpp
```diff
@@ -22,34 +22,30 @@
 #include <algorithm>
 #include "test_array_generator.hpp"
 
+#include "rosidl_generator_cpp/msg/array_primitives.hpp"
+#include "rosidl_generator_cpp/msg/array_nested.hpp"
+// TODO(jacobperron): Add test case for ArrayBoundedPrimitivesNested (need interface)
```
jacobperron: I'd like to know others thoughts on adding this (and the other missing interface files) to `test_interface_files`. This would help maintain the same level of test coverage in this package, but introduce several more interface files. I guess more interface files isn't a big worry since other packages could ignore those they are not interested in testing.

rosidl_generator_cpp/test/test_interfaces.cpp
```diff
@@ -153,77 +147,74 @@ TEST(Test_rosidl_generator_traits, has_bounded_size) {
     "PrimitivesDefault::has_bounded_size is true");
 
   static_assert(
-    rosidl_generator_traits::has_bounded_size<rosidl_generator_cpp::msg::PrimitivesStatic>::value,
-    "PrimitivesStatic::has_bounded_size is false");
-
-  static_assert(
-    rosidl_generator_traits::has_bounded_size<rosidl_generator_cpp::msg::PrimitivesBounded>::value,
-    "PrimitivesBounded::has_bounded_size is false");
+    !rosidl_generator_traits::has_bounded_size<rosidl_generator_cpp::msg::Primitives>::value,
```
jacobperron: I'd like to point out that before we expected this message to have bounded size. But I think since we [introduced a field with type `string`](https://github.com/ros2/test_interface_files/blob/41c8c697dab6b2c7254aa0cfe051842e0e3a8bd4/msg/Primitives.msg#L14), it no longer has bounded size.

dirk-thomas: Would it make sense to keep the original message and continue to test that the complete message is bounded? I guess the test wants to make sure that a message composed out of bounded fields is still bounded.

rosidl_generator_cpp/test/test_interfaces.cpp
dirk-thomas: Adding them for now might be ok. We should probably consider to consolidate some of the messages based on what exactly each of them is aiming to cover.

jacobperron: I think it's a good thing to cover all combinations of "nested type of nested type":

- ArrayBoundedSequenceNested
- ArrayUnboundedSequenceNested
- BoundedSequenceArrayNested
- BoundedSequenceUnboundedSequenceNested
- UnboundedSequenceArrayNested
- UnboundedSequenceBoundedSequenceNested

Which happen to be those that are missing here.

jacobperron:
Replaced by https://github.com/ros2/rosidl/pull/397

PR #369, opened 2019-04-27
Add dependency to action_msgs when generating action interfaces

jacobperron:
Fixes https://github.com/ros2/rcl_interfaces/issues/75

I'm not sure if this is the correct thing to do. @dirk-thomas, maybe you can tell me otherwise.

### DIFF
--- a/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
+++ b/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
@@ -133,6 +133,17 @@ macro(rosidl_generate_interfaces target)
   # afterwards all remaining interface files are .idl files
   list(APPEND _idl_tuples ${_idl_adapter_tuples})
 
+  # to generate .action interfaces, we need to depend on "action_msgs"
+  foreach(_tuple ${_interface_tuples})
+    get_filename_component(_extension "${_tuple}" EXT)
+    if("${_extension}" STREQUAL ".action")
+      find_package(action_msgs REQUIRED)
+      list_append_unique(_ARG_DEPENDENCIES "action_msgs")
+      ament_export_dependencies(action_msgs)
+      break()
+    endif()
+  endforeach()
+
   # collect all interface files from dependencies
   set(_dep_files)
   foreach(_dep ${_ARG_DEPENDENCIES})
--- a/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
+++ b/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
@@ -133,10 +133,11 @@ macro(rosidl_generate_interfaces target)
   # afterwards all remaining interface files are .idl files
   list(APPEND _idl_tuples ${_idl_adapter_tuples})
 
-  # to generate .action interfaces, we need to depend on "action_msgs"
+  # to generate action interfaces, we need to depend on "action_msgs"
   foreach(_tuple ${_interface_tuples})
-    get_filename_component(_extension "${_tuple}" EXT)
-    if("${_extension}" STREQUAL ".action")
+    string(REGEX REPLACE ".*:([^:]*)$" "\\1" _tuple_file "${_tuple}")
+    get_filename_component(_parent_dir "${_tuple_file}" DIRECTORY)
+    if("${_parent_dir}" STREQUAL "action")
       find_package(action_msgs REQUIRED)
       list_append_unique(_ARG_DEPENDENCIES "action_msgs")
       ament_export_dependencies(action_msgs)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The current patch didn't work if the action is defined in an `.idl` file.

jacobperron:
> The current patch didn't work if the action is defined in an .idl file.

Do you have a recommendation how to identify if an action is being defined in an `.idl` file?
A naive solution would be to include the dependency regardless of if there are any action definitions.
Or I can add a TODO/issue to come back to this since it is at least covering the common case of using `.action` files.

dirk-thomas:
Instead of checking the extension it can check the directory name the interface file is located in.

jacobperron:
Now we are checking the directory instead of the extension to infer if the interface is an action.

The current problem is ensuring that `action_msgs` is built before `rosidl_generate_interfaces` is called.
I'm not sure where to make this happen.

jacobperron:
Since `action_msgs` itself depends on other message packages (`builtin_interfaces` and `unique_identifier_msgs`) and the generators themselves it seems like the only thing we can do is require that the user add a build dependency to `action_msgs`. We can at least eliminate the cmake lines for finding and depending on it and produce a nice error message in the event that they forget to declare a build depend tag in package.xml.

jacobperron:
See https://github.com/ros2/example_interfaces/pull/6 for how this makes life a little easier for users.

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
```diff
@@ -133,6 +133,23 @@ macro(rosidl_generate_interfaces target)
   # afterwards all remaining interface files are .idl files
   list(APPEND _idl_tuples ${_idl_adapter_tuples})
 
+  # to generate action interfaces, we need to depend on "action_msgs"
+  foreach(_tuple ${_interface_tuples})
+    string(REGEX REPLACE ".*:([^:]*)$" "\\1" _tuple_file "${_tuple}")
```
dirk-thomas: The leading `.*` should be removed from the pattern.

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
```diff
@@ -133,6 +133,23 @@ macro(rosidl_generate_interfaces target)
...
+    get_filename_component(_parent_dir "${_tuple_file}" DIRECTORY)
+    if("${_parent_dir}" STREQUAL "action")
+      find_package(action_msgs QUIET)
+      if(NOT ${action_msgs_FOUND})
+        message(FATAL_ERROR "Unable to generate action interface for '${_tuple_file}'. "
+          "In order to generate action interfaces, you must add a build dependency to 'action_msgs' "
```
dirk-thomas: Not only a build dependency since you also want to export it. So a build and run dependency (`depend`) is necessary.

dirk-thomas:
Also a note about this "magic" should be added to the docblock of the macro.

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
jacobperron: I only want the second part of the tuple. Without the leading `.*` I will get the full path to the interface file.

jacobperron:
Building with connected PRs. Testing things known to use actions (`rcl_action`, `rclcpp_action`, and `rclpy`) and `test_msgs`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7263/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3443/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5943/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7079/)

jacobperron:
Looks like builds with OpenSplice aren't happy with this solution: http://build.ros2.org/job/Ddev__example_interfaces__ubuntu_bionic_amd64/4/

:fearful:

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
```diff
@@ -133,6 +137,23 @@ macro(rosidl_generate_interfaces target)
...
 
+  # to generate action interfaces, we need to depend on "action_msgs"
+  foreach(_tuple ${_interface_tuples})
+    string(REGEX REPLACE ".*:([^:]*)$" "\\1" _tuple_file "${_tuple}")
+    get_filename_component(_parent_dir "${_tuple_file}" DIRECTORY)
+    if("${_parent_dir}" STREQUAL "action")
+      find_package(action_msgs QUIET)
+      if(NOT ${action_msgs_FOUND})
+        message(FATAL_ERROR "Unable to generate action interface for '${_tuple_file}'. "
+          "In order to generate action interfaces you must add a depend tag for 'action_msgs' "
```
dirk-thomas: Linter warning: see #382.

PR #371, opened 2019-04-30
Handle INT64_MIN value

jacobperron:
Before the unary minus, the compiler first evaluates the literal, which is larger than the maximum allowed integer and results in a warning.
To correctly get the value for INT64_MIN, we substract from one number larger.
See https://en.cppreference.com/w/cpp/language/integer_literal.

Also adding missing integer suffixes for generated C code.

### DIFF
--- a/rosidl_generator_c/rosidl_generator_c/__init__.py
+++ b/rosidl_generator_c/rosidl_generator_c/__init__.py
@@ -138,14 +138,14 @@ def basic_value_to_c(type_, value):
     if 'boolean' == type_.typename:
         return 'true' if value else 'false'
 
-    if type_.typename in [
-        'short', 'unsigned short',
-        'char', 'wchar',
-        'double',
-        'octet',
-        'int8', 'uint8',
-        'int16', 'uint16',
-    ]:
+    if type_.typename in (
+        *CHARACTER_TYPES,
+        OCTET_TYPE,
+        'int8',
+        'uint8',
+        'int16',
+        'uint16',
+    ):
         return str(value)
 
     if type_.typename == 'int32':
@@ -158,7 +158,7 @@ def basic_value_to_c(type_, value):
         # Handle edge case for INT64_MIN
         # See https://en.cppreference.com/w/cpp/language/integer_literal
         if -9223372036854775808 == value:
-            return '({0} - 1)'.format(value + 1)
+            return '({0}ll - 1)'.format(value + 1)
         return '{value}ll'.format_map(locals())
 
     if type_.typename == 'uint64':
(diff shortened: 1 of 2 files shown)


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6820/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3115/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5608/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6642/)

rosidl_generator_c/rosidl_generator_c/__init__.py
```diff
@@ -138,11 +138,31 @@ def basic_value_to_c(type_, value):
     if 'boolean' == type_.typename:
         return 'true' if value else 'false'
 
-    if type_.typename in (*SIGNED_INTEGER_TYPES, *CHARACTER_TYPES, OCTET_TYPE):
+    if type_.typename in [
+        'short', 'unsigned short',
+        'char', 'wchar',
+        'double',
```
dirk-thomas: This looks wrong. Please keep using the type constants possible: `CHARACTER_TYPES`, `OCTET_TYPE`.

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
```diff
@@ -190,6 +190,10 @@ def primitive_value_to_cpp(type_, value):
         return '%sul' % value
 
     if type_.typename == 'int64':
+        # Handle edge case for INT64_MIN
+        # See https://en.cppreference.com/w/cpp/language/integer_literal
+        if -9223372036854775808 == value:
+            return '(%s - 1)' % (value + 1)
```
dirk-thomas: Shouldn't the `%s` be followed by a `ll`?

rosidl_generator_c/rosidl_generator_c/__init__.py
jacobperron: Hmm.. okay. I copied what was being done over in `rosidl_generator_cpp`:

https://github.com/ros2/rosidl/blob/d1edcaf35534e8860730370a62f1d6a024ee0f66/rosidl_generator_cpp/rosidl_generator_cpp/__init__.py#L176-L184

I'll update this file, but perhaps a follow-up is needed for rosidl_generator_cpp.

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
jacobperron: Yes, e1c1b28

rosidl_generator_c/rosidl_generator_c/__init__.py
jacobperron: e1c1b28

Issue #352, opened 2019-03-12
support for WString across all generators


dirk-thomas:
Follow up of ros2/rosidl#346.

dirk-thomas:
* Linux [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_linux&build=6895)](http://ci.ros2.org/job/ci_linux/6895/)
* Linux-aarch64 [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_linux-aarch64&build=3173)](http://ci.ros2.org/job/ci_linux-aarch64/3173/)
* macOS [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_osx&build=5661)](http://ci.ros2.org/job/ci_osx/5661/)
* Windows [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_windows&build=6716)](http://ci.ros2.org/job/ci_windows/6716/)

jacobperron:
Looks like you've addressed all the warnings/failures. Considering you are adding more work on top of this, we can merge as-is and catch anything we missed in the next round of CI.

dirk-thomas:
I went over all PRs again and resolved the remaining comments. In favor of time I will merge the changes now and trigger a set of CI builds afterwards to double check. :crossed_fingers:

PR #374, opened 2019-05-01
change C u16string data from uint16_t to uint_least16_t

dirk-thomas:
In order to match how a C++ `std::u16string` is defined.

It avoids the risk of `uint16_t` and `uint_least16_t` not being the same lengths when doing stuff like this https://github.com/ros2/rosidl_typesupport_opensplice/pull/27/files#diff-eb52319d09d2b6415f95c504f6609ba7R244:

> `dds_message->@(member.name)_[i] = cv.to_bytes(reinterpret_cast<char16_t *>(str->data)).c_str();`

or the need to copy character by character to prevent the datatype mismatch on some potential platforms.

This is ready for review but I will hole this PR until ros2/rosidl#352 has landed and then update downstream code and run CI.

### DIFF
--- a/rosidl_generator_c/include/rosidl_generator_c/u16string.h
+++ b/rosidl_generator_c/include/rosidl_generator_c/u16string.h
@@ -22,7 +22,8 @@
 /// U16String struct
 typedef struct rosidl_generator_c__U16String
 {
-  uint16_t * data;
+  /// The pointer to the first character.
+  uint_least16_t * data;  // using uint_least16_t to match a C++ std::u16string
   /// The length of the u16string (excluding the null byte).
   size_t size;
   /// The capacity represents the number of allocated characters (including the null byte).


dirk-thomas:
The design article needs to be updated accordingly (which currently falsely states `char16_t` as the mapping in C which doesn't exist as a C type).

dirk-thomas:
* Linux up to `test_msgs`: [Build Status](https://ci.ros2.org/job/ci_linux/6915/)

dirk-thomas:
Full CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6917/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3187/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5676/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6737/)

PR #375, opened 2019-05-01
Handle INT32_MIN value

jacobperron:
Similar to INT64_MIN value, but only because MSVC complains.


dirk-thomas:
Example of the warnings in CI: https://ci.ros2.org/view/packaging/job/packaging_windows/1360/warnings43Result/

PR #376, opened 2019-05-03
use non-bool type for structure_needs_at_least_one_member field

dirk-thomas:
Since user land code doesn't know about this field it might end up uninitialized. In case of FastRTPS it raises an exception if a boolean value is received which isn't either `0` or `1`.

Using a `uint8` instead since any uninitialized value is valid.

After this it becomes feasible to consider raising on the publish side if an uninitialized boolean value is published.


dirk-thomas:
In combination with:
* ros2/system_tests#355
* ros2/rosidl_typesupport_fastrtps#24

dirk-thomas:
Before: [Build Status](https://ci.ros2.org/job/ci_linux/6956/)
After: [Build Status](https://ci.ros2.org/job/ci_linux/6962/)

PR #379, opened 2019-05-14
Fix condition for including resize functions

jacobperron:
Resolves #378 

Bug introduced by #334.

PR #380, opened 2019-05-14
Fix resize function (#379) for C.

malsbat:
Commit 6f8194a673a8ec1467dc87858a1fee80b51060b5 fixed https://github.com/ros2/rosidl/pull/379 for C++.  This commit fixes the same issue for C.

Signed-off-by: Todd Malsbary <todd.malsbary@intel.com>


malsbat:
@jacobperron, some more testing revealed the same issue with the C resize function.

dirk-thomas:
Thanks for the fix.

PR #390, opened 2019-07-18
open interface files with utf-8 encoding

### LINKED ISSUE #388: Error: 'charmap' codec can't decode byte 0x8f in position 156
## Bug report

**Required Info:**

- Operating System:
  - Windows 10 ONLY
- Installation type:
  - Binary of nightly build (http://ci.ros2.org/view/packaging/job/packaging_windows/lastSuccessfulBuild/artifact/ws/ros2-package-windows-AMD64.zip)
- Version or commit hash:
- DDS implementation:
- Client library (if applicable):

#### Steps to reproduce issue

Pass the path of [WStrings.msg](https://github.com/ros2/test_interface_files/blob/master/msg/WStrings.msg) as the parameter of Python method [parse_message_file](https://github.com/ros2/rosidl/blob/ced307012bda3413d6c495914bf517d0d03748f1/rosidl_adapter/rosidl_adapter/parser.py#L427)

#### Expected behavior
Execute successfully
#### Actual behavior
Error: 'charmap' codec can't decode byte 0x8f in position 156: character maps to <undefined>
#### Additional information
- Python version: Python3.7 x64
- Suspection: there are some special characters though commented out, and I found possible some potential reason, see https://stackoverflow.com/questions/9233027/unicodedecodeerror-charmap-codec-cant-decode-byte-x-in-position-y-character

Thanks!

dirk-thomas:
Fixes #388.

[Build Status](https://ci.ros2.org/job/ci_linux/7610/)

### DIFF
--- a/rosidl_adapter/rosidl_adapter/parser.py
+++ b/rosidl_adapter/rosidl_adapter/parser.py
@@ -427,7 +427,7 @@ class MessageSpecification:
 def parse_message_file(pkg_name, interface_filename):
     basename = os.path.basename(interface_filename)
     msg_name = os.path.splitext(basename)[0]
-    with open(interface_filename, 'r') as h:
+    with open(interface_filename, 'r', encoding='utf-8') as h:
         return parse_message_string(
             pkg_name, msg_name, h.read())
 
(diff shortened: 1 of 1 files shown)


rosidl_adapter/rosidl_adapter/parser.py
```diff
@@ -427,7 +427,7 @@ def __str__(self):
 def parse_message_file(pkg_name, interface_filename):
     basename = os.path.basename(interface_filename)
     msg_name = os.path.splitext(basename)[0]
-    with open(interface_filename, 'r') as h:
+    with open(interface_filename, 'r', encoding='utf-8') as h:
```
sloretz: This seems fine to me. Just noting that this requires interface files to be utf-8, where they might have worked fine if they happened to be using the default system encoding before.

PR #391, opened 2019-07-19
use latin-1 encoding when reading/writing .idl files, prepend BOM to generated C/C++ files when necessary

dirk-thomas:
Addresses the problem on Windows related to ros2/test_interface_files#8.

[Build Status](https://ci.ros2.org/job/ci_windows/7438/) (only passed because these tests are disabled: ros2/system_tests#377).

### DIFF
--- a/rosidl_generator_c/rosidl_generator_c/__init__.py
+++ b/rosidl_generator_c/rosidl_generator_c/__init__.py
@@ -34,7 +34,20 @@ def generate_c(generator_arguments_file):
         'idl__struct.h.em': '%s__struct.h',
         'idl__type_support.h.em': '%s__type_support.h',
     }
-    generate_files(generator_arguments_file, mapping)
+    generate_files(
+        generator_arguments_file, mapping,
+        post_process_callback=prefix_with_bom_if_necessary)
+
+
+def prefix_with_bom_if_necessary(content):
+    try:
+        content.encode('ASCII')
+    except UnicodeError:
+        prefix = '\ufeff' + \
+            '// NOLINT: This file starts with a BOM ' + \
+            'since it contain non-ASCII characters\n'
+        content = prefix + content
+    return content
 
 
 BASIC_IDL_TYPES_TO_C = {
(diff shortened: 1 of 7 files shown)


dirk-thomas:
After all the issues in the pipeline have been addressed the second commit in this PR fixes the actual problem from ros2/rosidl#388 and prepends a BOM to generated C/C++ if necessary: [Build Status](https://ci.ros2.org/job/ci_windows/7470/)

rosidl_generator_c/rosidl_generator_c/__init__.py
```diff
@@ -34,7 +34,20 @@ def generate_c(generator_arguments_file):
...
+    generate_files(
+        generator_arguments_file, mapping,
+        post_process_callback=prefix_with_bom_if_necessary)
+
+
+def prefix_with_bom_if_necessary(content):
+    try:
+        content.encode('ASCII')
+    except UnicodeError:
+        prefix = '\ufeff' + \
```
sloretz: [2.13 Page 58 (pdf page 52)](https://www.unicode.org/versions/Unicode5.0.0/ch02.pdf) says the utf-8 BOM is `0xEF, 0xBB, 0xBF`. This looks like the UTF-16 BOM.

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
```diff
@@ -34,7 +34,20 @@ def generate_cpp(generator_arguments_file):
...
+    generate_files(
+        generator_arguments_file, mapping,
+        post_process_callback=prefix_with_bom_if_necessary)
+
+
+def prefix_with_bom_if_necessary(content):
+    try:
+        content.encode('ASCII')
+    except UnicodeError:
+        prefix = '\ufeff' + \
```
sloretz: same comment here

rosidl_generator_c/rosidl_generator_c/__init__.py
dirk-thomas: I just replicated what we already use in the only existing file with a BOM: https://github.com/ros2/rcl_interfaces/blob/75e94e1d10399fda56ccb7c635aef730e3d7def0/test_msgs/include/test_msgs/message_fixtures.hpp#L1

`hexdump -n 3 -C <filename>` returns `ef bb bf`

sloretz: Ah oops, I get it. `\ufeff` is the code point, and `ef bb bf` are the encoded bytes. 

```python
>>> u"\ufeff".encode('utf-8')
'\xef\xbb\xbf'
```

Issue #386, opened 2019-06-19
Cannot create service with same array member in both request and response


jdlangs:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - Dashing from binaries
- Version or commit hash:
  - `ros-dashing-rosidl-typesupport-introspection-c     0.7.4-1bionic.20190612.174026`
- DDS implementation:
  - FastRTPS
- Client library (if applicable):
  - N/A

#### Steps to reproduce issue

Define a custom service type that has a member in the request and response with the same field name and non-primitive array type.

E.g.,
```
geometry_msgs/Point[] points
---
geometry_msgs/Point[] points
```

#### Expected behavior
Successful creation of the custom service type

#### Actual behavior
Failure to compile the generated code:
```
/home/jlangsfeld/Workspaces/ros2/build/srv_pkg/rosidl_typesupport_introspection_c/srv_pkg/srv/sample_service__type_support.c:136:8: error: redefinition of ‘srv_pkg__srv__sample_service__rosidl_typesupport_introspection_c__size_function__Point__points’
 size_t srv_pkg__srv__sample_service__rosidl_typesupport_introspection_c__size_function__Point__points(
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/jlangsfeld/Workspaces/ros2/build/srv_pkg/rosidl_typesupport_introspection_c/srv_pkg/srv/sample_service__type_support.c:25:8: note: previous definition of ‘srv_pkg__srv__sample_service__rosidl_typesupport_introspection_c__size_function__Point__points’ was here
 size_t srv_pkg__srv__sample_service__rosidl_typesupport_introspection_c__size_function__Point__points(
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/jlangsfeld/Workspaces/ros2/build/srv_pkg/rosidl_typesupport_introspection_c/srv_pkg/srv/sample_service__type_support.c:144:14: error: redefinition of ‘srv_pkg__srv__sample_service__rosidl_typesupport_introspection_c__get_const_function__Point__points’
[...]

PR #387, opened 2019-07-12
[rosidl_typesupport_introspection_c] Use message namespaced type name as function prefix

jacobperron:
Fixes https://github.com/ros2/rosidl/issues/386

This is the same prefix that is already applied to C++ typesupport functions.

PR #392, opened 2019-07-22
[backport dashing] [rosidl_typesupport_introspection_c] Use message namespaced type name as function prefix

jacobperron:
Backport #387 

Fixes https://github.com/ros2/rosidl/issues/386

This is the same prefix that is already applied to C++ typesupport functions.


dirk-thomas:
Please use something like "[backport dashing]" in the title.

PR #393, opened 2019-07-23
fix error msg asserts due to change in pytest

wjwwood:
This is causing failures on nightly, but overlapped with fixes for the connect pr so I rolled them together.

Connects to https://github.com/ros2/launch/pull/277


jacobperron:
Looks like this should be backported to Dashing (see test failures: https://github.com/ros2/ros2/pull/754#issuecomment-518649968).

PR #396, opened 2019-07-24
Set _FOUND to trick ament_target_dependencies() for test

sloretz:
blocks ament/ament_cmake#180

### DIFF
--- a/rosidl_generator_c/CMakeLists.txt
+++ b/rosidl_generator_c/CMakeLists.txt
@@ -94,6 +94,7 @@ if(BUILD_TESTING)
   )
 
   include(cmake/register_c.cmake)
+  # Trick ament_target_dependencies() into thinking this package has been found
   set(rosidl_generator_c_FOUND "1")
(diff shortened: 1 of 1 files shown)


sloretz:
> What call to ament_target_dependencies is causing the issue?

@jacobperron this invokes the generator for the test messages, including the generator in this package.

https://github.com/ros2/rosidl/blob/c0cb4997bffce51dc49325294ecd5dd89a12771f/rosidl_generator_c/CMakeLists.txt#L107

so this is called,

https://github.com/ros2/rosidl/blob/c0cb4997bffce51dc49325294ecd5dd89a12771f/rosidl_generator_c/cmake/rosidl_generator_c_generate_interfaces.cmake#L127-L129

 but `rosidl_generator_c` is still being built so it can't have been `find_package()`'d. These variables are tricking `ament_target_dependencies()` into thinking it was `find_package()`'d anyways.

https://github.com/ros2/rosidl/blob/c0cb4997bffce51dc49325294ecd5dd89a12771f/rosidl_generator_c/CMakeLists.txt#L97-L99

Issue #272, opened 2018-04-25
Using C / C++ keywords as package name / field names results into compilation error


firesurfer:
I realised today that using C++ keywords as fieldnames results into a compilation error.
Example:

```
bool delete false
```
Will result into:
```
rosidl_generator_cpp/iboss_msgs/msg/request_stored_data__struct.hpp:90:16: error: expected member name or ';' after declaration specifiers
  _delete_type delete;
```

It would be great if either the message generator would throw an error like: Don't use c++ keywords or 
the message generator would somehow escape these fields.
If this is not a viable solution for you I think a hint in the documentation would be useful:
https://github.com/ros2/ros2/wiki/About-ROS-Interfaces

dirk-thomas:
> the message generator would somehow escape these fields.

Absolutely, that is what should happen in each language message generator specific to the keywords in that language.

Related to #118.

firesurfer:
Just for information: I did a short search and apparently it isn't possible in C++ to escape keywords. The usage of keywords as identifiers is forbidden.

dirk-thomas:
> it isn't possible in C++ to escape keywords

I am not sure I understand your comment. If the `delete` field would result in a C++ member `delete_` it doesn't collide with the keyword anymore. This step is commonly called "mangling" but I assumed what is what you referred to as "escaping".

firesurfer:
I thought of escaping in the way of doing someting like ` int @class = 0` what would be possible in C#. 
Changing a field named `delete` to  `delete_` it would work of course in C++.

tfoote:
I've started an design article on this topic at: https://github.com/ros2/design/pull/172

firesurfer:
I guess this can be closed @dirk-thomas

Rockjack00:
I see that this issue was closed a few years ago, but I am running into the same problem.  I don't see a change to the ROS 2 docs about [field names](https://docs.ros.org/en/rolling/Concepts/Basic/About-Interfaces.html#field-names) in interfaces.

> I've started an design article on this topic at: [ros2/design#172](https://github.com/ros2/design/pull/172)

It seems like mangling the fields was discussed here but never implemented.

Was this issue ignored or should it be reopened?

Issue #118, opened 2016-04-18
Consider narrowing the scope of `#undef ERROR` in generated message code


gerkey:
https://github.com/ros2/rosidl/commit/d4264b70387e91b8f46a40ff307881f28ff07c04#commitcomment-17146232

mikaelarguedas:
This is a known problem that we face is ROS1 as well (e.g. https://github.com/ros/geometry2/issues/172) and it would be great to see how we can address these collisions with Windows defined in the near future

dejanpan:
This issue has come up again here:

The generated file build\tf2_msgs\rosidl_generator_cpp\tf2_msgs\msg\tf2_error__struct.hpp contains the following code that conflicts with Windows' definition:
```
  // constants
  enum { NO_ERROR = 0u };
```

Windows SDK defines macro NO_ERROR in Error.h:
```
#define NO_ERROR            0
```

We have confirmed for myself that that culprit is Windows SDK. On a clean installation everything works OK but as soon as we installed Windows SDK the build fails in TF2 (now it fails in Release build too).

@wjwwood says:
`It might also be possible to change these constants in ROS 2 (with appropriate special handling in the ros1_bridge).`

tfoote:
This is also an issue with 'DELETE' defined in winnt.h: https://github.com/ros2/common_interfaces/issues/44

We'll either need some way to generically protect ourselves from some known set of conflicting strings or else have a blacklist of known strings that must be banned due to windows bad hygene in macro names. 

Might it make sense to have a known set of potentially colliding macros that automatically get undefed in our code generators if a conflict is detected?

dirk-thomas:
Options:
* undefine Windows specific MACRO
* name mangle the enum (e.g. appending an underscore)
* use different enum naming scheme (e.g. `kERROR`)

tfoote:
Related: https://github.com/ros2/design/pull/172

PR #401, opened 2019-08-23
Guard against DELETE being predefined on Windows

jacobperron:
Motivated by build errors encountered building visualization_msgs/msg/Marker.msg.

### DIFF
--- a/rosidl_generator_cpp/resource/msg__struct.hpp.em
+++ b/rosidl_generator_cpp/resource/msg__struct.hpp.em
@@ -1,18 +1,4 @@
 @# Included from rosidl_generator_cpp/resource/idl__struct.hpp.em
-// Protect against common names being predefined on Windows, in case somebody defines
-// constants by these names.
-#if defined(_WIN32)
-  #if defined(ERROR)
-    #undef ERROR
-  #endif
-  #if defined(NO_ERROR)
-    #undef NO_ERROR
-  #endif
-  #if defined(DELETE)
-    #undef DELETE
-  #endif
-#endif
-@
 @{
 from rosidl_generator_cpp import create_init_alloc_and_member_lists
 from rosidl_generator_cpp import escape_string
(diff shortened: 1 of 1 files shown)


jacobperron:
CI before and after: https://github.com/ros2/common_interfaces/pull/70#issuecomment-524122806

dirk-thomas:
Instead of blindly undefining these names (even when the message doesn't use those names) I would prefer is this only happens in the case that the message does contain constants with that name. (Also see #118 and #272.)

jacobperron:
I've updated the PR so that we only undefine if a constant is using the same name.
Also, the original value of the macro is restored after defining the constant.

Ultimately, name mangling as described in https://github.com/ros2/design/pull/172 is probably the way to go, but in order to unblock https://github.com/ros2/common_interfaces/pull/70 we have this solution for now.

Windows CI:
- https://github.com/ros2/common_interfaces/pull/70 [Build Status](https://ci.ros2.org/job/ci_windows/7784/)
- https://github.com/ros2/common_interfaces/pull/70 + this PR [Build Status](https://ci.ros2.org/job/ci_windows/7897/) (warning unrelated)

dirk-thomas:
> the original value of the macro is restored after defining the constant.

Is it still possible for code to use the constants? Is the current Jenkins build only testing if the generated code compiles or also if these constants can be used from other packages?

jacobperron:
> Is it still possible for code to use the constants?

Good question. Though I built and tested https://github.com/ros-visualization/interactive_markers/pull/44 as part of the job, it doesn't actually use the constant. I'll try triggering another job that includes code that actually uses one of the constants that are affected by this change.

jacobperron:
@dirk-thomas Looks like we can use the constants still. This build exercises `tf2`, which uses a constant with name `NO_ERROR`, and my `rviz2` branch, which uses a constant with name `DELETE`: [Build Status](https://ci.ros2.org/job/ci_windows/7903/) (warnings are unrelated)

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -31,6 +20,9 @@ from rosidl_parser.definition import OCTET_TYPE
 from rosidl_parser.definition import UNSIGNED_INTEGER_TYPES
 
 message_typename = '::'.join(message.structure.namespaced_type.namespaced_name())
+
+# Common Windows macros that may interfere with user defined constants
+msvc_common_macros = ('ERROR', 'NO_ERROR', 'DELETE')
```
dirk-thomas: Nipick: alphabetical order.

jacobperron: 9b902a8

PR #406, opened 2019-09-05
rosidl_adapter: struct.idl template: generate a verbatim for each comment line instead of joining the comments with nl

TSC21:
This aims to fix https://github.com/eProsima/IDL-Parser/issues/33, where the parser was breaking when finding a new line on the verbatim but without a backlash signalling.

The proposed solution adds a `@verbatim` line for each comment line, instead of joining each comment by new line. I did try another solution by joining the lines using ` \\ \n`, but that was resulting in a warning on the IDL Parser, which I avoided.

@LuisGP @dirk-thomas looking for your feedback and review. Thanks!

### DIFF
--- a/rosidl_adapter/rosidl_adapter/resource/struct.idl.em
+++ b/rosidl_adapter/rosidl_adapter/resource/struct.idl.em
@@ -46,7 +46,9 @@ else:
 @[end if]@
 @#
 @[if msg.annotations.get('comment', [])]@
-    @@verbatim (language="comment", text=@(string_to_idl_string_literal('\n'.join(msg.annotations['comment']))))
+@[    for comment in msg.annotations['comment']]@
+    @@verbatim (language="comment", text=@(string_to_idl_string_literal(comment)))
+@[    end for]@
 @[end if]@
     struct @(msg.msg_name) {
 @[if msg.fields]@
@@ -55,7 +57,9 @@ else:
 
 @[end if]@
 @[    if field.annotations.get('comment', [])]@
-      @@verbatim (language="comment", text=@(string_to_idl_string_literal('\n'.join(field.annotations['comment']))))
+@[        for comment in field.annotations['comment']]@
+      @@verbatim (language="comment", text=@(string_to_idl_string_literal(comment)))
+@[        end for]@
 @[    end if]@
 @[    if field.default_value is not None]@
       @@default (value=@(to_idl_literal(get_idl_type(field.type), field.default_value)))


LuisGP:
LGTM

TSC21:
@wjwwood @nuclearsandwich could I get your review here. Thank you!

dirk-thomas:
Instead of splitting multi line comments into multiple annotations I would rather keep a single annotation with the `text` attribute containing valid newline characters according to the IDL spec.

If the currently generated output isn't compliant to how the IDL spec expects the multi line text to be formatted / escaped please let me know how it should be represented and I am happy to update the template accordingly.

nuclearsandwich:
> @nuclearsandwich could I get your review here. Thank you!

I defer to @dirk-thomas's experience with IDL for this one.

TSC21:
> Instead of splitting multi line comments into multiple annotations I would rather keep a single annotation with the text attribute containing valid newline characters according to the IDL spec.

That was my preference as well but as you can read on the description, adding `\` to the end of the line is not well handled by the IDL Parser. How do you suggest this to be done? Thanks!

dirk-thomas:
If the currently generate idl files are spec-compliant and some existing idl parsers can correctly parse them maybe that specific idl parser needs to be fixed?

TSC21:
> If the currently generate idl files are spec-compliant and some existing idl parsers can correctly parse them maybe that specific idl parser needs to be fixed?

@LuisGP your call here

LuisGP:
> That was my preference as well but as you can read on the description, adding `\` to the end of the line is not well handled by the IDL Parser. How do you suggest this to be done? Thanks!

That's not the case. The IDL Parser parses correctly terminating lines with `\`, but seems that there exists a blank space between the backslash and the end of line, so the warning is correct too, the backslash is escaping the blank space, not the end of line.

Another question is if a multiline comment (or a String Literal, more generally) can be defined without adding backslashes at the end of each line or it should be defined following a C like format.

TSC21:
> > That was my preference as well but as you can read on the description, adding `\` to the end of the line is not well handled by the IDL Parser. How do you suggest this to be done? Thanks!
> 
> That's not the case. The IDL Parser parses correctly terminating lines with `\`, but seems that there exists a blank space between the backslash and the end of line, so the warning is correct too, the backslash is escaping the blank space, not the end of line.
> 
> Another question is if a multiline comment (or a String Literal, more generally) can be defined without adding backslashes at the end of each line or it should be defined following a C like format.

@LuisGP so what's the exact solution here? Using a blank space between the backslash and the end of line was the only way I was seeing of adding the backslash to the end of the line in Python, but it's throwing the warning as you know. This needs to be solved because your IDL Parser currently is not able to correctly parse multiline comments.

LuisGP:
@TSC21, @dirk-thomas  
After a deep reading of the standard and examination of the [lex files provided by CORBA](https://www.omg.org/attachments/scanner.ll), we concluded that a `string_literal` cannot contain raw newline characters. So, if you want a multiline comment the resulting IDL code must be in one of the following forms:

    @verbatim (language="comment", text=" Optical flow in XYZ body frame in SI units.\n Check http://en.wikipedia.org/wiki/International_System_of_Units")

or

    @verbatim (language="comment", text=" Optical flow in XYZ body frame in SI units.")
    @verbatim (language="comment", text=" Check http://en.wikipedia.org/wiki/International_System_of_Units")

So, escaping the newline character will produce a singled-line comment (that maybe is useful too under some circumstances), but we think that a multi-line string literal isn't spec-compliant.

I personally think that the second approach, each comment line in their own `@verbarim` annotation produces a clearer IDL code.

dirk-thomas:
From reading the IDL spec it sounds right that newlines should be escaped:

<details>
<summary>IDL excerpts</summary>

> 7.4.15 Building Block Annotations

Rule 227 defines the parameter value of an annotation as a `<const_expr>`. And the `<const_expr>` boils down to a `<string_literal>` in this case. String literals are defined in 7.2.6.3 as:

> Strings are null-terminated sequences of characters. Strings are of type string if they are made of non-wide characters.

> A string literal is a sequence of character literals (as defined in 7.2.6.2), with the exception of the character with numeric value 0, surrounded by double quotes.

> Both wide and non-wide string literals must be specified using characters from the ISO Latin-1 (8859-1) character set. A string literal shall not contain the character ‘\0’ .

And under 7.2.6.2.2 Escape Sequences to Represent Character Literals it states the following:

> Nongraphic characters shall be represented using escape sequences as defined below in Table 7-9. Note that escape sequences shall be used to represent single quote and backslash characters in character literals.

Table 7-9 includes `newline: \n` (beside several other).

</details>

Regarding the use of multiple annotations of the same type (implying that they should be be concatenated) I couldn't find any evidence for that in the IDL spec. Can you point to any reference why you think that is the specified behavior?

Especially in the context of annotations possibly having default values I would think that the last annotation of the same type could also override the previous ones.

Though a sentence in section 7.2.6.3 indicates that there is another alternative to single super long lines:

> Adjacent string literals are concatenated.

Base on that the following should be viable:

```
@verbatim (language="comment", text=
  "Optical flow in XYZ body frame in SI units.\n"
  "Check http://en.wikipedia.org/wiki/International_System_of_Units")
```

LuisGP:
> Regarding the use of multiple annotations of the same type (implying that they should be be concatenated) I couldn't find any evidence for that in the IDL spec. Can you point to any reference why you think that is the specified behavior?

I don't mean I think it is the specific behavior but my preference. I prefer that form to avoid very long lines.

> Though a sentence in section 7.2.6.3 indicates that there is another alternative to single super long lines:
> 
> > Adjacent string literals are concatenated.
> 
> Base on that the following should be viable:
> 
> ```
> @verbatim (language="comment", text=
>   "Optical flow in XYZ body frame in SI units.\n"
>   "Check http://en.wikipedia.org/wiki/International_System_of_Units")
> ```

But this solution seems to be better in all aspects. I tested with Fast-RTPS-Gen and doesn't work, but should be quick to fix.

dirk-thomas:
Please see #410 which updates the IDL grammar file to support adjacent string literals as well as the idl templates to use adjacent strings instead of embedding a newline character into the string literal.

PR #410, opened 2019-09-10
support adjacent string literals, use them for multi-line comments

dirk-thomas:
Replaces #406.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8101/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4100/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6602/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7979/)

PR #411, opened 2019-09-10
avoid zero length comment when the comment only contains a unit

dirk-thomas:
Notices while working on #410.

Not planning to run CI on this trivial swap.


TSC21:
@dirk-thomas, @jacobperron is it possible to backport this feature to Dashing? Thank you!

PR #412, opened 2019-09-13
Add is_message trait in support of tf2 conversions

mjcarroll:
Restores the `IsMessage` trait from ROS1 `message_traits`.

This will simply be a true value for things that are messages and false otherwise.

ROS1 Ref: https://github.com/ros/roscpp_core/blob/kinetic-devel/roscpp_traits/include/ros/message_traits.h#L111

Signed-off-by: Michael Carroll <michael@openrobotics.org>


dirk-thomas:
It would be good to add similar traits for services and actions.

rosidl_generator_cpp/resource/action__traits.hpp.em
```diff
@@ -0,0 +1,85 @@
+@# Included from rosidl_generator_cpp/resource/idl__traits.hpp.em
```
dirk-thomas: Does the `resource/idl__traits.hpp.em` template need to be updated in order to include this new file?

Also the test doesn't cover actions yet which based on the above seems to be broken.

mjcarroll: It does.

Also, because action generation depends on `action_msgs`, it's likely going to have to reside in a higher package.  I discussed with @jacobperron, and I'm going to move forward with putting the trait tests in `rclcpp_action`, unless there is another alternative.

jacobperron: We can't build test action interfaces in `rosidl_generator_cpp` because of a circular dependency on `action_msgs`. So unless we can test `is_action` without generating an action, then until we remove the circular dependency we won't be able to add tests to this package.

rosidl_generator_cpp/test/test_traits.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+// Empty testing struct, with template instantiation
+struct Message2 {};
+
+template<>
+struct rosidl_generator_traits::is_message<Message2>: std::true_type {};
+
+TEST(Test_rosidl_generator_traits, is_message) {
+  // A message is not a service
+  ASSERT_TRUE(is_message<rosidl_generator_cpp::msg::Empty>());
```
jacobperron: nit: using `EXPECT_*` instead, we'd see multiple failures instead of just the first to fail.

rosidl_generator_cpp/test/test_traits.cpp
```diff
@@ -0,0 +1,76 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+#include "rosidl_generator_cpp/msg/empty.hpp"
+#include "rosidl_generator_cpp/msg/string.hpp"
+#include "rosidl_generator_cpp/srv/complex_type_zero_fill.hpp"
```
jacobperron: I just noticed that these tests need to be updated after rebasing. Since https://github.com/ros2/rosidl/pull/397, we are using a different set of test interfaces, ie. `ComplexTypeZeroFill.msg` and `String.msg` do not exist.

mjcarroll: Rebased and updated in 4fc81a9

PR #420, opened 2019-11-07
better memory handling for sequences

rotu:
1) don't leak memory if resized to 0
2) realloc instead of dealloc/malloc

### DIFF
--- a/rosidl_generator_c/src/primitives_sequence_functions.c
+++ b/rosidl_generator_c/src/primitives_sequence_functions.c
@@ -25,12 +25,9 @@
     if (!sequence) { \
       return false; \
     } \
-    TYPE_NAME * data = NULL; \
-    if (size) { \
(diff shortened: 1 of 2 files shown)


mjcarroll:
Signoff for DCO please!

rotu:
@mjcarroll DCO'd

dirk-thomas:
I would assume the same applies to other C functions ending in `_init()`. They are simply not supposed to be called again (without `fini` in between).

If we want to support this use case we should probably update all similar functions as well as describe this in the docblock of the API.

mjcarroll:
@rotu Do you have a chance to address the feedback from @dirk-thomas?  It would be ideal to get this updated across the board.

rotu:
Closing this PR. I’m not sure realloc is a good idea since it doesn’t guarantee memory alignment like malloc does

Issue #422, opened 2019-11-20
[rosidl_generator_c] double free issue when initialization is failed


DongheeYe:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04 
- Installation type:
  - source 
- Version or commit hash:
  - 0.8.1 (36fac1e367bd98a493a7a0935b2c7b5ae86f5e7d)
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
``` sh
colcon build --symlink-install --merge-install 
```

#### Expected behavior
There MUST be no double free.
For example, build/std_msgs/rosidl_generator_c/std_msgs/msg/string__functions.c
```code 
// generated code
std_msgs__msg__String *
std_msgs__msg__String__create()                          // (1)
{
  std_msgs__msg__String * msg = (std_msgs__msg__String *)malloc(sizeof(std_msgs__msg__String));
  if (!msg) {
    return NULL;
  }
  memset(msg, 0, sizeof(std_msgs__msg__String));
  bool success = std_msgs__msg__String__init(msg);        // (2)
  if (!success) {                                         // (6) success is false
    free(msg);                                            // (7) double free msg
    return NULL;
  }
  return msg;
}


bool
std_msgs__msg__String__init(std_msgs__msg__String * msg)
{
  if (!msg) {
    return false;
  }
  // data
  if (!rosidl_generator_c__String__init(&msg->data)) {      // (3) return false
    std_msgs__msg__String__destroy(msg);                    // (4) call destroy
    return false;
  }
  return true;
}

void
std_msgs__msg__String__destroy(std_msgs__msg__String * msg)
{
  if (msg) {
    std_msgs__msg__String__fini(msg);
  }
  free(msg);                                                 // (5) free msg
}

```

#### Actual behavior
There is double free when malloc is failed in rosidl_generator_c__String__init.

#### Additional information
I thought if initialization is failed, then just should call fini.

dirk-thomas:
Can you please share some steps to reproduce the double free.

DongheeYe:
Please refer to the number comment of code.

(1) User calls std_msgs__msg__String__create. 
(2) std_msgs__msg__String__init will be called.
(3) rosidl_generator_c__String__init can return false. (Refer to the below code)
(4) Then, std_msgs__msg__String__destroy will be call.
(5) msg object will be freed in std_msgs__msg__String__destroy and return false.
[...]

PR #423, opened 2019-11-20
Fix double free issue when initialization is failed

DongheeYe:
Fixes https://github.com/ros2/rosidl/issues/422

### DIFF
--- a/rosidl_generator_c/resource/msg__functions.c.em
+++ b/rosidl_generator_c/resource/msg__functions.c.em
@@ -105,7 +105,7 @@ for member in message.structure.members:
             # initialize each array element
             lines.append('for (size_t i = 0; i < %d; ++i) {' % member.type.size)
             lines.append('  if (!%s__init(&msg->%s[i])) {' % (basetype_to_c(member.type.value_type), member.name))
-            lines.append('    %s__destroy(msg);' % message_typename)
+            lines.append('    %s__fini(msg);' % message_typename)
             lines.append('    return false;')
             lines.append('  }')
             lines.append('}')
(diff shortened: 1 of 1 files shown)


allenh1:
This looks identical to #300. @clalancette do you know if there was a reason to change it to `destroy` instead of `fini`? Maybe a comment should be left above them so that this doesn't change back to `destroy` again?

I originally wrote `destroy` on those lines, but it should definitely be `fini`.

clalancette:
> This looks identical to #300. @clalancette do you know if there was a reason to change it to `destroy` instead of `fini`? Maybe a comment should be left above them so that this doesn't change back to `destroy` again?
> 
> I originally wrote `destroy` on those lines, but it should definitely be `fini`.

For what it is worth, this didn't get changed back; the code that I fixed back in 2018 is still using `fini`.  In the meantime, new code got added that uses `destroy` instead of `fini` (commit 416ebd84a678178f8e8941a1961af8abe7514251, it looks like, which was all of the IDL refactoring, so its not surprising).

I guess we could add a comment, but it isn't clear to me where we would add it so that people would see it.

clalancette:
Seems reasonable to me.  I'll kick-off CI on this, running the communication tests and some message packages just to make sure things generate properly.

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8788/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4609/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7200/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8707/)

clalancette:
All right, this is looking good to me, so I'm going to go ahead and merge.  Thanks for the contribution!

jpsamper2009:
@dirk-thomas We could cherry-pick this fix into the dashing branch? Thanks!

clalancette:
I've opened up https://github.com/ros2/rosidl/pull/426 and https://github.com/ros2/rosidl/pull/427 to backport this to Dashing and Eloquent.

PR #427, opened 2019-12-11
[Eloquent backport]: Fix double free issue when initialization is failed (#423)

clalancette:
Backport of #423.


dirk-thomas:
Since this was the only change between `master` and `eloquent` I fast-forwarded the existing commit from the `master` branch to the `eloquent` branch instead.

PR #442, opened 2020-02-26
Splitted rosidl_generator_c and rosidl_generator_cpp in two: rosidl_generator_x and rosidl_runtime_x

ahcorde:
As discussed in the issue https://github.com/ros2/ros2/issues/442 I have spplited in to `rosidl_generator_c`and `rosidl_generator_cpp`.

I don't really know if `rosidl_runtime_cpp` is needed. because this package is not creating any runtime dependency such as a .so/dylib/dll. But I have seen this comment https://github.com/ros2/rosidl/blob/master/rosidl_generator_cpp/include/rosidl_generator_cpp/message_initialization.hpp#L18 from @clalancette and I dediced to splitt it too, also because I think it makes sense for consistency.


rosidl_typesupport_introspection_c/resource/msg__type_support.c.em
```diff
@@ -62,7 +62,7 @@ for member in message.structure.members:
         member_names.append(member.name)
     elif isinstance(type_, AbstractWString):
         member_names = includes.setdefault(
-            'rosidl_generator_c/u16string_functions.h', [])
+            'rosidl_generator_c/string_functions.h', [])
```
dirk-thomas: Why should this change?

rosidl_generator_c/package.xml
```diff
@@ -7,15 +7,14 @@
   <maintainer email="william@osrfoundation.org">William Woodall</maintainer>
   <license>Apache License 2.0</license>
 
+  <buildtool_depend>ament_cmake_python</buildtool_depend>
   <buildtool_depend>ament_cmake_ros</buildtool_depend>
-
-  <build_depend>rosidl_typesupport_interface</build_depend>
+
```
dirk-thomas: Nitpick: trailing whitespace.

rosidl_generator_c/CMakeLists.txt
```diff
@@ -13,33 +13,10 @@ endif()
...
-    COMPILE_OPTIONS -Wall -Wextra -Wpedantic)
-endif()
-if(WIN32)
-  target_compile_definitions(${PROJECT_NAME}
-    PRIVATE "ROSIDL_GENERATOR_C_BUILDING_DLL")
-endif()
+find_package(rosidl_runtime_c REQUIRED)
 
 ament_export_dependencies(rosidl_cmake)
 ament_export_dependencies(rosidl_typesupport_interface)
```
dirk-thomas: Is this still necessary - since the dependency has been removed from the manifest?

rosidl_generator_c/resource/action__type_support.h.em
```diff
@@ -11,7 +11,7 @@
 // Forward declare the get type support functions for this type.
 ROSIDL_GENERATOR_C_PUBLIC_@(package_name)
 const rosidl_action_type_support_t *
-  ROSIDL_TYPESUPPORT_INTERFACE__ACTION_SYMBOL_NAME(
+ROSIDL_TYPESUPPORT_INTERFACE__ACTION_SYMBOL_NAME(
```
dirk-thomas: This change shouldn't be part of the PR since the change is already present on the target branch. This probably needs to be rebased correctly.

Same below.

rosidl_generator_c/test/test_interfaces.c
```diff
@@ -23,7 +23,7 @@
 
 #include "rosidl_generator_c/primitives_sequence_functions.h"
 #include "rosidl_generator_c/string_functions.h"
-#include "rosidl_generator_c/u16string_functions.h"
+#include "rosidl_generator_c/string_functions.h"
```
dirk-thomas: Why should this change?

Same below.

rosidl_generator_cpp/CMakeLists.txt
```diff
@@ -2,6 +2,15 @@ cmake_minimum_required(VERSION 3.5)
 
 project(rosidl_generator_cpp)
 
+# Default to C11
+if(NOT CMAKE_C_STANDARD)
+  set(CMAKE_C_STANDARD 11)
+endif()
```
dirk-thomas: Why was this added? The package seems to only contain C++ code.

rosidl_runtime_cpp/package.xml
```diff
@@ -0,0 +1,25 @@
...
+  <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
+  <license>Apache License 2.0</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <buildtool_export_depend>ament_cmake</buildtool_export_depend>
+
+  <exec_depend>rosidl_parser</exec_depend>
+
+  <test_depend>ament_cmake_gtest</test_depend>
```
dirk-thomas: The package doesn't have any gtests?

rosidl_runtime_cpp/package.xml
```diff
@@ -0,0 +1,25 @@
...
+  <version>0.8.2</version>
+  <description>Generate the ROS interfaces in C++.</description>
+  <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
+  <license>Apache License 2.0</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <buildtool_export_depend>ament_cmake</buildtool_export_depend>
+
+  <exec_depend>rosidl_parser</exec_depend>
```
dirk-thomas: Wasn't the goal for this package to *not* depend on Python stuff? Also the package shouldn't need this.

rosidl_generator_cpp/CMakeLists.txt
ahcorde: we are compiling the test maybe include this inside `if(BUILD_TESTING) ... endif()` ?

rosidl_generator_c/CMakeLists.txt
ahcorde: tests need to link against this dependency

rosidl_generator_cpp/CMakeLists.txt
dirk-thomas: The tests have been built before this patch and are C++. So why would it be necessary now?

rosidl_generator_c/CMakeLists.txt
dirk-thomas: Then the dependency can't be removed from the manifest.

dirk-thomas: Also this line only exports the package to downstream packages. It doesn't make it available to the tests in this repo.

ahcorde: Restored

dirk-thomas: The commit 83e43928555bdb8300bf29dde5b5ac25cb8eeea0 is incorrect. `rosidl_typesupport_interface` is not being used at build time of the package but only exported to downstream packages. So the `build_depend` should be removed but the `build_export_depend` must be restored.

rosidl_generator_c/resource/action__type_support.h.em
dirk-thomas: The same comment still applies to the message / service macros.

dirk-thomas:
Commonly a package only provides headers with a namespace / directory matching the package name. Should we consider updating the namespace / directory of the moved headers? If yes, how much code would be affected by this breaking change?

ahcorde:
I was trying to do so, but this will require a lot of changes in many packages. I was trying to reduce the amount of noise regarding this change.

Also the resources are in the generator but the headers are in the runtime. This will require also some changes in the resources. 

but if this is the way I can put some time on it

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9511/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5229/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7774/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9445/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/270/)

rosidl_generator_cpp/CMakeLists.txt
```diff
@@ -45,32 +43,65 @@ if(BUILD_TESTING)
...
-
   ament_add_gtest(test_bounded_vector test/test_bounded_vector.cpp)
+  if(TARGET test_bounded_vector)
+    add_dependencies(test_bounded_vector ${PROJECT_NAME})
+    target_include_directories(test_bounded_vector PUBLIC
+      ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME})
+    ament_target_dependencies(test_bounded_vector
+      rosidl_runtime_cpp
+      rosidl_runtime_c)
+  endif()
```
dirk-thomas: Why are all these extra lines for each test added? There weren't present before...

ahcorde: This was done before with the `include_directories` because all headers were located in the generator packages, I can include this with `target_include_directories`, but they are ROS 2 dependencies so I think `ament_target_dependencies` is the right choice.

ahcorde:
Removed `rosidl_generator_c` and `rosidl_generator_cpp` from `rosidl_typesupport_introspection_c` and `rosidl_typesupport_introspection_cpp` because these packages are not generators and they are not making use of the CMakeLists macros

rosidl_typesupport_introspection_c/CMakeLists.txt
```diff
@@ -15,7 +15,8 @@ find_package(ament_cmake_python REQUIRED)
 find_package(ament_cmake_ros REQUIRED)
 
 ament_export_dependencies(rosidl_cmake)
-ament_export_dependencies(rosidl_generator_c)
+# ament_export_dependencies(rosidl_generator_c)
```
dirk-thomas: Please remove this line rather than commenting it out.

Same below.

ahcorde: Removed comments

rosidl_generator_c/CMakeLists.txt
```diff
@@ -2,44 +2,12 @@ cmake_minimum_required(VERSION 3.5)
...
 project(rosidl_generator_c C)
 
-# Default to C11
-if(NOT CMAKE_C_STANDARD)
-  set(CMAKE_C_STANDARD 11)
-endif()
-
-if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-  add_compile_options(-Wall -Wextra -Wpedantic)
-endif()
```
dirk-thomas: For the remaining tests we still want to keep the C standard and compiler flags I would assume.

ahcorde: Restored https://github.com/ros2/rosidl/pull/442/commits/e3fe8789ba6237adb9da5e347e0ed90b3c1a7448

rosidl_runtime_c/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+)
+ament_target_dependencies(${PROJECT_NAME}
+  "rosidl_typesupport_interface")
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  set_target_properties(${PROJECT_NAME} PROPERTIES
+    COMPILE_OPTIONS -Wall -Wextra -Wpedantic)
+endif()
+if(WIN32)
+  target_compile_definitions(${PROJECT_NAME}
+  PRIVATE "ROSIDL_GENERATOR_C_BUILDING_DLL")
```
dirk-thomas, line 34: Needs one more level of indentation.

ahcorde: fixed https://github.com/ros2/rosidl/pull/442/commits/132d521a4496eef109762018749eb1f717125ed9

rotu:
This breaks binary compatibility in its public API. Shouldn't it bump the major version number?

wjwwood:
> This breaks binary compatibility in its public API. Shouldn't it bump the major version number?

We will, I assume, when we next release from master.

dirk-thomas:
> This breaks binary compatibility in its public API. Shouldn't it bump the major version number?

Since the packages in this repo have 0.x versions there is no need to bump the major version. Anything can change anytime in pre 1.x versions.

Independent of this change the goal is to bring several repos / packages to at least version 1.0 for the Foxy release.

ahcorde:
- Updated all repos involved in this change. 
 - Skipping `ros1_bridge`, `qt_*`, `rqt_*`, and `rviz_*` packages. 
 - Testing against `test_rclcpp` and `rosbag2_tests`.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10031/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5626/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8161/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9938/)

 - One warning made the builds unstable but there is a ([PR pending](https://gitlab.com/micro-ROS/ros_tracing/ros2_tracing/-/merge_requests/163))

```
CMake Deprecation Warning at /home/jenkins-agent/workspace/ci_linux-aarch64/ws/install/ament_cmake_export_interfaces/share/ament_cmake_export_interfaces/cmake/ament_export_interfaces.cmake:37 (message):
ament_export_interfaces() is deprecated, use ament_export_targets() instead
CMakeLists.txt:85 (ament_export_interfaces)
``` 

In MacOS
 - projectroot.cppcheck

Windows
 - some failures in test_rclcpp

rosidl_generator_c/CMakeLists.txt
```diff
@@ -13,33 +13,10 @@ endif()
...
-  "rosidl_typesupport_interface")
-if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-  set_target_properties(${PROJECT_NAME} PROPERTIES
-    COMPILE_OPTIONS -Wall -Wextra -Wpedantic)
-endif()
-if(WIN32)
-  target_compile_definitions(${PROJECT_NAME}
-    PRIVATE "ROSIDL_GENERATOR_C_BUILDING_DLL")
-endif()
+find_package(rosidl_runtime_c REQUIRED)
```
dirk-thomas, line 16: This introduced a regression: see #503.

PR #444, opened 2020-03-06
Fix bug inserting bounded_vector in reverse order

carlossvg:
This PR fixes a bug which allowed to insert `bounded_vector` in reverse order. 

This fix was originally implemented by @sumanth-nirmal.

### DIFF
--- a/rosidl_runtime_cpp/test/test_bounded_vector.cpp
+++ b/rosidl_runtime_cpp/test/test_bounded_vector.cpp
@@ -50,3 +50,19 @@ TEST(rosidl_generator_cpp, bounded_vector_rvalue) {
   ASSERT_EQ(vv[0], 1);
   ASSERT_EQ(vv[1], 2);
 }
+
+TEST(rosidl_generator_cpp, bounded_vector_insert) {
+  rosidl_runtime_cpp::BoundedVector<int, 6> v1;
+  v1.push_back(1);
+  auto it = v1.begin();
+  rosidl_runtime_cpp::BoundedVector<int, 3> v2;
+  v2.push_back(2);
+  v2.push_back(3);
+  v2.push_back(4);
+  // insert v2 after the first position in reverse order
+  ASSERT_THROW(v1.insert(it + 1, v2.end(), v2.begin()), std::length_error);
+  // insert v2 after the 1st position
+  ASSERT_NO_THROW(v1.insert(it + 1, v2.begin(), v2.end()));
+  // insert v2 after the 4th position
+  ASSERT_THROW(v1.insert(it + 4, v2.begin(), v2.end()), std::length_error);
+}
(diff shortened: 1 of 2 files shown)


dirk-thomas:
> This PR fixes a bug which allowed to insert `bounded_vector` in reverse order.

Wouldn't that be expected behavior if you pass `__first` and `__last` with a negative distance? Why is that something which needs to be prevented?

carlossvg:
> > This PR fixes a bug which allowed to insert `bounded_vector` in reverse order.
> 
> Wouldn't that be expected behavior if you pass `__first` and `__last` with a negative distance? Why is that something which needs to be prevented?

You're right, I made a wrong assumption. I think the fix is related to comparing unsigned and signed integers. @sumanth-nirmal Could you clarify this?

carlossvg:
Sorry for the delayed response. 

The motivation for this change was to fix a lintern warning complaining about a signed and unsigned comparison. We added a static cast to std::distance and an additional condition to prevent a cast from a negative value. This last part may not be really necessary as in that case the base vector would generate the exception anyway.

dirk-thomas:
@carlossvg Please quote the exact linter warning and in which environment you see it. Also please rebase the PR.

carlossvg:
> @carlossvg Please quote the exact linter warning and in which environment you see it. Also please rebase the PR.

The error came from PClint. It says: 

- "(note: 9117) implicit conversion from 'long' to 'unsigned long' changes signedness of underlying type \[AUTOSAR Rule M5-0-4\]"

dirk-thomas:
I rebased the patch and triggered a set of CI builds testing `rosidl_runtime_cpp`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10959/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6299/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8921/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10862/)

dirk-thomas:
@carlossvg Thank you for the patch!

rosidl_runtime_cpp/test/test_bounded_vector.cpp
```diff
@@ -50,3 +50,19 @@ TEST(rosidl_generator_cpp, bounded_vector_rvalue) {
...
+  rosidl_runtime_cpp::BoundedVector<int, 3> v2;
+  v2.push_back(2);
+  v2.push_back(3);
+  v2.push_back(4);
+  // insert v2 after the first position in reverse order
+  ASSERT_THROW(v1.insert(it + 1, v2.end(), v2.begin()), std::length_error);
+  // insert v2 after the 1st position
+  ASSERT_NO_THROW(v1.insert(it + 1, v2.begin(), v2.end()));
+  // insert v2 after the 4th position
+  ASSERT_THROW(v1.insert(it + 4, v2.begin(), v2.end()), std::length_error);
```
jwakely, line 67: This is undefined behaviour, the iterator `it` was invalidated by the insert on line 65. Even if it wasn't, adding 4 to it goes past `v1.end()`, so that's undefined too.

PR #445, opened 2020-03-06
Fix move assignment of bounded vector

carlossvg:
This PR fixes the move assignment of bounded vector. `std::move` should be used instead of `std::forward` because the parameter it's not a forwarding reference but a rvalue reference of the template class `BoundedVector`.
 
This fix was originally implemented by Andreas Pasternak.


dirk-thomas:
@gonzodepedro Can you please rebase this patch and run CI.

jacobperron:
The file this patch applies to was moved in https://github.com/ros2/rosidl/pull/442

I've reapplied the patch in https://github.com/ros2/rosidl/pull/496.

Issue #447, opened 2020-03-11
Review of the rosidl_generator_cpp package


dirk-thomas:
# Comments / feedback

Please provide feedback in separate comments - one per topic - to allow hiding them when they have been resolved - to keep the discussion readable and pending items visible.
Beside any additional feedback it is also valuable to comment if you agree/disagree with the check boxes indicating proposed to-dos.

# Scope of the package

At the moment the package has two purposes:
1. provide the message generator for C++ which generates C++ code / library at the build time of each message package
2. provide the C++ headers which messages will depend on when build in downstream packages

This poses an unnecessary runtime dependency for message packages.
Therefore this package should be split into two parts.

* [x] This will be addressed in #443 which will separate the first part (the generator) from the second part (the runtime component).

# Package dependencies

* build dependency on `ament_cmake`: since it is an `ament_cmake` package

* export the build dependencies on `ament_cmake`: it registers itself as an extension

  * [x] Reducing the dependency to `ament_cmake_core` might be feasible?
    See #459.

* export the build dependencies on `rosidl_cmake`: it uses the CMake API provided by that package when generating code for message packages

* export the build dependencies on `rosidl_generator_c`: since the headers transitively include headers from that package

All of these dependencies look reasonable.

* [x] They will be split as part of #443 between the generator and runtime package

:information_source: The test dependencies are not being reviewed since they are not part of the public interface of the package.

The group membership currently lists both groups `rosidl_generator_packages` as well as `rosidl_runtime_packages`.

* [x] Those will also be split as part of #443 between the generator and runtime package

# Directories and Files

## bin

The script contains the CLI for the C++ generator.
All logic is in the Python module.

## cmake

* `register_cpp.cmake` contains a macro to register the `rosidl_generate_idl_interfaces` extension.
* `rosidl_generator_cpp_extras.cmake` sets some CMake variables an invokes the above macro to register the extension.
* `rosidl_generator_cpp_generate_interfaces.cmake` contains the CMake logic of the C++ generator extension which generates the C++ code for interfaces and installs the headers.

## include

All C++ headers end in `.hpp`.

All headers are in the subdirectories `rosidl_generator_cpp` and `rosidl_typesupport_cpp`.

* ~~The headers in `rosidl_typesupport_cpp` declare the templated function to retrieve type support handles.
      In the C generator these symbols are in the same directory (though not templated).
      Consider unifying this to make it more similar between the two generators.~~ See below comments why these headers are in a separate namespace. 

[...]

PR #448, opened 2020-03-24
Add a utility for rigorously initializing a message instance

mxgrey:
This implements the feature request made [here](https://github.com/ros2/rosidl/issues/447#issuecomment-598003951).

I still have to write tests, which I assume should go in [here](https://github.com/ros2/rosidl/blob/d7078e3be48ade33a0739d2967d0dd649bd5435b/rosidl_generator_cpp/test/test_msg_initialization.cpp). I wanted to open this PR right away in case anyone has an objection to the implementation.

One thing I'd appreciate feedback on is: Should the implementations of the field initializer structs go into a separate implementation header to avoid bloating the `__struct.hpp` header? We could declare the field initializer structs as nested classes while putting their definitions in a separate header that automatically gets included. I think it's mostly a question of aesthetics, although splitting out the header will add some non-trivial complexity to the header generation.


mxgrey:
Tests are added, so I'd say this is ready for review.

My one remaining concern is how many lines of code the initialization structs take up in the class definition. That could be flattened out significantly if we want to define a macro in a commonly shared header. Otherwise if we're not concerned about how many lines of code are being consumed, then this PR is good to go.

dirk-thomas:
Before merging this I would be curious if we can move this into a separate header - simply to keep the `*__struct.hpp` header readable if users take a peek. Would you be interested in trying that direction? If not, I can try to take a look at it.

mxgrey:
I agree with wanting to make the header more readable. I can see two directions we can take:

1. Leave the `Init__` struct declarations nested in the class and then put the definitions of those classes in another header. This would add complexity to the file generation procedure, and I'd have to get very creative to figure out how to make this work reliably.

2. We can add some macros to `rclcpp_generator_cpp/message_initialization.hpp` which take care of implementing all these details, so the actual generated header file would just look something like:

```
  ROSIDL_GENERATOR_CPP_INIT_FIRST_FIELD(MessageName, first_field_name);
  ROSIDL_GENERATOR_CPP_INIT_MID_FIELD(MessageName, second_field_name);
  ROSIDL_GENERATOR_CPP_INIT_MID_FIELD(MessageName, third_field_name);
  /* ... */
  ROSIDL_GENERATOR_CPP_INIT_LAST_FIELD(MessageName, last_field_name);
```

And there'd be a fourth macro called `ROSIDL_GENERATOR_CPP_INIT_ONLY_FIELD(MessageName, only_field_name)` to handle the case where a message only has one field.

Both options would occupy the same number of lines of code inside the class definition. Option (1) would add significant complexity to the file generation but it would scream at you less inside the class definition. Option (2) screams at you a bit in the class definition, but it would require absolutely no changes to the file generation scheme.

I'm more inclined towards option (2). It'll be very easy to draft, so I'll make those changes really quick, and if anyone objects to it we can revert afterwards.

mxgrey:
The generated message headers should be much smaller now. Here's the new snippet from `multi_nested__struct.hpp`:

```
  // Initializers
  ROSIDL_GENERATOR_CPP__INIT_LAST_FIELD(MultiNested_, unbounded_sequence_of_unbounded_sequences);
  ROSIDL_GENERATOR_CPP__INIT_MID_FIELD(MultiNested_, unbounded_sequence_of_bounded_sequences, unbounded_sequence_of_unbounded_sequences);
  ROSIDL_GENERATOR_CPP__INIT_MID_FIELD(MultiNested_, unbounded_sequence_of_arrays, unbounded_sequence_of_bounded_sequences);
  ROSIDL_GENERATOR_CPP__INIT_MID_FIELD(MultiNested_, bounded_sequence_of_unbounded_sequences, unbounded_sequence_of_arrays);
  ROSIDL_GENERATOR_CPP__INIT_MID_FIELD(MultiNested_, bounded_sequence_of_bounded_sequences, bounded_sequence_of_unbounded_sequences);
  ROSIDL_GENERATOR_CPP__INIT_MID_FIELD(MultiNested_, bounded_sequence_of_arrays, bounded_sequence_of_bounded_sequences);
  ROSIDL_GENERATOR_CPP__INIT_MID_FIELD(MultiNested_, array_of_unbounded_sequences, bounded_sequence_of_arrays);
  ROSIDL_GENERATOR_CPP__INIT_MID_FIELD(MultiNested_, array_of_bounded_sequences, array_of_unbounded_sequences);
  ROSIDL_GENERATOR_CPP__INIT_FIRST_FIELD(MultiNested_, array_of_arrays, array_of_bounded_sequences);

  // Build an instance of this message in a way that requires you to explicitly
  // specify a value for every field. The compiler will help you guarantee that
  // every field is initialized.
  static Init__array_of_arrays build()
  {
    return Init__array_of_arrays();
  }
```

It's still a bit noisy, but I think it's roughly the same noise level as the rest of the generated header.

rosidl_generator_cpp/test/test_msg_initialization.cpp
```diff
@@ -216,3 +218,167 @@ TEST(Test_msg_initialization, skip_constructor) {
         return "" == i;
       }));
 }
+
+TEST(Test_msg_initialization, build) {
+  rosidl_generator_cpp::msg::BasicTypes basic =
+      rosidl_generator_cpp::msg::BasicTypes::build()
```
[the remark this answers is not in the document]
mxgrey: You're right, uncrustify wants a single indent here.

My latest commit addresses (most) of the complaints from uncrustify. For now I'm ignoring uncrustify's complaints about the following:

```
  rosidl_generator_cpp::msg::MultiNested multi_nested =
    rosidl_generator_cpp::msg::MultiNested::build()
    .array_of_arrays({arrays, arrays, arrays})
    .array_of_bounded_sequences({
      bounded_sequences,
      bounded_sequences,
      bounded_sequences
    })
    .array_of_unbounded_sequences({
      unbounded_sequences,
      unbounded_sequences,
      unbounded_sequences
    })
    .bounded_sequence_of_arrays({arrays, arrays})
    .bounded_sequence_of_bounded_sequences({
      bounded_sequences,
      bounded_sequences
    })
    .bounded_sequence_of_unbounded_sequences({
      unbounded_sequences,
      unbounded_sequences
    })
    .unbounded_sequence_of_arrays({arrays})
    .unbounded_sequence_of_bounded_sequences({bounded_sequences})
    .unbounded_sequence_of_unbounded_sequences({unbounded_sequences});
```

which uncrustify would prefer to have as: 

```
  rosidl_generator_cpp::msg::MultiNested multi_nested =
    rosidl_generator_cpp::msg::MultiNested::build()
    .array_of_arrays({arrays, arrays, arrays})
    .array_of_bounded_sequences({
    bounded_sequences,
    bounded_sequences,
    bounded_sequences
    })
    .array_of_unbounded_sequences({
    unbounded_sequences,
    unbounded_sequences,
    unbounded_sequences
    })
    .bounded_sequence_of_arrays({arrays, arrays})
    .bounded_sequence_of_bounded_sequences({
    bounded_sequences,
    bounded_sequences
    })
    .bounded_sequence_of_unbounded_sequences({
    unbounded_sequences,
    unbounded_sequences
    })
    .unbounded_sequence_of_arrays({arrays})
    .unbounded_sequence_of_bounded_sequences({bounded_sequences})
    .unbounded_sequence_of_unbounded_sequences({unbounded_sequences});
```

but I think it's confusing to have everything aligned on the same column even though certain items are nested deeper in the construction hierarchy than others.

For now I'm ignoring uncrustify on these and leaving it as-is, but I can change it if there are strict rules about adhering to uncrustify on this.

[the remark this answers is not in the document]
wjwwood: I like yours better than uncrustify's. I'm fine with it, I just thought it was odd that the four space indent was the result.

mxgrey:
Is there anything left to do for this PR? Does anyone have an opinion about the new macro approach?

dirk-thomas:
I still want to explore an alternative approach which is implemented in a separate header without having to inject all the macros into the struct header.

mxgrey:
I think that would be undesirable because we'd be polluting the message namespace with lots of class names, since every field of every message requires its own class. Users of auto-complete (like myself) would see their auto-complete hammered with something like

```
Init__MessageName1__field_name_1
Init__MessageName1__field_name_2
/* ... */
Init__MessageName1__field_name_N
Init__MessageName2__field_name_1
/* ... */
Init__MessageNameM__field_name_N
```

We could hide that by putting those all into some kind of `detail` namespace, but I think making them nested classes is a more elegant approach. Also, declaring them outside of the class definition means we can't have a static `build()` member function in the class definition unless we forward declare these initializer classes and then define them later, which means we're potentially adding two new headers instead of one.

Maybe there's a creative solution that I'm not thinking of, but I'm very skeptical that we'll get something cleaner than this macro approach.

mxgrey:
I also think it's worth pointing out that these macros take up fewer lines of code in the header than even the field type aliases do.

dirk-thomas:
See 656165be649b8680672dc02311919c7eab2f1352 which moves the builder functionality into a completely separate header. At the moment you have to `#include` the header with the `_builder.hpp` suffix explicitly.

The test is using the function `pkgname::msg::build_MsgName()`. I was working on an alternative templated function to enable a generic invocation `rosidl_generator_cpp::build<pkgname::msg::MgsName>()` but ran out of time for today...

mxgrey:
I can understand wanting this separation, but I really think the previous API had better ergonomics:

* The need to include a whole new header seems like an unnecessary disadvantage, effectively hiding a useful feature. If we're going to keep this header separation, I would prefer to at least have it included in the API header of the message.

* It seems more intuitive to me to access the builder function as a static function from the message that's being built, not as a completely separate class. I feel that being its own class implies that the class is useful in some way on its own, whereas being nested in the message class makes it more clear that it's only a compilation trick to get you a message instance.

* Out of all the auto-generated gibberish in the class definition of a message, a line of macros seems pretty innocuous to me.

I guess these details are ultimately about personal preference and aesthetics, which are notoriously difficult things to find a consensus on. So if others think this separation is aesthetically/ergonomically preferable, then I suppose it's fine with me.

But I would definitely advise against making a `rclcpp_generator::build<T>` template. That was my original intention for how to implement this, but then I realized we'd be forcing users to add `rclcpp_generator` all over the place with no benefit at all. Implementing it as a template just adds noise without accomplishing anything for anyone.

dirk-thomas:
> The need to include a whole new header seems like an unnecessary disadvantage, effectively hiding a useful feature. If we're going to keep this header separation, I would prefer to at least have it included in the API header of the message.

As mentioned above: "At the moment you have to `#include` the header with the `_builder.hpp` suffix explicitly. It could of course be included by the main header."

> It seems more intuitive to me to access the builder feature as a static function from the message type that's being built, not as a completely separate class. I feel that being its own class implies that the class is useful in some way on its own, whereas being nested in the message class makes it more clear that it's only a compilation trick to get you a message instance.

There is no "completely separate class". I propose either a global function named after the data class or a templated function.

The same approach is already being applied for e.g. the function. They are being defined in a separate header to keep each file focused on one thing. It also allows to use the struct without any of the additional parts (builder, functions, whatever else) to keep it as light weight as possible. So if someone doesn't want to use the builder there is no need to test / certify that code.

> Out of all the auto-generated gibberish in the class definition of a message, a sequence of N lines of macros (where N is the number of message fields) seems pretty innocuous to me.

The fact that a file is already long sounds like a bad rational for justifying even more to it.

jacobperron:
> At the moment you have to #include the header with the _builder.hpp suffix explicitly. It could of course be included by the main header.

+1 for including `*__builder.hpp` in the main header.

> I propose either a global function named after the data class or a templated function.

I don't care if it's a global function or a method on the class, but I do think the shed would look nicer having the message namespace leading up to the builder, e.g.

```c++
geometry_msgs::msg::Point::build_Point()
// or
geometry_msgs::msg::Point::build::Point()
```
instead of:
```c++
rosidl_generator_cpp::build<geometry_msgs::Point>()
```

dirk-thomas:
> `geometry_msgs::msg::Point::build::Point()`

`geometry_msgs::msg::Point` is already the type of the struct.

jacobperron:
@dirk-thomas I edited my comment, I didn't meant to have the extra struct name in the namespace.

mxgrey:
I think if I were to choose between `geometry_msgs::msg::build_Point()` and `geometry_msgs::msg::build::Point()` I would prefer the latter (`build::Point()`). That helps keep the `geometry_msgs::msg` namespace less "polluted" (according to my own subjective ideas of symbol pollution).

mxgrey:
> Afaik this is only possible when being defined in the struct (which as mentioned above for separation of concern / avoiding optional API to become mandatory I would rather like to not do).

To be super pedantic, it only requires that the function is declared in the struct (and that the initial return type is forward declared somewhere). It could be declared inside the struct and then defined in a separate header. But then that leads to confusion if people try to use it without having all the necessary headers included, so I'd say that's probably the worst combination of design choices.

rosidl_generator_cpp/resource/msg__builder.hpp.em
```diff
@@ -0,0 +1,85 @@
...
+namespace builder
+{
+
+@[  for index in range(len(message.structure.members) - 1, -1, -1)]@
+@{
+field_name = message.structure.members[index].name
+if index < len(message.structure.members) - 1:
+  next_field_name = message.structure.members[index + 1].name
+}@
+class Init_@(message.structure.namespaced_type.name)_@(field_name)
```
hidmic, line 26: I'd think that builders should be templated by allocator type just like interface structs are (see [here](https://github.com/ros2/rosidl/blob/4d837670d5758a3d45b8c86f3d867f44edd1c221/rosidl_generator_cpp/resource/msg__struct.hpp.em#L88-L89)).

mxgrey: At the risk of beating a dead horse, we would get this trivially if we used the earlier static member function approach.

dirk-thomas:
> only requires that the function is declared in the struct (and that the initial return type is forward declared somewhere).

That is not possible since the return type is not a pointer / reference but a value so it can't be forward declared.

dirk-thomas:
We discussed the possible option in the ROS meeting:

* `geometry_msgs::msg::build_Point()` disliked because of polluting the namespace and the mixed naming.
* `geometry_msgs::msg::build::Point()` disliked because of he ambiguity for search&replace or when using `using namespace`.
* `geometry_msgs::msg::Point::build()` requires definition of returned structs in the same header (or at least in an included header), disliked because of lower modularity.
* `geometry_msgs::build<geometry_msgs::msg::Point>() preferred option, the namespace should be the package name rather than something generic like `rosidl_generator_cpp` used in the previous proposal. See 2a3da859a01337feccbfbb5ec220bafa84dcbd01 which actually implements this option.

rosidl_generator_cpp/resource/msg__builder.hpp.em
dirk-thomas: Since none of the ROS 2 API works at the moment with custom allocators I would rather not make this more complicated and just leave custom allocators out of this.

dirk-thomas:
CI builds up to `demo_nodes_cpp`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9971/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5573/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8112/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9839/)

rosidl_generator_cpp/resource/msg__builder.hpp.em
hidmic: Hmm I do see `rclcpp::Publisher` propagating the given `AllocatorT` accordingly. I don't see any `MessageT` allocator being used though. So I'm fine with not doing anything else on this PR, but a follow-up issue to track the missing (or half-implemented or to be deprecated) feature would be nice.

rosidl_generator_cpp/test/test_msg_builder.cpp
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  ASSERT_EQ(1u, multi_nested.unbounded_sequence_of_unbounded_sequences.size());
+  for (const auto & u : multi_nested.unbounded_sequence_of_unbounded_sequences) {
+    ASSERT_EQ(unbounded_sequences, u);
+  }
+
+  rosidl_generator_cpp::msg::Empty empty =
+    rosidl_generator_cpp::build<rosidl_generator_cpp::msg::Empty>()
+    .structure_needs_at_least_one_member(5);
```
hidmic, line 203: @dirk-thomas is this still necessary? I thought we had dropped that restriction.

rosidl_generator_cpp/resource/msg__builder.hpp.em
```diff
@@ -0,0 +1,80 @@
...
+message_typename = '::'.join(message.structure.namespaced_type.namespaced_name())
+}@
+
+@[for ns in message.structure.namespaced_type.namespaces]@
+namespace @(ns)
+{
+
+@[end for]@
+@[if not message.structure.members]@
+::@(message_typename) build_@(message.structure.namespaced_type.name)()
```
hidmic, line 12: @dirk-thomas weren't we dropping `*::build_*()` functions in favour of `*::build<>()`?

rosidl_generator_cpp/test/test_msg_builder.cpp
dirk-thomas: > I thought we had dropped that restriction.

Where do you think that happened? This PR doesn't does the special member in any way.

dirk-thomas:
> ... and pending green CI

Above CI is green (considering the yellow builds with unrelated test failures).

rosidl_generator_cpp/resource/msg__builder.hpp.em
dirk-thomas: Actually this block could be removed. `message.structure.members` is always a non-empty list.

dirk-thomas:
I added special handling for `EMPTY_STRUCTURE_REQUIRED_MEMBER_NAME` in 1e0e96ebef47ef9ad154487702899f6e82bb4297 so that the builder approach doesn't have to know about that member.

rosidl_generator_cpp/resource/msg__builder.hpp.em
dirk-thomas: Done in 1e0e96ebef47ef9ad154487702899f6e82bb4297.

dirk-thomas:
Minimal CI to check latest commit: [Build Status](https://ci.ros2.org/job/ci_linux/9998/)

PR #456, opened 2020-04-10
rename rosidl_generator_cpp namespace to rosidl_runtime_cpp

dirk-thomas:
Follow up of #442 as part of #447 to match namespace with the package the name the symbols are defined in.

PR #457, opened 2020-04-11
use rosidl_runtime_c headers

Karsten1987:
Fixes https://github.com/ros2/rosidl/pull/456#issuecomment-612259111

The headers were moved, but the code not updated to reflect these changes.


dirk-thomas:
See https://github.com/ros2/rosidl/pull/456#issuecomment-612281511 for the proper fix.

Issue #446, opened 2020-03-09
Review of the rosidl_generator_c package


dirk-thomas:
# Comments / feedback

Please provide feedback in separate comments - one per topic - to allow hiding them when they have been resolved - to keep the discussion readable and pending items visible.
Beside any additional feedback it is also valuable to comment if you agree/disagree with the check boxes indicating proposed to-dos.

# Scope of the package

At the moment the package has two purposes:
1. provide the message generator for C which generates C code / library at the build time of each message package
2. provide the C headers / library which messages will depend on at runtime

This poses an unnecessary runtime dependency for message packages.
Therefore this package should be split into two parts.

* [x] This will be addressed in #443 which will separate the first part (the generator) from the second part (the runtime component).

# Package dependencies

* build dependency on `ament_cmake_ros`: it uses the shared/static library option provided by that package

* export the build dependencies on `ament_cmake`: it registers itself as an extension

  * [x] Reducing the dependency to `ament_cmake_core` might be feasible?
    See #459

* export the build dependencies on `rosidl_cmake`: it uses the CMake API provided by that package when generating code for message packages

* build and run dependency on `rosidl_typesupport_interface`: the code compiled in this package as well as the headers exposed by message packages use headers from that package

All of these dependencies look reasonable.

* [x] They will be split as part of #443 between the generator and runtime package

:information_source: The test dependencies are not being reviewed since they are not part of the public interface of the package.

The group membership currently lists both groups `rosidl_generator_packages` as well as `rosidl_runtime_packages`.

* [x] Those will also be split as part of #443 between the generator and runtime package

# Directories and Files

## bin

The script contains the CLI for the C generator.
All logic is in the Python module.

## cmake

* `register_c.cmake` contains a macro to register the `rosidl_generate_idl_interfaces` extension.
* `rosidl_generator_c_extras.cmake` sets some CMake variables an invokes the above macro to register the extension.
* `rosidl_generator_c_generate_interfaces.cmake` contains the CMake logic of the C generator extension which generates the C code for interfaces, builds a library and installs the artifacts.

## include

All C headers end in `.h`.

All headers are in the subdirectory `rosidl_generator_c`.
While #443 moves the headers into the new package `rosidl_runtime_c` it doesn't rename the directory.
As a consequence functions, macros and types are still prefixed with the package name `rosidl_generator_c`.

* [x] Consider renaming the directory / prefix to match the package name after the headers have been moved.
      The trade off will be the amount of external code which needs to be updated for this.
[...]

PR #458, opened 2020-04-11
rename rosidl_generator_c 'namespace' to rosidl_runtime_c

dirk-thomas:
Follow up of #442 as part of #446 to match namespace with the package the name the symbols are defined in. Similar to #456.

The two commits in this PR should not be squashed.

Only Linux with linter tests since the change shouldn't have any runtime impact (if it compiles successfully):[Build Status](https://ci.ros2.org/job/ci_linux/10112/)


rosidl_generator_c/test/test_interfaces.c
```diff
@@ -527,7 +527,7 @@ int test_unbounded_sequences()
   }
 
   // uint8_values
-  res = rosidl_generator_c__uint8__Sequence__init(&seq->uint8_values, ARR_SIZE);
+  res = rosidl_runtime_c__int8__Sequence__init(&seq->uint8_values, ARR_SIZE);
```
dirk-thomas, line 530: No idea how I broke this doing search & replace, but this changes from `uint8` to `int8` results in a compiler warning. See #460.

PR #461, opened 2020-04-14
move non-entry point headers into detail subdir

dirk-thomas:
Follow up of #446 and #447 moving the non-entry point headers into a subdirectory. Since we don't distinguish the directories anywhere based on the language (C or C++) but do so based on the file extension I kept all of the moved headers in a `detail` subdirectory.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10194/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5747/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8293/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10082/)


jacobperron:
Please add an entry to the Foxy release notes regarding this change.

PR #466, opened 2020-04-20
document string structs and sequence functions

dirk-thomas:
Follow up of #446.

Beside adding docblocks this patch only splits the content of the `ROSIDL_RUNTIME_C__DECLARE_PRIMITIVE_SEQUENCE_FUNCTIONS` macros into `ROSIDL_RUNTIME_C__DECLARE_PRIMITIVE_SEQUENCE_INIT` and `ROSIDL_RUNTIME_C__DECLARE_PRIMITIVE_SEQUENCE_FINI` in order to document them. The second argument was removed from the existing macro since it wasn't used.

### DIFF
--- a/rosidl_runtime_c/include/rosidl_runtime_c/primitives_sequence_functions.h
+++ b/rosidl_runtime_c/include/rosidl_runtime_c/primitives_sequence_functions.h
@@ -26,60 +26,112 @@ extern "C"
 {
 #endif
 
-#define ROSIDL_RUNTIME_C__DECLARE_PRIMITIVE_SEQUENCE_FUNCTIONS(STRUCT_NAME, TYPE_NAME) \
+/**
+ * \def ROSIDL_RUNTIME_C__DECLARE_PRIMITIVE_SEQUENCE_INIT(STRUCT_NAME)
+ *
+ * \brief Allocate the memory for the sequence.
+ *
+ * Calling the function with an already allocated sequence will leak the
+ * previously allocated memory.
+ *
+ * param sequence a pointer to a sequence struct
+ * param size the number of items to allocate in the sequence, both sequence
+ *   fields `size` and `capacity` are set to this parameter
+ * return true if successful, false if the passed sequence pointer is null
+ *   or the memory allocation failed
+ */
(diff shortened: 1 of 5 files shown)


rosidl_runtime_c/include/rosidl_runtime_c/primitives_sequence_functions.h
```diff
@@ -26,60 +26,112 @@ extern "C"
...
+ *
+ * \brief Allocate the memory for the sequence.
+ *
+ * Calling the function with an already allocated sequence will leak the
+ * previously allocated memory.
+ *
+ * param sequence a pointer to a sequence struct
+ * param size the number of items to allocate in the sequence, both sequence
+ *   fields `size` and `capacity` are set to this parameter
+ * return true if successful, false if the passed sequence pointer is null
```
jacobperron, line 40: Should this be using Doxygen commands, `\param` and `\return`?

rosidl_runtime_c/include/rosidl_runtime_c/primitives_sequence_functions.h
```diff
@@ -26,60 +26,112 @@ extern "C"
...
+    rosidl_runtime_c__ ## STRUCT_NAME ## __Sequence * sequence, size_t size);
+
+/**
+ * \def ROSIDL_RUNTIME_C__DECLARE_PRIMITIVE_SEQUENCE_FINI(STRUCT_NAME)
+ *
+ * \brief Deallocate the memory of the sequence.
+ *
+ * Calling the function with an already deallocated sequence is a no-op.
+ *
+ * param sequence a pointer to a sequence struct
```
jacobperron, line 56: Same question here. `\param`?

rosidl_runtime_c/include/rosidl_runtime_c/primitives_sequence_functions.h
dirk-thomas: It can't be since the docblock is for the macro and therefore can't declare the parameter / return value of the function generated by the macro.

rosidl_runtime_c/include/rosidl_runtime_c/primitives_sequence_functions.h
dirk-thomas: Same answer.

PR #478, opened 2020-04-30
use typesupport targets instead of libraries

dirk-thomas:
Addresses https://github.com/ros2/ros2/issues/915#issuecomment-621616901

Packaging build: https://ci.ros2.org/job/ci_packaging_linux/333/


ros-discourse:

https://discourse.ros.org/t/best-practices-for-replacing-ament-target-dependencies-in-kilted-while-maintaining-compatibility/43938/6

PR #483, opened 2020-05-08
Fix variable suffix in rosidl_export_typesupport_targets

ivanpauno:
`_suffix` wasn't being set here:

https://github.com/ros2/rosidl/blob/f3f639df4cef29fda1dc18074fe0b1e2c3b108e6/rosidl_cmake/cmake/rosidl_cmake_export_typesupport_targets-extras.cmake.in#L15

This PR fixes that.

### DIFF
--- a/rosidl_cmake/cmake/rosidl_export_typesupport_targets.cmake
+++ b/rosidl_cmake/cmake/rosidl_export_typesupport_targets.cmake
@@ -15,13 +15,15 @@
 #
 # Export typesupport targets to downstream packages.
 #
+# :param library_suffix: the suffix of the library
+# :type library_suffix: string
 # :param ARGN: a list of targets.
 #   Each element must be an exported CMake library target.
 # :type ARGN: list of strings
 #
 # @public
 #
-macro(rosidl_export_typesupport_targets)
+macro(rosidl_export_typesupport_targets library_suffix)
   if(_${PROJECT_NAME}_AMENT_PACKAGE)
     message(FATAL_ERROR
       "rosidl_export_typesupport_targets() must be called before "
@@ -44,7 +46,10 @@ macro(rosidl_export_typesupport_targets)
           "not-imported targets")
       endif()
 
-      list(APPEND _ROSIDL_CMAKE_EXPORT_TYPESUPPORT_TARGETS "${PROJECT_NAME}::${_target}")
+      list(APPEND
+        _ROSIDL_CMAKE_EXPORT_TYPESUPPORT_TARGETS
+        "${library_suffix}:${_target}"
+      )
     endforeach()
   endif()
 endmacro()
--- a/rosidl_cmake/cmake/rosidl_cmake_export_typesupport_targets-extras.cmake.in
+++ b/rosidl_cmake/cmake/rosidl_cmake_export_typesupport_targets-extras.cmake.in
@@ -7,7 +7,12 @@ set(_exported_typesupport_targets
 # populate @PROJECT_NAME@_TARGETS_<suffix>
 if(NOT _exported_typesupport_targets STREQUAL "")
   # loop over typesupport targets
-  foreach(_target ${_exported_typesupport_targets})
+  foreach(_tuple ${_exported_typesupport_targets})
+    string(REPLACE ":" ";" _tuple "${_tuple}")
+    list(GET _tuple 0 _suffix)
+    list(GET _tuple 1 _target)
+
+    set(_target "@PROJECT_NAME@::${_target}")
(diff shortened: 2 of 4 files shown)


rosidl_cmake/cmake/rosidl_export_typesupport_targets.cmake
```diff
@@ -15,13 +15,15 @@
...
+# :param library_suffix: the suffix of the library
+# :type library_suffix: string
 # :param ARGN: a list of targets.
 #   Each element must be an exported CMake library target.
 # :type ARGN: list of strings
 #
 # @public
 #
-macro(rosidl_export_typesupport_targets)
+macro(rosidl_export_typesupport_targets library_suffix)
```
dirk-thomas, line 26: This is an API breaking change. Why should we land that after the API freeze?

rosidl_cmake/cmake/rosidl_export_typesupport_targets.cmake
```diff
@@ -44,7 +46,10 @@ macro(rosidl_export_typesupport_targets)
           "not-imported targets")
       endif()
 
-      list(APPEND _ROSIDL_CMAKE_EXPORT_TYPESUPPORT_TARGETS "${PROJECT_NAME}::${_target}")
+      list(APPEND
+        _ROSIDL_CMAKE_EXPORT_TYPESUPPORT_TARGETS
+        "${library_suffix}:${_target}"
```
dirk-thomas, line 51: This looks wrong to me. Modern CMake targets have (by convention) the form `<pkgname>::<libname>`. The library suffix seems to be the `_target_suffix` from the caller which has two underscores and the typesupport package name? Where is this target being defined?

rosidl_cmake/cmake/rosidl_export_typesupport_targets.cmake
ivanpauno: The function is supposed to append the passed targets to a `@PROJECT_NAME@_TARGETS${_suffix}` variable.
Instead of that, it's appending it to `@PROJECT_NAME@_TARGETS`, as `${_suffix}` is always empty (if it's not empty, it's a value randomly set previously by another thing in the same scope).

I didn't see a way of fixing behavior without modifying the API, and thus I modified it.

rosidl_cmake/cmake/rosidl_export_typesupport_targets.cmake
ivanpauno: > This looks wrong to me. Modern CMake targets have (by convention) the form `<pkgname>::<libname>`

That hasn't changed.
Please read relevant changes in other files:

https://github.com/ros2/rosidl/blob/d42b3925208871cabd0385282112f73cf14f52a0/rosidl_cmake/cmake/rosidl_cmake_export_typesupport_targets-extras.cmake.in#L11-L15

rosidl_cmake/cmake/rosidl_export_typesupport_targets.cmake
hidmic: It's unfortunate indeed, but I don't see a way around it either.

rosidl_cmake/cmake/rosidl_export_typesupport_targets.cmake
dirk-thomas: From the change it simply wasn't clear to me that you don't pass targets anymore but a tuple.

ivanpauno: > From the change it simply wasn't clear to me that you don't pass targets anymore but a tuple.

yes, I understand. Should I change the variable `_ROSIDL_CMAKE_EXPORT_TYPESUPPORT_TARGETS` name to something different?

dirk-thomas: You could rename the variable - I just don't have a good naming proposal. It can stay as is too.

ivanpauno:
CI config (testing only this PR, not including https://github.com/ros2/rosidl_typesupport_fastrtps/pull/40):

* build: `--packages-up-to test_msgs`
* test: `--packages-select rosidl_cmake`

Jobs:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10840/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6213/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8831/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10727/)

PR #487, opened 2020-05-12
Clean up BoundedVector

jwakely:
- Stop using reserved names for parameters
- Make comparison operators friends
- Remove unnecessary conversion operator

I've only done minimal testing of these changes, just checking the BoundedVector header can be included and compiled.

### DIFF
--- a/rosidl_runtime_cpp/include/rosidl_runtime_cpp/bounded_vector.hpp
+++ b/rosidl_runtime_cpp/include/rosidl_runtime_cpp/bounded_vector.hpp
@@ -640,12 +640,6 @@ public:
   using Base::clear;
 
 private:
-  /// Cast to base type, to make it easier to dispatch to base implementations
-  operator Base &()
-  {
-    return *this;
-  }
-
   /// Vector equality comparison.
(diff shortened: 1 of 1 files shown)


dirk-thomas:
> thoughts?

Nothing from me.

jacobperron:
It's okay with me with full CI 👍

jwakely:
I've force-pushed to fix the linter issues.

hidmic:
Running full CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10775/)

jwakely:
Are those CI failures due to my patch or was `master` already showing failures?

Looks like the builds are green now so I'll rebase and force-push this again.

rotu:
Thank you so much! This makes the code tons more readable (plus some of the identifiers you've removed [were technically UB](https://en.cppreference.com/w/cpp/language/identifiers))

jwakely:
They were indeed, that's why I removed them. As a standard library implementor I rely on reserved names being reserved, and not clashing with identifiers in user code.

PR #493, opened 2020-05-20
Fix misuses of input iterators in BoundedVector

### LINKED ISSUE #486: BoundedVector misuses InputIterators
Input iterators are single pass, meaning once you've incremented them, they're gone. That means functions like [this](https://github.com/ros2/rosidl/blob/afff057f283959d33da065349eb971b8b3653170/rosidl_runtime_cpp/include/rosidl_runtime_cpp/bounded_vector.hpp#L283) are broken for input iterators. The call to `distance(first, last)` will consume the entire range, and then `assign(first, last)` will do nothing.

```
#include <sstream>
#include <iterator>
#include <cassert>
#include <rosidl_runtime_cpp/bounded_vector.hpp>
int main()
{
  std::istringstream s("1 2 3");
  BoundedVector<int> b;
  std::istream_iterator<int> iter(s), end;
  b.assign(iter, s);
  assert(b.size() == 3);
}
```
You should either change that function to only accept forward iterators, or use tag dispatching to select different functions depending on `std::iterator_traits<InputIterator>::iterator_category`. For input iterators you could just do something like:

```
Base::clear();
while (first != last)
  push_back(*first++);
```
For forward iterators you can use `std::distance(first, last)` safely.

jwakely:
Input iterators are single pass, so calling std::distance(first, last)
will consume the input and leave [first,last) as an empty range.

This change makes assign(InputIterator, InputIterator) and
insert(const_iterator, InputIterator, InputIterator) dispatch to
overloaded functions that choose the right overload based on the
iterator category. For forward iterators the original implementation is
used. For input iterators assign will construct a new vector and if that
doesn't throw, use swap to replace the contents of *this. For insert
each value is inserted at the end and if the bound is not exceeded
std::rotate is used to shift them to the desired position, otherwise the
inserted elements are removed from the end before the function exits.

This retains the strong exception safety guarantee for both operations,
so that exceeding the bound does not alter the existing contents of the
vector.

Fixes #486

Signed-off-by: Jonathan Wakely <github@kayari.org>

### DIFF
--- a/rosidl_runtime_cpp/include/rosidl_runtime_cpp/bounded_vector.hpp
+++ b/rosidl_runtime_cpp/include/rosidl_runtime_cpp/bounded_vector.hpp
@@ -286,10 +286,8 @@ public:
   void
   assign(InputIterator first, InputIterator last)
   {
-    if (std::distance(first, last) > UpperBound) {
-      throw std::length_error("Exceeded upper bound");
-    }
-    Base::assign(first, last);
+    using cat = typename std::iterator_traits<InputIterator>::iterator_category;
+    do_assign(first, last, cat());
   }
 
   /// Assign an initializer list to a %BoundedVector.
@@ -629,11 +627,8 @@ public:
     InputIterator first,
     InputIterator last)
   {
-    auto dist = std::distance(first, last);
-    if ((dist < 0) || (size() + static_cast<size_t>(dist) > UpperBound)) {
-      throw std::length_error("Exceeded upper bound");
-    }
-    return Base::insert(position, first, last);
+    using cat = typename std::iterator_traits<InputIterator>::iterator_category;
+    return do_insert(position, first, last, cat());
   }
 
   using Base::erase;
@@ -641,6 +636,73 @@ public:
   using Base::clear;
 
 private:
+  /// Assign elements from an input range.
+  template<
+    typename InputIterator
+  >
+  void
+  do_assign(InputIterator first, InputIterator last, std::input_iterator_tag)
+  {
+    BoundedVector(first, last).swap(*this);
+  }
+
+  /// Assign elements from a forward range.
+  template<
+    typename FwdIterator
+  >
+  void
+  do_assign(FwdIterator first, FwdIterator last, std::forward_iterator_tag)
+  {
+    if (static_cast<std::size_t>(std::distance(first, last)) > UpperBound) {
+      throw std::length_error("Exceeded upper bound");
+    }
+    Base::assign(first, last);
+  }
+
+  // Insert each value at the end and then rotate them to the desired position.
+  // If the bound is exceeded, the inserted elements are removed again.
+  template<
+    typename InputIterator
+  >
+  typename Base::iterator
+  do_insert(
+    typename Base::const_iterator position,
+    InputIterator first,
+    InputIterator last,
+    std::input_iterator_tag)
+  {
+    const auto orig_size = size();
+    const auto idx = position - cbegin();
+    try {
+      while (first != last) {
+        push_back(*first++);
+      }
+    } catch (const std::length_error &) {
+      Base::resize(orig_size);
+      throw;
+    }
+    auto pos = begin() + idx;
+    std::rotate(pos, begin() + orig_size, end());
+    return begin() + idx;
+  }
+
+  template<
+    typename FwdIterator
+  >
+  typename Base::iterator
+  do_insert(
+    typename Base::const_iterator position,
+    FwdIterator first,
+    FwdIterator last,
+    std::forward_iterator_tag)
+  {
+    auto dist = std::distance(first, last);
+    if ((dist < 0) || (size() + static_cast<size_t>(dist) > UpperBound)) {
+      throw std::length_error("Exceeded upper bound");
+    }
+    return Base::insert(position, first, last);
+  }
+
   /// Vector equality comparison.
   /**
    * This is an equivalence relation.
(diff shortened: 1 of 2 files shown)


rosidl_runtime_cpp/include/rosidl_runtime_cpp/bounded_vector.hpp
```diff
@@ -640,6 +640,71 @@ class BoundedVector
...
+    typename Base::const_iterator position,
+    InputIterator first,
+    InputIterator last,
+    std::input_iterator_tag)
+  {
+    const auto orig_size = size();
+    const auto idx = position - cbegin();
+    try {
+      while (first != last)
+        push_back(*first++);
```
hidmic, line 682: @jwakely nit: use braces for this `while` statement's block

jwakely: Fixed and force-pushed.

jwakely:
I've added some unit test and re-pushed.

rosidl_runtime_cpp/test/test_bounded_vector.cpp
```diff
@@ -50,3 +52,39 @@ TEST(rosidl_generator_cpp, bounded_vector_rvalue) {
   ASSERT_EQ(vv[0], 1);
   ASSERT_EQ(vv[1], 2);
 }
+
+TEST(rosidl_generator_cpp, bounded_vector_comparisons) {
+  rosidl_runtime_cpp::BoundedVector<int, 2> v;
+  ASSERT_EQ(v, v);
+  const decltype(v) vv;
```
hidmic, line 59: @jwakely why not 

```c++
  rosidl_runtime_cpp::BoundedVector<int, 2> v, vv;
```

above? It's a tiny bit clearer.

rosidl_runtime_cpp/test/test_bounded_vector.cpp
```diff
@@ -50,3 +52,39 @@ TEST(rosidl_generator_cpp, bounded_vector_rvalue) {
...
+  ASSERT_GT(v, vv);
+  ASSERT_LT(vv, v);
+}
+
+TEST(rosidl_generator_cpp, bounded_vector_input_iterators) {
+  rosidl_runtime_cpp::BoundedVector<int, 4> v;
+  std::istringstream ss("1 2 3");
+  std::istream_iterator<int> ii(ss), end;
+  v.assign(ii, end);
+  ASSERT_EQ(v.size(), 3);
```
hidmic, line 71: @jwakely checking `v`'s content would also be good. Same elsewhere.

rosidl_runtime_cpp/test/test_bounded_vector.cpp
```diff
@@ -50,3 +52,39 @@ TEST(rosidl_generator_cpp, bounded_vector_rvalue) {
...
+  ss.clear();
+  ss.str("10 11 12 13 14 15");
+  ii == ss;
```
hidmic, line 74: @jwakely `ii` equals `ss`?  I lost you here.

rosidl_runtime_cpp/test/test_bounded_vector.cpp
jwakely: I was just copying what was already on line 49.

jwakely: If you don't like that style, you should probably change the existing tests so people don't try to be consistent with them.

rosidl_runtime_cpp/test/test_bounded_vector.cpp
jwakely: Done

rosidl_runtime_cpp/test/test_bounded_vector.cpp
jwakely: Gah, it should be `ii = ss`. It happens to work because the iterator still refers to that stream anyway, but it should be `ii = ss`. Fixed now.

rosidl_runtime_cpp/test/test_bounded_vector.cpp
hidmic: Hmm, I see. Not happy with either, but fine to leave as-is.

jwakely:
Yes. You were always missing such a test :stuck_out_tongue:

jwakely:
The new unit test (which is testing existing functionality) is triggering a new warning, which was always present but didn't warn because you weren't testing those member functions. At some point I'm going to point out it's your code that was broken, and missing unit tests, and there's a limit to how much time I'm prepared to spend being a good samaritan on a project I have no personal interest in.

jwakely:
Tests green again, including the tests added for the existing functionality and for the part I fixed.

sloretz:
@hidmic @jwakely friendly ping :) Looks like this one needs a rebase

hidmic:
@jwakely I'll go ahead and do the rebase. Hope you don't mind.

jwakely:
Please feel free. I hadn't noticed it had bitrotted.

jwakely:
The conflict was caused by #444 which is ... odd. Defending against completely bogus code like trying to insert `end(), begin()` seems unnecessary, and the new test has undefined behaviour, see https://github.com/ros2/rosidl/pull/444/files#r442387420

You should have merged my pull request before it got invalidated by that :wink:

hidmic:
No worries, if it's UB it has to be fixed one way or another.

hidmic:
Sanity check up to `rosidl_runtime_cpp`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11205/)

hidmic:
Since you reviewed #444, mind to take a look @dirk-thomas ?

hidmic:
Alright, CI's green. Thank you again for pushing this @jwakely!

PR #502, opened 2020-06-23
Add pytest.ini so tests succeed locally.

clalancette:

This is part of ros2/ros2#951, which has more explanation and CI.


dirk-thomas:
Adding the file to the repository root wouldn't work if users only fetch individual packages (e.g. using `rosinstall_generator`). Therefore I would suggest to add the file to each package which needs it separately.

clalancette:
> Adding the file to the repository root wouldn't work if users only fetch individual packages (e.g. using `rosinstall_generator`). Therefore I would suggest to add the file to each package which needs it separately.

Makes sense, will change it.

PR #503, opened 2020-06-24
rosidl_generator_c: don't depend on rosidl_runtime_c when tests are disabled

lopsided98:
`rosidl_generator_c` currently depends `rosidl_runtime_c` (a `test_depend` in `package.xml`) even when `BUILD_TESTING` is disabled, causing the build to fail if the `test_depend`s are not available. This PR makes the dependency conditional on `BUILD_TESTING`.


dirk-thomas:
Thank you for the fix. It seems this regression was introduced in #442.

PR #511, opened 2020-08-02
Modifications to python generator lib to return generated files

atyshka:
When creating my own message generator, I found in some cases it was helpful to have a list of all generated files. In my case, it was to create a swift module that included all the generated C headers, but this could help other things like umbrella headers. This PR simply changes the generate_files function to return the generated files rather than nothing. I don't believe this should impact existing code


dirk-thomas:
Thanks for the patch.

Simple CI build mostly for running the linters: [Build Status](https://ci.ros2.org/job/ci_osx/9583/)

PR #514, opened 2020-08-13
Add function for getting a types fully qualified name

jacobperron:
As opposed the type name 'foo::msg::Bar', the new function let's us get the ROS namespaced name 'foo/msg/Bar'.

This seems generally useful, and would simplify logic in https://github.com/ros2/rviz/pull/591 for example.

I'm open to changing the name of the new function if there are better suggestions.

### DIFF
--- a/rosidl_generator_cpp/CMakeLists.txt
+++ b/rosidl_generator_cpp/CMakeLists.txt
@@ -92,13 +92,13 @@ if(BUILD_TESTING)
       ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}
     )
   endif()
-  ament_add_gtest(test_fully_qualified_name test/test_fully_qualified_name.cpp)
-  if(TARGET test_fully_qualified_name)
-    add_dependencies(test_fully_qualified_name ${PROJECT_NAME})
-    ament_target_dependencies(test_fully_qualified_name
+  ament_add_gtest(test_name test/test_name.cpp)
+  if(TARGET test_name)
+    add_dependencies(test_name ${PROJECT_NAME})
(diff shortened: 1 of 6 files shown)


jacobperron:
Good suggestion. What do you think about adding three new functions instead?

- `name()` (e.g. returns `Bar`)
- `namespace()` (e.g. returns `foo/msg`)
- `name_with_namespace()` (e.g. returns `foo/msg/Bar`)

wjwwood:
I like `fully_qualified_name`, and `name` seems ok, but we can't do just `namespace` I think due to keyword collisions. Maybe `msg_namespace` or `namespace_` or we could prefix all with `type`, e.g. `type_name`, `type_namespace`, and `type_fully_qualified name`/`fully_qualified_type_name`.

The other option is just to have `name` return the full name and let users split it themselves.

I don't have a strong opinion one way or the other.

jacobperron:
> The other option is just to have name return the full name and let users split it themselves.

I like this option. I think it's unlikely a user only wants the last part of the name (or just the namespace), and in this case they can do the splitting themselves.

jacobperron:
@wjwwood I've renamed the function to `name()`, see 0e79466.

jacobperron:
Testing rosidl_generator_cpp, rosidl_runtime_cpp, and test_msgs:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11973/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7006/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9757/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11911/)

PR #517, opened 2020-08-24
[dashing backport] Add pytest.ini so tests succeed locally. (#502)

jacobperron:
Backport #502 to Dashing.

Connects to ros2/ros2#1016


clalancette (approved):

jacobperron:
In retrospect, I don't think this PR is necessary (see https://github.com/ros2/ros2/issues/1016#issuecomment-690744417).

PR #523, opened 2020-09-04
add to_yaml() function for C++ messages

### LINKED ISSUE #259: [rosidl_generator_cpp] add <<operator for message classes
We should provide a `<<operator` like in ROS 1 to ease printing to streams. Some places in the codebase print manually message content to stream in the meantime, e.g. https://github.com/ros2/demos/blob/144fb6d80810f149fa2457ae8542b332a6497674/demo_nodes_cpp/src/parameters/parameter_events.cpp#L26

dirk-thomas:
Closes #259. Instead of defining a `<<` operator on the message this patch adds a `to_yaml()` function taking a message as well as an `ostream`. That allows to implement other formats like `to_xml()` side by side.

Atm all logic is in header files since `rosidl_generator_cpp` doesn't create a library.

CI builds testing `rosidl_generator_cpp`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12223/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7206/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9948/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12122/)

### DIFF
--- a/rosidl_generator_cpp/CMakeLists.txt
+++ b/rosidl_generator_cpp/CMakeLists.txt
@@ -105,7 +105,7 @@ if(BUILD_TESTING)
   if(MSVC)
     # https://developercommunity.visualstudio.com/content/problem/919371/c2017-illegal-escape-sequence-when-using-in-a-raw.html
     set_source_files_properties(
-      test/test_traits.cpp PROPERTIES COMPILE_FLAGS "/experimental:preprocessor")
+      test/test_traits.cpp PROPERTIES COMPILE_FLAGS "/Zc:preprocessor")
   endif()
   ament_add_gtest(test_traits test/test_traits.cpp)
   if(TARGET test_traits)
--- a/rosidl_generator_cpp/CMakeLists.txt
+++ b/rosidl_generator_cpp/CMakeLists.txt
@@ -102,6 +102,11 @@ if(BUILD_TESTING)
       ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}
     )
   endif()
+  if(MSVC)
+    # https://developercommunity.visualstudio.com/content/problem/919371/c2017-illegal-escape-sequence-when-using-in-a-raw.html
+    set_source_files_properties(
+      test/test_traits.cpp PROPERTIES COMPILE_FLAGS "/experimental:preprocessor")
+  endif()
   ament_add_gtest(test_traits test/test_traits.cpp)
   if(TARGET test_traits)
     add_dependencies(test_traits ${PROJECT_NAME})
--- a/rosidl_generator_cpp/resource/msg__traits.hpp.em
+++ b/rosidl_generator_cpp/resource/msg__traits.hpp.em
@@ -5,7 +5,9 @@ from rosidl_parser.definition import ACTION_GOAL_SUFFIX
 from rosidl_parser.definition import ACTION_RESULT_SUFFIX
 from rosidl_parser.definition import Array
 from rosidl_parser.definition import AbstractGenericString
+from rosidl_parser.definition import BasicType
 from rosidl_parser.definition import BoundedSequence
+from rosidl_parser.definition import EMPTY_STRUCTURE_REQUIRED_MEMBER_NAME
 from rosidl_parser.definition import NamespacedType
 from rosidl_parser.definition import AbstractSequence
 from rosidl_parser.definition import UnboundedSequence
@@ -60,6 +62,79 @@ for member in message.structure.members:
 @
 namespace rosidl_generator_traits
 {
+
+void to_yaml(
+  const @(message_typename) & msg,
+  std::ostream & out, size_t indentation = 0)
+{
+@[if len(message.structure.members) == 1 and message.structure.members[0].name == EMPTY_STRUCTURE_REQUIRED_MEMBER_NAME]@
+  (void)msg;
+  (void)indentation;
+  out << "null\n";
+@[else]@
+@[  for i, member in enumerate(message.structure.members)]@
+@[    if i]@
+
+@[    end if]@
+  // member: @(member.name)
+  {
+    if (indentation > 0) {
+      out << std::string(indentation, ' ');
+    }
+@[    if isinstance(member.type, BasicType)]@
+    out << "@(member.name): ";
+@[      if member.type.typename in ('octet', 'char', 'wchar')]@
+    character_value_to_yaml(msg.@(member.name), out);
+@[      else]@
+    value_to_yaml(msg.@(member.name), out);
+@[      end if]@
+    out << "\n";
+@[    elif isinstance(member.type, AbstractGenericString)]@
+    out << "@(member.name): ";
+    value_to_yaml(msg.@(member.name), out);
+    out << "\n";
+@[    elif isinstance(member.type, NamespacedType)]@
+    out << "@(member.name):\n";
+    to_yaml(msg.@(member.name), out, indentation + 2);
+@[    elif isinstance(member.type, (AbstractSequence, Array))]@
+    if (msg.@(member.name).size() == 0) {
+      out << "@(member.name): []\n";
+    } else {
+      out << "@(member.name):\n";
+      for (auto item : msg.@(member.name)) {
+        if (indentation > 0) {
+          out << std::string(indentation, ' ');
+        }
+@[      if isinstance(member.type.value_type, BasicType)]@
+        out << "- ";
+@[        if member.type.value_type.typename in ('octet', 'char', 'wchar')]@
+        character_value_to_yaml(item, out);
+@[        else]@
+        value_to_yaml(item, out);
+@[        end if]@
+        out << "\n";
+@[      elif isinstance(member.type.value_type, AbstractGenericString)]@
+        out << "- ";
+        value_to_yaml(item, out);
+        out << "\n";
+@[      elif isinstance(member.type.value_type, NamespacedType)]@
+        out << "-\n";
+        to_yaml(item, out, indentation + 2);
+@[      end if]@
+      }
+    }
+@[    end if]@
+  }
+@[  end for]@
+@[end if]@
+}  // NOLINT(readability/fn_size)
+
+std::string to_yaml(const @(message_typename) & msg)
+{
+  std::ostringstream out;
+  to_yaml(msg, out);
+  return out.str();
+}
 
 template<>
 inline const char * data_type<@(message_typename)>()
--- a/rosidl_generator_cpp/resource/msg__traits.hpp.em
+++ b/rosidl_generator_cpp/resource/msg__traits.hpp.em
@@ -63,7 +63,7 @@ for member in message.structure.members:
 namespace rosidl_generator_traits
 {
 
-void to_yaml(
+inline void to_yaml(
   const @(message_typename) & msg,
   std::ostream & out, size_t indentation = 0)
 {
@@ -129,7 +129,7 @@ void to_yaml(
 @[end if]@
 }  // NOLINT(readability/fn_size)
 
-std::string to_yaml(const @(message_typename) & msg)
+inline std::string to_yaml(const @(message_typename) & msg)
 {
   std::ostringstream out;
   to_yaml(msg, out);
--- a/rosidl_generator_cpp/test/test_traits.cpp
+++ b/rosidl_generator_cpp/test/test_traits.cpp
@@ -1,3 +1,4 @@
+﻿// NOLINT: This file starts with a BOM since it contain non-ASCII characters
 // Copyright 2019 Open Source Robotics Foundation, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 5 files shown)


rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
```diff
@@ -15,11 +15,137 @@
...
 
+void value_to_yaml(bool value, std::ostream & out)
+{
+  out << (value ? "true" : "false");
+}
+
+void character_value_to_yaml(unsigned char value, std::ostream & out)
+{
+  auto flags = out.flags();
+  out << "0x" << std::hex << std::setw(2) << std::setfill('0') << +value;
```
ivanpauno, line 33: why the `+` prefixing `value`?

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
```diff
@@ -15,11 +15,137 @@
...
+  out.flags(flags);
+}
+
+void character_value_to_yaml(char16_t value, std::ostream & out)
+{
+  auto flags = out.flags();
+  out << "\"\\u" << std::hex << std::setw(4) << std::setfill('0') << \
+    *reinterpret_cast<const int16_t *>(&value) << "\"";
```
ivanpauno, line 41: why the cast is needed?
isn't safer to `static_cast` between both?
```suggestion
    static_cast<const int16_t>(value) << "\"";
```

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
```diff
@@ -15,11 +15,137 @@
...
+}
+
+void value_to_yaml(const std::u16string & value, std::ostream & out)
+{
+  out << "\"";
+  std::wstring_convert<std::codecvt_utf8_utf16<char16_t>, char16_t> convert;
+  auto flags = out.flags();
+  size_t index = 0;
+  while (index < value.size()) {
+    int16_t character = *reinterpret_cast<const int16_t *>(&value[index]);
```
ivanpauno, line 132: same about using `static_cast` instead

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
```diff
@@ -15,11 +15,137 @@
...
+    if (!(character & 0xff80)) {  // ASCII
+      std::string character_as_string = convert.to_bytes(character);
+      out << std::hex << character_as_string.c_str();
+    } else if (!(character & 0xff00)) {  // only 1 byte set
+      out << "\\x" << std::hex << std::setw(2) << std::setfill('0') << \
+        character;
+    } else {
+      out << "\\u" << std::hex << std::setw(4) << std::setfill('0') << \
+        character;
+    }
```
ivanpauno, line 142: note to self: double check if this is correct or not

rosidl_generator_cpp/CMakeLists.txt
```diff
@@ -102,6 +102,11 @@ if(BUILD_TESTING)
       ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}
     )
   endif()
+  if(MSVC)
+    # https://developercommunity.visualstudio.com/content/problem/919371/c2017-illegal-escape-sequence-when-using-in-a-raw.html
+    set_source_files_properties(
+      test/test_traits.cpp PROPERTIES COMPILE_FLAGS "/experimental:preprocessor")
+  endif()
```
ivanpauno, line 109: Instead of relying on the pre-processor supporting raw strings, can we just use normal multi-line strings with the proper escape characters?
I understand that raw strings are much nicer, but I'm not what will happen when compiling in a system that doesn't support them (e.g.: an old windows VS).
If it's a way to detect if the system supports raw strings, and skip compiling the test in case if not I think it would be ok to use them.

ivanpauno (commented):
I have left some minor comments, otherwise lgtm!

---

If in the future we want to support different serialization formats, we could use policy based design, where a `serialize_to_x` template function takes a policy providing overloads for the corresponding `character_value_to_yaml`, `value_to_yaml` functions.

In that way, different formats can be supported without the need of generating more code.

rosidl_generator_cpp/resource/msg__traits.hpp.em
```diff
@@ -61,6 +63,79 @@ for member in message.structure.members:
...
+{
+@[if len(message.structure.members) == 1 and message.structure.members[0].name == EMPTY_STRUCTURE_REQUIRED_MEMBER_NAME]@
+  (void)msg;
+  (void)indentation;
+  out << "null\n";
+@[else]@
+@[  for i, member in enumerate(message.structure.members)]@
+@[    if i]@
+
+@[    end if]@
```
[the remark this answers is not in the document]
dirk-thomas: Separating the blocks of code for each member by an empty line.

rosidl_generator_cpp/resource/msg__traits.hpp.em
```diff
@@ -61,6 +63,79 @@ for member in message.structure.members:
...
+@[        end if]@
+        out << "\n";
+@[      elif isinstance(member.type.value_type, AbstractGenericString)]@
+        out << "- ";
+        value_to_yaml(item, out);
+        out << "\n";
+@[      elif isinstance(member.type.value_type, NamespacedType)]@
+        out << "-\n";
+        to_yaml(item, out, indentation + 2);
+@[      end if]@
```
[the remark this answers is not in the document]
dirk-thomas: The complexity of these templates is already pretty high. The deduplication is likely not worth the decreased readability. The current patterns is common across the various templates.

rosidl_generator_cpp/test/test_traits.cpp
```diff
@@ -13,14 +14,230 @@
 // limitations under the License.
 
 #include <gtest/gtest.h>
+// #include "rosidl_generator_cpp/idl/idl_only_types.hpp"
```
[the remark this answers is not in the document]
dirk-thomas: Because the `test_interface_files` package doesn't export this file atm: https://github.com/ros2/test_interface_files/blob/c4d710e3394ad09a71a257f5490688653fb441b3/CMakeLists.txt#L32

rosidl_generator_cpp/test/test_traits.cpp
```diff
@@ -13,14 +14,230 @@
...
+        msg).c_str());
+
+    msg.wchar_value = u'貓';
+    EXPECT_STREQ(
+      R"(wchar_value: "\u8c93"
+long_double_value: 1.12500
+)",
+      to_yaml(
+        msg).c_str());
+  }*/
```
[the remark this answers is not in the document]
dirk-thomas: Because the `test_interface_files` package doesn't export this file atm: https://github.com/ros2/test_interface_files/blob/c4d710e3394ad09a71a257f5490688653fb441b3/CMakeLists.txt#L32

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
dirk-thomas: > why the cast is needed?

`value` is a `char16_t` and the code wants to interpret it as a numeric value of type `int16_t`. Therefore I think `reinterpret_cast` is appropriate.

> isn't safer to `static_cast` between both?

What do you mean with "safer"?

rosidl_generator_cpp/resource/msg__traits.hpp.em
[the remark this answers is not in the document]
ivanpauno: sounds reasonable

rosidl_generator_cpp/CMakeLists.txt
dirk-thomas: > Instead of relying on the pre-processor supporting raw strings, can we just use normal multi-line strings with the proper escape characters?

The test code intentionally uses a raw string literal so that the expected `yaml` can be copy-n-pasted into a parser / validator.

> I understand that raw strings are much nicer, but I'm not what will happen when compiling in a system that doesn't support them (e.g.: an old windows VS).
If it's a way to detect if the system supports raw strings, and skip compiling the test in case if not I think it would be ok to use them.

According to REP 2000 all active ROS 2 distros require at Visual Studio 2019. So I don't see that we need to support any older versions and therefore don't need any conditional logic.

rosidl_generator_cpp/test/test_traits.cpp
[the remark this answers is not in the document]
ivanpauno: It would be nice to have a comment here explaining why.
If not, it's hard to understand the reason.

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
dirk-thomas: Because the value needs to be streamed as a numeric value.

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
ivanpauno: > What do you mean with "safer"?

`reinterpret_cast` allows to do more (and unsafer) conversions than `static_cast`.
By unsafer, I mean that it allows transmuting between everything.

If `static_cast` does the job, it should be preferred.

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
ivanpauno: I didn't know that trick, it was hard to find a place explaining it.

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
dirk-thomas: I understand that. My point is that we *are* reinterpreting the data type. If you try it with a `static_cast` I am pretty certain it will result in a compile error because the types can't be statically casted between.

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
dirk-thomas: See e46ba0ca046d13cc1b6a2b05f3d166921f9671c8 which explicitly casts to `uint16_t` to make it easier to read.

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
ivanpauno: All conversions between character types and integer types are allowed (see [is_integral](https://en.cppreference.com/w/cpp/types/is_integral), [Integral conversions](
 https://en.cppreference.com/w/cpp/language/implicit_conversion)). I have particularly tried that one, it works.

Considering that `char16_t` is binary equivalent to [uint_least16_t](https://en.cppreference.com/w/cpp/language/types), using `static_cast<const uint_least16_t>(value) << "\"";` would be the best choice (static cast from an unsigned integer to a signed integer was undefined until cpp20 if the number cannot be represented in the target type, does static_cast to `int16_t` might be an issue).

rosidl_generator_cpp/resource/msg__traits.hpp.em
```diff
@@ -61,6 +63,79 @@ for member in message.structure.members:
 namespace rosidl_generator_traits
 {
 
+void to_yaml(
```
ivanpauno, line 66: Should be made inline.

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
```diff
@@ -15,11 +15,138 @@
...
 #define ROSIDL_RUNTIME_CPP__TRAITS_HPP_
 
+#include <codecvt>
+#include <string>
 #include <type_traits>
 
 namespace rosidl_generator_traits
 {
 
+void value_to_yaml(bool value, std::ostream & out)
```
ivanpauno, line 25: Should be made inline.
Same for the other functions here.

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
dirk-thomas: > I have particularly tried that one, it works.

That is surprising. I did the same and got the following compiler error:

```
In file included from ws/build/rosidl_generator_cpp/rosidl_generator_cpp/rosidl_generator_cpp/msg/detail/empty__traits.hpp:9,
                 from ws/build/rosidl_generator_cpp/rosidl_generator_cpp/rosidl_generator_cpp/msg/empty.hpp:9,
                 from ws/src/ros2/rosidl/rosidl_generator_cpp/test/test_msg_datatype.cpp:16:
ws/install/rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp: In function ‘void rosidl_generator_traits::character_value_to_yaml(char16_t, std::ostream&)’:
ws/install/rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp:42:41: error: invalid static_cast from type ‘char16_t*’ to type ‘const int16_t*’ {aka ‘const short int*’}
   42 |     *static_cast<const int16_t *>(&value) << "\"";
      |                                         ^
make[2]: *** [CMakeFiles/test_msg_datatype.dir/build.make:63: CMakeFiles/test_msg_datatype.dir/test/test_msg_datatype.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:257: CMakeFiles/test_msg_datatype.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

rosidl_generator_cpp/test/test_traits.cpp
```diff
@@ -152,7 +156,11 @@ long_double_value: 1.12500
     msg.defaults_values.push_back(rosidl_generator_cpp::msg::Defaults());
     std::string yaml = to_yaml(msg);
 #ifdef _WIN32
-    std::replace(yaml.begin(), yaml.end(), "0.000000", "0.00000");
+    // update yaml to handle variance of floating point decimals on Windows
+    size_t index = 0;
+    while ((index = yaml.find("0.000000", index)) != std::string::npos) {
+      yaml = yaml.replace(index, 8, "0.00000");
```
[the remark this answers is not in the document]
dirk-thomas: No, because the previously found match is being removed by the `replace`.

rosidl_generator_cpp/resource/msg__traits.hpp.em
dirk-thomas: I don't see why. The idea is that all the implementations should be moved into a `.cpp` file anyway.

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
dirk-thomas: See same comment above.

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
ivanpauno: because you cannot `static_cast` the pointer, but the value:
```cpp
static_cast<int16_t>(value);
```
It's still problematic to `static_cast` an unsigned integer to a `signed` one, so `uint_least16_t` should be preferred:
```cpp
static_cast<uint_least16_t>(value);
```

I don't think this is a blocker for merging the PR, but I think we should minimize `reinterpret_cast` usage when possible (which in this case is).

dirk-thomas: Got it. Done in 5ffd92b1955315ffd0e6710f5bf68c54275b8b0c.

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
dirk-thomas: Done in c71a01c401d9b78cc563c86844dc2692c2953832.

rosidl_generator_cpp/resource/msg__traits.hpp.em
ivanpauno: Inline works as a linkage specifier and must always be used in functions defined in header files to avoid violating the "one definition rule" (ODR).
Methods and friend functions defined in a header files are implicitly inline.
See [ODR](https://en.cppreference.com/w/cpp/language/definition#One_Definition_Rule), [inline](https://en.cppreference.com/w/cpp/language/inline).
e.g.:
```cpp
# file1.cpp
#include "package_name/msg/my_msg.hpp"
...
```
```cpp
# file2.cpp
#include "package_name/msg/my_msg.hpp"
...
```
```cpp
# CMakeLists.txt
...
add_library(<lib_name>
  <path_to_file1.cpp> <path_to_file2.cpp>)
...
```

will violate ODR and produce linkage errors (this ODR violation can also happen with linking with a dependency, using `inline` avoids the problem).

> The idea is that all the implementations should be moved into a .cpp file anyway.

Yeah, so if we want to do that we should do it before galactic, as removing the `inline` specifier will break ABI.

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
```diff
@@ -130,7 +130,7 @@ void value_to_yaml(const std::u16string & value, std::ostream & out)
   auto flags = out.flags();
   size_t index = 0;
   while (index < value.size()) {
-    int16_t character = *reinterpret_cast<const int16_t *>(&value[index]);
+    int16_t character = static_cast<int16_t>(value[index]);
```
ivanpauno, line 133:
```suggestion
    int16_t character = static_cast<uint_least16_t>(value[index]);
```

rosidl_generator_cpp/resource/msg__traits.hpp.em
dirk-thomas: > Inline works as a linkage specifier and must always be used in functions defined in header files to avoid violating the "one definition rule" (ODR).

Indeed :+1:

> Yeah, so if we want to do that we should do it before galactic, as removing the inline specifier will break ABI.

That was the idea. But if preferred we can first add this API with `inline` and then break ABI when moving it into `.cpp` files: done in 77da551af6ab2a9647e16d67c580f66ed9db1b1c.

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
dirk-thomas: Applied in 63bf202799dff3d24aebbca55b2c49a0acb5214a.

ivanpauno (approved):
Looks great to me!

dirk-thomas:
CI builds testing `rosidl_generator_cpp`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12262/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7236/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9980/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12152/)

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
```diff
@@ -15,11 +15,138 @@
...
 
+inline void value_to_yaml(bool value, std::ostream & out)
+{
+  out << (value ? "true" : "false");
+}
+
+inline void character_value_to_yaml(unsigned char value, std::ostream & out)
+{
+  auto flags = out.flags();
+  out << "0x" << std::hex << std::setw(2) << std::setfill('0') << \
```
sloretz, line 33: This causes a downstream build failure in `rmw_dds_common`

https://ci.ros2.org/job/ci_linux/12266/console

```
--- stderr: rmw_dds_common
15:49:21 In file included from /home/jenkins-agent/workspace/ci_linux/ws/build/rmw_dds_common/rosidl_generator_cpp/rmw_dds_common/msg/detail/gid__traits.hpp:9,
15:49:21                  from /home/jenkins-agent/workspace/ci_linux/ws/build/rmw_dds_common/rosidl_generator_cpp/rmw_dds_common/msg/gid.hpp:9,
15:49:21                  from /home/jenkins-agent/workspace/ci_linux/ws/src/ros2/rmw_dds_common/rmw_dds_common/include/rmw_dds_common/gid_utils.hpp:21,
15:49:21                  from /home/jenkins-agent/workspace/ci_linux/ws/src/ros2/rmw_dds_common/rmw_dds_common/src/gid_utils.cpp:21:
15:49:21 /home/jenkins-agent/workspace/ci_linux/ws/install/rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp: In function ‘void rosidl_generator_traits::character_value_to_yaml(unsigned char, std::ostream&)’:
15:49:21 /home/jenkins-agent/workspace/ci_linux/ws/install/rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp:33:35: error: ‘setw’ is not a member of ‘std’
15:49:21    33 |   out << "0x" << std::hex << std::setw(2) << std::setfill('0') << \
15:49:21       |                                   ^~~~
15:49:21 /home/jenkins-agent/workspace/ci_linux/ws/install/rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp:33:51: error: ‘setfill’ is not a member of ‘std’; did you mean ‘fill’?
```

sloretz: Possibly just needs the `<iomanip>` include https://en.cppreference.com/w/cpp/io/manip/setw

PR #528, opened 2020-09-15
Added benchmark to rosidl_runtime_cpp bounded vector

ahcorde:
This PR builds on top of this other PR https://github.com/ros2/rosidl/pull/521. Added some benchmark test to the `BoundedVector` class 

Compiling up-to `rosidl_runtime_cpp` and testing `rosidl_runtime_cpp` with `-DAMENT_RUN_PERFORMANCE_TESTS=ON`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12316/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7285/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10029/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12211/)

Signed-off-by: ahcorde <ahcorde@gmail.com>

PR #529, opened 2020-09-15
Added benchmark test for rosidl_generator_c

ahcorde:
This PR builds on top of these other PRs
 - https://github.com/ros2/rosidl/pull/521
 - https://github.com/ros2/rosidl/pull/528

Addes some performance test to the interfaces `init()` and `destroy()` functions.

Signed-off-by: ahcorde <ahcorde@gmail.com>


cottsay (changes requested):
I don't understand why the tests are exercising array assignments. We're not really expecting C operations like `arr->bool_values[i] = true;` to change in performance, and it doesn't result in any lines of code in `rosidl_generator_c` being exercised, so I don't see any reason to track it here. `rosidl_runtime_c__String__assign` is already tracked by another test, so I don't see any reason to measure that performance either.

After looking at the generated code output from this package, I really only see allocation and de-allocation functions that could be worth testing. Am I missing something?

Issue #522, opened 2020-09-03
[rosidl_typesupport_introspection_c] get item function for arrays is semantically broken


ivanpauno:
See https://github.com/ros2/rmw_cyclonedds/pull/228#issuecomment-685177509.

---

Why wasn't this found before?
I think it deserves some explanation, as the function has been broken for years.

The "reference implementations" (cyclone and fastrtps dynamic) of the introspection typesupport were relying on conceptually broken cpp and were also leaking.
Instead of that, they should have been using the `resize` and `get` functions provided by the introspection typesupport.
They weren't using those functions at all and those functions don't have any test, thus the error wasn't found before.

When a fix was applied in https://github.com/ros2/rmw_cyclonedds/pull/228, it was found that the semantics of the "get_function" for arrays in the C introspection typesupport is different to the semantics of similar functions in the CPP introspection typesupport and also different to similar functions for bounded/unbounded sequences in the C typesupport.

See also:
- https://github.com/ros2/rmw_cyclonedds/issues/219
- https://github.com/ros2/rmw_fastrtps/issues/428

ivanpauno:
@iuhilnehc-ynos FYI

PR #531, opened 2020-09-24
[rosidl_typesupport_introspection_c] Fix get_function and get_const_function semantics for arrays

ivanpauno:
Fixes https://github.com/ros2/rosidl/issues/522.
For all arrays/sequences/bounded_sequences in rosidl_typesupport_introspection_c/cpp, the following code works:

```c
member->get_function(ros_message + member->offset_, index)
```
except for rosidl_typesupport_introspection_c arrays, where the correct code is:

```c
member->get_function(&(ros_message + member->offset_), index)
```

This PR fixes the semantic difference.

The following PRs must be merged simultaneously:
- https://github.com/ros2/rmw_cyclonedds/pull/248
- https://github.com/ros2/rmw_fastrtps/pull/448
- https://github.com/ros2/ros2_documentation/pull/877.

This error wasn't detected before, because both `rmw_fastrtps_dynamic_cpp` and `rmw_cyclonedds_cpp` were doing "weird stuff" instead of using the "get function" and "resize function" provided by the introspection typesupport.
For further reference, see:
- ros2/rmw_cyclonedds#219
- ros2/rmw_fastrtps#428

### DIFF
--- a/rosidl_typesupport_introspection_c/resource/msg__type_support.c.em
+++ b/rosidl_typesupport_introspection_c/resource/msg__type_support.c.em
@@ -137,9 +137,9 @@ const void * @(function_prefix)__get_const_function__@(member.type.value_type.na
   const void * untyped_member, size_t index)
 {
 @[    if isinstance(member.type, Array)]@
-  const @('__'.join(member.type.value_type.namespaced_name())) ** member =
-    (const @('__'.join(member.type.value_type.namespaced_name())) **)(untyped_member);
-  return &(*member)[index];
+  const @('__'.join(member.type.value_type.namespaced_name())) * member =
+    (const @('__'.join(member.type.value_type.namespaced_name())) *)(untyped_member);
+  return &member[index];
 @[    else]@
   const @('__'.join(member.type.value_type.namespaced_name()))__Sequence * member =
     (const @('__'.join(member.type.value_type.namespaced_name()))__Sequence *)(untyped_member);
@@ -151,9 +151,9 @@ void * @(function_prefix)__get_function__@(member.type.value_type.name)__@(membe
   void * untyped_member, size_t index)
 {
 @[    if isinstance(member.type, Array)]@
-  @('__'.join(member.type.value_type.namespaced_name())) ** member =
-    (@('__'.join(member.type.value_type.namespaced_name())) **)(untyped_member);
-  return &(*member)[index];
+  @('__'.join(member.type.value_type.namespaced_name())) * member =
+    (@('__'.join(member.type.value_type.namespaced_name())) *)(untyped_member);
+  return &member[index];
 @[    else]@
   @('__'.join(member.type.value_type.namespaced_name()))__Sequence * member =
     (@('__'.join(member.type.value_type.namespaced_name()))__Sequence *)(untyped_member);


dirk-thomas:
> The following PRs must be merged simultaneously:

This error wasn't detected before, because both `rmw_fastrtps_dynamic_cpp` and `rmw_cyclonedds_cpp` were doing "weird stuff" instead of using the "get function" and "resize function" provided by the introspection typesupport.

dirk-thomas:
> The following PRs must be merged simultaneously:

How about Connext? If it doesn't need the change is its current behavior broken?

> This error wasn't detected before, because both `rmw_fastrtps_dynamic_cpp` and `rmw_cyclonedds_cpp` were doing "weird stuff" instead of using the "get function" and "resize function" provided by the introspection typesupport.

Both were using the provided function. And the "weird stuff" was simply accounting the difference in the API - to exactly pass what was needed.

Certainly unifying the signature makes sense.

dirk-thomas:
> The following PRs must be merged simultaneously:

These changes will also need to be released into Rolling.

dirk-thomas (approved):

ivanpauno:
> How about Connext? If it doesn't need the change is its current behavior broken?

rmw_connexy_dynamic_cpp is not supported, and the code was not handling the semantic difference correctly https://github.com/ros2/rmw_connext/blob/c0a12cd0c6b8a14191763417e2ee13bbd9624492/rmw_connext_dynamic_cpp/src/templates.hpp#L1271.

> Both were using the provided function. And the "weird stuff" was simply accounting the difference in the API - to exactly pass what was needed.

They weren't using those functions, that was fixed in https://github.com/ros2/rmw_cyclonedds/pull/228, https://github.com/ros2/rmw_fastrtps/pull/429. The code handling the introspection typesupport before those PRs is interesting.

> These changes will also need to be released into Rolling.

:+1:

dirk-thomas:
> rmw_connexy_dynamic_cpp is not supported, and the code was not handling the semantic difference correctly https://github.com/ros2/rmw_connext/blob/c0a12cd0c6b8a14191763417e2ee13bbd9624492/rmw_connext_dynamic_cpp/src/templates.hpp#L1271.

I would suggest to still update the code (since it is trivial to do now) and probably hard to figure out that it is needed in the future.

ivanpauno:
> I would suggest to still update the code (since it is trivial to do now) and probably hard to figure out that it is needed in the future.

Yeah, what I mean is that `rmw_connext_dynamic_cpp` doesn't need a patch. That code was always using:
```
member->get_function(ros_message + member->offset_, index)
```

dirk-thomas:
> Yeah, what I mean is that rmw_connext_dynamic_cpp doesn't need a patch. That code was always using:

Oh, got it now. Thanks for clarifying. So the behavior must have been broken in the past then. :white_check_mark:

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12450/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7417/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10160/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12365/)

ivanpauno:
rosdistro PRs:
- https://github.com/ros/rosdistro/pull/26688
- https://github.com/ros/rosdistro/pull/26689
- https://github.com/ros/rosdistro/pull/26690

PR #538, opened 2020-10-20
Strip action service suffixes from C include prefix

jacobperron:
This adds logic for the two missing services that are part of an action: `send_goal` and `get_result`.
For example, "fibonacci__send_goal.h" is not a valid header and we should instead use "fibonacci.h".

C++ and Python generators are not affected by this change since they combine all parts of an
action definition into a single file.

But generators like rosidl_generator_java create separate files for the service and message
definitions composing an action, and can take advantage of this common logic.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12794/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7753/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10508/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12741/)

### DIFF
--- a/rosidl_generator_c/rosidl_generator_c/__init__.py
+++ b/rosidl_generator_c/rosidl_generator_c/__init__.py
@@ -87,6 +87,10 @@ def idl_structure_type_to_c_include_prefix(namespaced_type, subdirectory=None):
         include_prefix = include_prefix[:-8]
     elif include_prefix.endswith('__feedback'):
         include_prefix = include_prefix[:-10]
+    elif include_prefix.endswith('__send_goal'):
+        include_prefix = include_prefix[:-11]
+    elif include_prefix.endswith('__get_result'):
+        include_prefix = include_prefix[:-12]
     return include_prefix
 
 


jacobperron:
For context, I had to apply a similar patch to the `rosidl_generator_java` templates in https://github.com/ros2-java/ros2_java/pull/140.

clalancette:
@jacobperron Could you give a quick example of what the generated output would look like before and after this change?  It is not entirely clear to me.

ivanpauno:
> @jacobperron Could you give a quick example of what the generated output would look like before and after this change? It is not entirely clear to me.

AFAIU:

```python3
# valid before this PR (and after)
assert (
  idl_structure_type_to_c_include_prefix(
    NamespacedType(('std_srvs', 'srv'), 'Empty__request')) ==
  'std_srvs/srv/empty'
)
assert (
  idl_structure_type_to_c_include_prefix(
    NamespacedType(('action_tutorials_interfaces', 'action'), 'Fibonacci__feedback')) ==
  'action_tutorials_interfaces/action/fibonacci'
)
# after this PR, the following is also valid:
assert (
  idl_structure_type_to_c_include_prefix(
    NamespacedType(('action_tutorials_interfaces', 'action'), 'Fibonacci__send_goal')) ==
  'action_tutorials_interfaces/action/fibonacci'
)
```

I'm not completely sure though

rosidl_generator_c/rosidl_generator_c/__init__.py
```diff
@@ -87,6 +87,10 @@ def idl_structure_type_to_c_include_prefix(namespaced_type, subdirectory=None):
         include_prefix = include_prefix[:-8]
     elif include_prefix.endswith('__feedback'):
         include_prefix = include_prefix[:-10]
+    elif include_prefix.endswith('__send_goal'):
+        include_prefix = include_prefix[:-11]
+    elif include_prefix.endswith('__get_result'):
+        include_prefix = include_prefix[:-12]
```
ivanpauno, line 93: unrelated improvement, but I would prefer not counting characters:

```python3
suffixes_to_strip = [
  '__request',
  ....
  '__get_result'
]
for suf in suffixes_to_strip:
  if include_prefix.endswith(suf):
    return include_prefix[:-len(suf)]
```

ivanpauno (approved):
After checking https://github.com/ros2-java/ros2_java/pull/140, this seems a valuable thing to do here.

jacobperron:
> Could you give a quick example of what the generated output would look like before and after this change? It is not entirely clear to me.

@clalancette Yep, Ivan's example is correct. AFAIK, this change shouldn't affect the generated output of any C++ or Python code, but it's useful in other generators like Java (e.g. https://github.com/ros2-java/ros2_java/pull/140).

rosidl_generator_c/rosidl_generator_c/__init__.py
jacobperron: I suggest we do that kind of refactor in a separate PR.

PR #542, opened 2020-10-28
Fix parsing of floating_pt_literal and fixed_pt_literal

### LINKED ISSUE #541: Fails to build messages with float type and small default values
## Bug report

**Required Info:**

- Operating System:
  - Kubuntu 20.04 
- Installation type:
  - binaries 
- Version or commit hash:
  - ros-foxy latest
- DDS implementation:
  - Fast-RTPS 
- Client library (if applicable):
  - N/A 

#### Steps to reproduce issue

Create a message field with a default value smaller than 0.0001: 
```
float64 very_small_number 0.00009
```

#### Expected behavior
successful build
#### Actual behavior
```
  File "/opt/ros/foxy/share/rosidl_generator_cpp/cmake/../../../lib/rosidl_generator_cpp/rosidl_generator_cpp", line 40, in <module>
    sys.exit(main())
  File "/opt/ros/foxy/share/rosidl_generator_cpp/cmake/../../../lib/rosidl_generator_cpp/rosidl_generator_cpp", line 34, in main
    return generate_cpp(
  File "/opt/ros/foxy/lib/python3.8/site-packages/rosidl_generator_cpp/__init__.py", line 38, in generate_cpp
    generate_files(
  File "/opt/ros/foxy/lib/python3.8/site-packages/rosidl_cmake/__init__.py", line 96, in generate_files
    raise(e)
  File "/opt/ros/foxy/lib/python3.8/site-packages/rosidl_cmake/__init__.py", line 74, in generate_files
    idl_file = parse_idl_file(locator)
  File "/opt/ros/foxy/lib/python3.8/site-packages/rosidl_parser/parser.py", line 64, in parse_idl_file
    content = parse_idl_string(string, png_file=png_file)
  File "/opt/ros/foxy/lib/python3.8/site-packages/rosidl_parser/parser.py", line 73, in parse_idl_string
    content = extract_content_from_ast(tree)
  File "/opt/ros/foxy/lib/python3.8/site-packages/rosidl_parser/parser.py", line 158, in extract_content_from_ast
    add_message_members(request, struct_defs[0])
  File "/opt/ros/foxy/lib/python3.8/site-packages/rosidl_parser/parser.py", line 367, in add_message_members
    annotations = get_annotations(member)
  File "/opt/ros/foxy/lib/python3.8/site-packages/rosidl_parser/parser.py", line 513, in get_annotations
    get_const_expr_value(const_expr)
  File "/opt/ros/foxy/lib/python3.8/site-packages/rosidl_parser/parser.py", line 528, in get_const_expr_value
    assert len(expr) == 1, str(expr)
AssertionError: [Tree(primary_expr, [Tree(literal, [Tree(floating_pt_literal, [Token(DIGIT, '9'), Tree(floating_pt_literal_e, [])])])]), Tree(primary_expr, [Tree(literal, [Tree(integer_literal, [Tree(octal_literal, [Token(__ANON_1, '5')])])])])]
```

clalancette:
The problem comes about because the msg -> idl transformer
can generate IDL defaults like "9e-05", which the lark
grammar wasn't ready to handle.  But Lark provides a FLOAT
terminal type, which *does* handle it, so use that instead.

Also add a test for this situation.

@dirk-thomas Looking for any insight from you why we didn't use the FLOAT builtin before.  Maybe it was an oversight, but maybe there was a reason.  Once I have some idea of whether this is the right thing to do, I'll run a full CI run on all messages, services, and actions just to make sure nothing breaks here.

Should fix #541 .

### DIFF
--- a/rosidl_parser/rosidl_parser/grammar.lark
+++ b/rosidl_parser/rosidl_parser/grammar.lark
@@ -8,9 +8,11 @@
 // 7.4.15.4.2 Applying Annotations
 //   Only supported on modules, structs, members, enums and enum identifiers
 
+%import common.DECIMAL
 %import common.DIGIT
 %import common.FLOAT
 %import common.HEXDIGIT
+%import common.INT
 %import common.LETTER
 %import common.WS
 %ignore WS
@@ -79,10 +81,8 @@ wide_string_literal: "L\"\"" | "L\"" /(\\\"|[^"])+/ "\""
 floating_pt_literal: FLOAT
 
 // 7.2.6.5 Fixed-Point Literals
-fixed_pt_literal: DIGIT+ "." DIGIT+ "d"i
-  |"." DIGIT+ "d"i
-  | DIGIT+ "." "d"i
-  | DIGIT+ "d"i
+fixed_pt_literal: DECIMAL "d"i
+  | INT "d"i
 
 
 // 7.3 Preprocessing
--- a/rosidl_parser/rosidl_parser/grammar.lark
+++ b/rosidl_parser/rosidl_parser/grammar.lark
@@ -9,6 +9,7 @@
 //   Only supported on modules, structs, members, enums and enum identifiers
 
 %import common.DIGIT
+%import common.FLOAT
 %import common.HEXDIGIT
 %import common.LETTER
 %import common.WS
@@ -75,13 +76,7 @@ string_literal: "\"\"" | "\"" /(\\\"|[^"])+/ "\""
 wide_string_literal: "L\"\"" | "L\"" /(\\\"|[^"])+/ "\""
 
 // 7.2.6.4 Floating-point Literals
-floating_pt_literal: DIGIT+ floating_pt_literal_dot DIGIT+ floating_pt_literal_e DIGIT*
-  | floating_pt_literal_dot DIGIT+ (floating_pt_literal_e DIGIT*)?
-  | DIGIT+ floating_pt_literal_dot DIGIT* (floating_pt_literal_e DIGIT*)?
-  | DIGIT+ floating_pt_literal_e DIGIT*
-// separate rules to identify the components
-floating_pt_literal_dot: "."
-floating_pt_literal_e: "e"i
+floating_pt_literal: FLOAT
 
 // 7.2.6.5 Fixed-Point Literals
 fixed_pt_literal: DIGIT+ "." DIGIT+ "d"i
--- a/rosidl_parser/rosidl_parser/parser.py
+++ b/rosidl_parser/rosidl_parser/parser.py
@@ -588,10 +588,6 @@ def get_floating_pt_literal_value(floating_pt_literal):
     for child in floating_pt_literal.children:
         if isinstance(child, Token):
             value += child.value
-        elif child.data == 'floating_pt_literal_dot':
-            value += '.'
-        elif child.data == 'floating_pt_literal_e':
-            value += 'e'
         else:
             assert False, 'Unsupported tree: ' + str(floating_pt_literal)
     return float(value)
--- a/rosidl_parser/rosidl_parser/parser.py
+++ b/rosidl_parser/rosidl_parser/parser.py
@@ -557,6 +557,12 @@ def get_const_expr_value(const_expr):
                 value = -value
             return value
 
+        if child.data == 'fixed_pt_literal':
+            value = get_fixed_pt_literal_value(child)
+            if negate_value:
+                value = -value
+            return value
+
         if child.data == 'boolean_literal':
             assert len(child.children) == 1
             child = child.children[0]
@@ -590,6 +596,16 @@ def get_floating_pt_literal_value(floating_pt_literal):
             value += child.value
         else:
             assert False, 'Unsupported tree: ' + str(floating_pt_literal)
+    return float(value)
+
+
+def get_fixed_pt_literal_value(fixed_pt_literal):
+    value = ''
+    for child in fixed_pt_literal.children:
+        if isinstance(child, Token):
+            value += child.value
+        else:
+            assert False, 'Unsupported tree: ' + str(fixed_pt_literal)
     return float(value)
 
 
--- a/rosidl_parser/test/msg/MyMessage.idl
+++ b/rosidl_parser/test/msg/MyMessage.idl
@@ -71,6 +71,16 @@ module rosidl_parser {
       float int_with_explicit_positive_scientific;
       @default ( value=9e-05 )
       float int_with_negative_scientific;
+
+      // Tests of the fixed point parser (7.2.6.5)
+      @default ( value=8.7d )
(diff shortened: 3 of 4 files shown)


rosidl_parser/rosidl_parser/grammar.lark
```diff
@@ -75,13 +76,7 @@ string_literal: "\"\"" | "\"" /(\\\"|[^"])+/ "\""
...
 
 // 7.2.6.4 Floating-point Literals
-floating_pt_literal: DIGIT+ floating_pt_literal_dot DIGIT+ floating_pt_literal_e DIGIT*
-  | floating_pt_literal_dot DIGIT+ (floating_pt_literal_e DIGIT*)?
-  | DIGIT+ floating_pt_literal_dot DIGIT* (floating_pt_literal_e DIGIT*)?
-  | DIGIT+ floating_pt_literal_e DIGIT*
-// separate rules to identify the components
-floating_pt_literal_dot: "."
-floating_pt_literal_e: "e"i
+floating_pt_literal: FLOAT
```
dirk-thomas, line 79: `FLOAT` is defined as follows:

> `FLOAT: INT _EXP | DECIMAL _EXP?`

and the latter part doesn't match the definition of `floating_pt_literal` in the IDL spec. Therefore I don't think it can be used as is.

clalancette: Hm, I'm not sure I agree.  7.2.6.4 in the IDL spec says:

```
A floating-point literal consists of an integer part, a decimal point (.), a fraction part, an e or E, and an optionally
signed integer exponent. The integer and fraction parts both consist of a sequence of decimal (base ten) digits. Either
the integer part or the fraction part (but not both) may be missing; either the decimal point or the letter e (or E) and the
exponent (but not both) may be missing.
```

That seems to say to me that the following cases are allowed (everything else is denied):
* "1.1e10"
* "1.1e-10"
* "1.1"
* "1."
* ".1"
* "1e-5"
* "1e5"

Looking through the grammar for the Lark built-in FLOAT, it looks like:

```
INT: DIGIT+
SIGNED_INT: ["+"|"-"] INT
DECIMAL: INT "." INT? | "." INT

// float = /-?\d+(\.\d+)?([eE][+-]?\d+)?/
_EXP: ("e"|"E") SIGNED_INT
FLOAT: INT _EXP | DECIMAL _EXP?
```

That seems to allow the same cases.  However, I might be missing something, so could you explain a case that doesn't work with that grammar, but is allowed by IDL (or vice-versa)?

dirk-thomas: `FLOAT` allows a simple `DECIMAL` which in the IDL spec is not a `floating_pt_literal` but a `fixed_pt_literal`.

clalancette: But `7.2.6.5` says:

```
A fixed-point decimal literal consists of an integer part, a decimal point (.) , a fraction part and a d or D. The integer and fraction parts both consist of a sequence of decimal (base 10) digits. Either the integer part or the fraction part (but not both) may be missing; the decimal point (but not the letter d or D) may be missing.
```

So my understanding is that any fixed point has to end with a literal `d` or `D`, which is not the case for FLOAT.

dirk-thomas: You are right that 7.2.6.5 / `fixed_pt_literal` is different since it must have a trailing `d` / `D`.

My previous point about 7.2.6.4 still stands: 

> `FLOAT` allows a simple `DECIMAL` which in the IDL spec is not a `floating_pt_literal` (7.2.6.4)

7.2.6.4 says:

> ... either the decimal point or the letter e (or E ) and the
exponent (**but not both**) may be missing.

clalancette: > `FLOAT` allows a simple `DECIMAL` which in the IDL spec is not a `floating_pt_literal` (7.2.6.4)

Just to be clear: DECIMAL is not the same thing as INT.  DECIMAL is defined in the lark grammar as:

```
DECIMAL: INT "." INT? | "." INT
```

That is, "5.", "5.0", or ".5" are what is allowed.  An integer such as "5" won't match a DECIMAL (though it would match an INT).

And since `FLOAT` is defined as:

```
FLOAT: INT _EXP | DECIMAL _EXP?
```

That takes care of having either an `e|E` or a `.`, or both, but not neither.  Thus, I still think that we can reuse FLOAT here.

If you still don't agree, can you give me an example of a number that you think is allowed by the IDL spec, but is not allowed for lark's FLOAT type?  Or one that is not allowed by the IDL spec, but is allowed for lark's FLOAT type?

dirk-thomas: > And since `FLOAT` is defined as:
> `FLOAT: INT _EXP | DECIMAL _EXP?`
> That takes care of having either an `e|E` or a `.`, or both, but not neither.  Thus, I still think that we can reuse FLOAT here.

:+1:  I think I got it now and agree with you on: `floating_pt_literal: FLOAT`.

rosidl_parser/rosidl_parser/grammar.lark
```diff
@@ -75,13 +76,7 @@ string_literal: "\"\"" | "\"" /(\\\"|[^"])+/ "\""
...
 
 // 7.2.6.5 Fixed-Point Literals
 fixed_pt_literal: DIGIT+ "." DIGIT+ "d"i
```
dirk-thomas, line 82: The definition of `fixed_pt_literal` seems incorrect since it doesn't allow the following part of 7.2.6.5:

> Either the integer part or the fraction part (but not
both) may be missing; the decimal point (but not the letter d or D ) may be missing.

Proposal: `fixed_pt_literal: DECIMAL "d"i`

clalancette: You are right, the fixed point literal also wasn't quite correct.  However, I ended up with the following grammar instead:

```
fixed_pt_literal: DECIMAL "d"i
  | INT "d"i
```

That's to take care of the case where no dot is in the number, as allowed by 7.2.6.5.  See 77f0373 for fixes.  I also added some more test cases to the floating_pt_literal.  I think this is now ready for review, so switching it to a real PR.

dirk-thomas (approved):
This should probably be labels as a `bug` and added to the various project boards for release / backport.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12924/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7876/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10640/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12872/)

clalancette:
The failing tests on Windows happened in the nightlies too.  So I'm going ahead and merging this, thanks for the review.

PR #543, opened 2020-11-02
Use ASSERT_TRUE to check for nullptr.

clalancette:
clang static analysis reports this as a false positive for
a possible memory leak.  This is incorrect, but the FAQ
for googletest also has this tidbit:

https://github.com/google/googletest/blob/master/googletest/docs/faq.md#why-does-googletest-support-expect_eqnull-ptr-and-assert_eqnull-ptr-but-not-expect_nenull-ptr-and-assert_nenull-ptr

Thus, just switch to ASSERT_TRUE for the check, which is
good enough and makes clang static analysis happier.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12939/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7887/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10655/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12887/)

sloretz (approved):

PR #548, opened 2020-11-10
[eloquent] refactor regex for valid package/field names (#508)

jacobperron:
Backport #508 to Eloquent.

### DIFF
--- a/rosidl_adapter/rosidl_adapter/parser.py
+++ b/rosidl_adapter/rosidl_adapter/parser.py
@@ -57,8 +57,14 @@ PRIMITIVE_TYPES = [
     'time',  # for compatibility only
 ]
 
-VALID_PACKAGE_NAME_PATTERN = re.compile('^[a-z]([a-z0-9_]?[a-z0-9]+)*$')
-VALID_FIELD_NAME_PATTERN = re.compile('^[a-z]([a-z0-9_]?[a-z0-9]+)*$')
+VALID_PACKAGE_NAME_PATTERN = re.compile(
+    '^'
+    '(?!.*__)'  # no consecutive underscores
+    '(?!.*_$)'  # no underscore at the end
+    '[a-z]'  # first character must be alpha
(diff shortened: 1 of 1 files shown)


jacobperron:
Testing `rosidl_adapter` and packages that recursively depend on it:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13205/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8136/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10921/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13223/)

clalancette:
I have to say that I'm concerned about putting this change in right at the end of Eloquent's life.  While I know it fixes an issue, and it's been in Rolling for a while, regexes are tricky.  I just don't want to break an application that already works in Eloquent.

Is there a compelling bug report that has come in for Eloquent for us to make this change?

jacobperron:
There's been no explicit bug report for Eloquent, but I suspect the original reported bug for Foxy (https://github.com/colcon/colcon-core/issues/364) also exists in Eloquent. FWIW, we've also backported this change to Dashing (https://github.com/ros2/rosidl/pull/549).

clalancette:
> There's been no explicit bug report for Eloquent, but I suspect the original reported bug for Foxy ([colcon/colcon-core#364](https://github.com/colcon/colcon-core/issues/364)) also exists in Eloquent. 

Got it.

> FWIW, we've also backported this change to Dashing (#549).

Right.  It definitely makes sense for Dashing.  I'm more concerned about Eloquent, since we have no vehicle with which to fix it if we find a problem.

But I'm willing to be persuaded otherwise.  @mjcarroll thoughts?

mjcarroll:
Yeah, I think I'm a little concerned dropping this one in without another sync/patch release on Eloquent.  I'll add it to known issues for Eloquent.

PR #555, opened 2020-11-19
Remove unnecessary assert on pointer created with new

brawner:
This return path was flagged by clang static analysis as potentially leaking memory. I'm pretty sure it won't, but clang was a bit confused by it I think. In either case, new will throw an exception if it can't allocate instead of returning NULL.

Signed-off-by: Stephen Brawner <brawner@gmail.com>


clalancette (approved):
Looks good with green CI.

brawner:
Testing `--packages-select rosidl_generator_cpp`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13128/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8059/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10837/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13122/)

PR #557, opened 2020-12-21
Treat \t as whitespace

### LINKED ISSUE #556: msg parser does not consider tabs as whitespace
When building a message files with a tab separating the type and field name the tab and field name are considered as a single symbol rather than two symbols.

Example with tab fails to build:
```
uint32	BINARYGROUP_COMMON      = 01
```	
Example with space builds ok:
```
uint32 BINARYGROUP_COMMON      = 01
```	

- Operating System: U20.04
- Installation type: Bin
- Version or commit hash: foxy
- DDS implementation: fast_rtps
- Client library (if applicable): N/A

dawonn:
Fixes #556.

### DIFF
--- a/rosidl_adapter/rosidl_adapter/parser.py
+++ b/rosidl_adapter/rosidl_adapter/parser.py
@@ -459,6 +459,9 @@ def parse_message_string(pkg_name, msg_name, message_string):
     for line in lines:
         line = line.rstrip()
 
+        # replace tabs with spaces
+        line = line.replace('\t', ' ')
+
         # ignore empty lines
         if not line:
             # file-level comments stop at the first empty line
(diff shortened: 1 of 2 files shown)


clalancette (changes requested):
Treating tabs as spaces seems reasonable enough to me.  This change also seems OK.

Can you please add a test for this?

clalancette (approved):
Seems reasonable to me.  I'll run CI on it next.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13341/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8270/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11072/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13392/)

clalancette:
The failures on all platforms are known, so this looks good to go.  Thanks for the contribution, merging.

PR #559, opened 2020-12-22
Add hex support for msg constants

### LINKED ISSUE #558: support msg constants in hex
It would be great to support constants in hex notation, esp for bitmasks in my case:

```
uint32	BINARYGROUP_COMMON = 0x01		
uint32	BINARYGROUP_TIME = 0x02
```

- Operating System: U20.04
- Installation type: binary
- Version or commit hash: foxy
- DDS implementation: fast_rtps
- Client library (if applicable): N/A

dawonn:
fixes #558

### DIFF
--- a/rosidl_adapter/rosidl_adapter/parser.py
+++ b/rosidl_adapter/rosidl_adapter/parser.py
@@ -705,7 +705,11 @@ def parse_primitive_value_string(type_, value_string):
         try:
             value = int(value_string)
         except ValueError:
-            raise ex
+            try:
+                value = int(value_string, 0)
+            except ValueError:
+                raise ex
+
         if value < 0 or value > 255:
             raise ex
         return value
@@ -737,7 +741,10 @@ def parse_primitive_value_string(type_, value_string):
         try:
             value = int(value_string)
         except ValueError:
-            raise ex
+            try:
+                value = int(value_string, 0)
+            except ValueError:
+                raise ex
 
         # check that value is in valid range
         if value < lower_bound or value > upper_bound:
--- a/rosidl_adapter/test/test_parse_primitive_value_string.py
+++ b/rosidl_adapter/test/test_parse_primitive_value_string.py
@@ -85,6 +85,120 @@ def test_parse_primitive_value_string_integer():
                 Type(integer_type), str(upper_bound + 1))
 
 
+def test_parse_primitive_value_string_hex():
+    integer_types = {
+        'byte': [8, True],
+        'char': [8, True],
+        'int8': [8, False],
+        'uint8': [8, True],
+        'int16': [16, False],
+        'uint16': [16, True],
+        'int32': [32, False],
+        'uint32': [32, True],
+        'int64': [64, False],
+        'uint64': [64, True],
+    }
+    for integer_type, (bits, is_unsigned) in integer_types.items():
+        lower_bound = 0 if is_unsigned else -(2 ** (bits - 1))
+        upper_bound = (2 ** (bits if is_unsigned else (bits - 1))) - 1
+
+        value = parse_primitive_value_string(
+            Type(integer_type), hex(lower_bound))
+        assert value == lower_bound
+        value = parse_primitive_value_string(
+            Type(integer_type), hex(0))
+        assert value == 0
+        value = parse_primitive_value_string(
+            Type(integer_type), hex(upper_bound))
+        assert value == upper_bound
+
+        with pytest.raises(InvalidValue):
+            parse_primitive_value_string(
+                Type(integer_type), 'value')
+        with pytest.raises(InvalidValue):
+            parse_primitive_value_string(
+                Type(integer_type), hex(lower_bound - 1))
+        with pytest.raises(InvalidValue):
+            parse_primitive_value_string(
+                Type(integer_type), hex(upper_bound + 1))
+
+
+def test_parse_primitive_value_string_oct():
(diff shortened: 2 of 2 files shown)


clalancette:
@dawonn I think the idea makes sense, but I would like to:

1. Get input from the rest of the @ros2/team 
2. Figure out whether this was allowed in ROS 1 (and if not, why not)
3. Figure out whether there is a specification for the ROS 1 message types, and if so, update it
4. Have tests for this functionality

dawonn:
Sounds good, I'll work on the tests in the next day or two.

FWIW: This feature would really help with hardware driver interfaces that I'm developing so that the values match across the device API and User Manuals.

sloretz:
> 3. Figure out whether there is a specification for the ROS message types, and if so, update it

http://design.ros2.org/articles/legacy_interface_definition.html

rosidl_adapter/rosidl_adapter/parser.py
```diff
@@ -459,6 +459,9 @@ def parse_message_string(pkg_name, msg_name, message_string):
     for line in lines:
         line = line.rstrip()
 
+        # replace tabs with spaces
+        line = line.replace('\t', ' ')
```
jacobperron, line 463: Is this relevant to supporting hex values?

jacobperron (commented):
Seems okay to me.

jacobperron:
Technically, this change also adds support for base-2 and base-8 constants too, e.g. `0b01` and `0o01`.

rosidl_adapter/rosidl_adapter/parser.py
clalancette: I think this is pulling #557 into this PR.  I'll suggest dropping it from this one, so we can consider them separately.

clalancette:
> > 1. Figure out whether there is a specification for the ROS message types, and if so, update it
> 
> http://design.ros2.org/articles/legacy_interface_definition.html

Great, thanks @sloretz .  @dawonn So I think the first step here is to update https://github.com/ros2/design/blob/gh-pages/articles/116_legacy_interface_definition.md to add hex, binary, and octal into the "constants" section.  Once we have that, then we can go ahead and get the actual code change in.

rosidl_adapter/rosidl_adapter/parser.py
dawonn: dropped.

dawonn:
1. https://github.com/ros2/design/pull/308
2. Tests added for bin, oct, & hex.

gavanderhoorn:
@jacobperron wrote:
> Technically, this change also adds support for base-2 and base-8 constants too, e.g. `0b01` and `0o01`.

Wouldn't the grammar need to be updated as well?

https://github.com/ros2/rosidl/blob/5b74acd256f0f9cf92f279cea4845f991a48776b/rosidl_parser/rosidl_parser/grammar.lark#L37-L40

The IDL spec also doesn't appear to allow for any other bases than 8, 10 and 16.

sloretz:
> Wouldn't the grammar need to be updated as well?

IIUC this is adding a feature to the legacy ROS 1 style msg/srv/action format, not the IDL parser. The grammar for IDL parsing should be unchanged by this PR.

jacobperron:
> IIUC this is adding a feature to the legacy ROS 1 style msg/srv/action format, not the IDL parser. The grammar for IDL parsing should be unchanged by this PR.

That sounds right to me.

To be sure, it might be good to extend our [test interfaces files](https://github.com/ros2/test_interface_files) to exercise this new feature (e.g. add new fields to `Constants.msg`). Depending which message definitions are changed, it might also involve updating the test fixtures [here](https://github.com/ros2/rcl_interfaces/blob/836c8653825f0cf6e6ee46d6a3837b370678004b/test_msgs/src/test_msgs/message_fixtures.py) and [here](https://github.com/ros2/rcl_interfaces/blob/master/test_msgs/include/test_msgs/message_fixtures.hpp).

gavanderhoorn:
Wouldn't that make it possible to create `.msg` (etc) files which would be problematic when considering their "IDL use"? Or is this something where ROS does not need to follow the IDL spec?

sloretz:
> Wouldn't that make it possible to create .msg (etc) files which would be problematic when considering their "IDL use"? Or is this something where ROS does not need to follow the IDL spec?

I think using `int` here in `rosidl_adapter` means the constants written to the OMG IDL file would always be base 10 regardless of the base used in the `.msg` file.

jacobperron:
> I think using int here in rosidl_adapter means the constants written to the OMG IDL file would always be base 10 regardless of the base used in the .msg file.

This is what I thought too. As I suggest, modifying our current test files would help confirm.

clalancette (approved):
This looks like it is backwards compatible, and faithfully implements https://github.com/ros2/design/pull/308 .  I'm going to run full CI on it just to make sure nothing breaks.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13627/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8509/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11342/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13688/)

clalancette:
The failures on macOS and Windows are well-known, so not caused by this PR.  Going ahead and merging, thanks @dawonn

PR #562, opened 2020-12-28
Remove unnecessary short and unsigned short for cpp generator

seendsouza:
According to this: https://index.ros.org/doc/ros2/Concepts/About-ROS-Interfaces/
short (int16) and unsigned short (uint16) are DDS types not field type names.

Signed-off-by: Sean D'Souza <sdammobubbles@gmail.com>

### DIFF
--- a/rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
+++ b/rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
@@ -193,7 +193,6 @@ def primitive_value_to_cpp(type_, value):
         return 'true' if value else 'false'
 
     if type_.typename in [
-        'short', 'unsigned short',
         'char', 'wchar',
         'double', 'long double',
         'octet',


sloretz (changes requested):
I think these types are necessary.

ROS 2 message generation uses the OMG IDL file format (design doc of subset used by ROS 2 here: https://design.ros2.org/articles/idl_interface_definition.html). The ROS IDL (sometimes called the [legacy interface](https://design.ros2.org/articles/legacy_interface_definition.html)) `.msg`, `.srv`, `.action` files are converted to OMG IDL files before they're given to the generators, and it's also possible to define messages in the OMG IDL format directly.

If I understand correctly, this PR would remove the ability to generate C++ code [for `int16` and `uint16` types in the legacy format](https://github.com/ros2/test_interface_files/blob/124823291a0a7d35e983b2e00abc8803ec4cdbb2/msg/BasicTypes.msg#L8-L9), and would remove the ability to generate code for the `short` and `unsigned short` fields if one uses the OMG IDL spect directly.

seendsouza:
My bad. I wasn't aware that short and unsigned short were part of the OMG IDL format. I'll close the PR then.

PR #564, opened 2021-01-16
Switch to std::allocator_traits.

clalancette:
In C++17, Allocator::rebind<>::other is deprecated (and Windows
now warns about it).  See
https://en.cppreference.com/w/cpp/memory/allocator for more information.
Switch to std::allocator_traits which works in both C++14 and C++17.

### DIFF
--- a/rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
+++ b/rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
@@ -68,9 +68,9 @@ MSG_TYPE_TO_CPP = {
     'uint64': 'uint64_t',
     'int64': 'int64_t',
     'string': 'std::basic_string<char, std::char_traits<char>, ' +
-              'typename ContainerAllocator::template rebind<char>::other>',
-    'wstring': 'std::basic_string<char16_t, std::char_traits<char16_t>, ' +
-               'typename ContainerAllocator::template rebind<char16_t>::other>',
+              'typename std::allocator_traits<ContainerAllocator>::template rebind_alloc<char>>',
+    'wstring': 'std::basic_string<char16_t, std::char_traits<char16_t>, typename ' +
+               'std::allocator_traits<ContainerAllocator>::template rebind_alloc<char16_t>>',
 }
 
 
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13442/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8353/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11161/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13496/)

wjwwood (approved):
lgtm

On a related, but not blocking note, Dirk and I realized sometime last year that using custom allocators with our rclcpp API just doesn't work at all. It may still be useful to use this allocator with the message without using it with rclcpp, but that seems like a small likelyhood. One of the possible solutions was to switch to the polymorphic allocator. Having C++17 now opens that possibility up to us as it was added in C++17 and modified somewhat heavily in C++20:

https://en.cppreference.com/w/cpp/memory/polymorphic_allocator

But I still think it would be the right thing to use even now without C++20.

clalancette:
> One of the possible solutions was to switch to the polymorphic allocator. Having C++17 now opens that possibility up to us as it was added in C++17 and modified somewhat heavily in C++20:
> 
> https://en.cppreference.com/w/cpp/memory/polymorphic_allocator
> 
> But I still think it would be the right thing to use even now without C++20.

Makes sense.  Is there an issue tracking that problem?  If not, would you mind opening one (just because I don't have the context here).

The failing tests in CI are failing in the nightlies as well, so those are nothing new.  I'm going to go ahead and merge this, thanks for the review.

wjwwood:
> Makes sense. Is there an issue tracking that problem? If not, would you mind opening one (just because I don't have the context here).

Good question, I'll look.

wjwwood:
I've opened an issue here: https://github.com/ros2/rosidl/issues/566

ros-discourse:

https://discourse.ros.org/t/ros2-compatibility-with-c-20/19014/3

Issue #306, opened 2018-10-23
Investigate using rcutils allocators for the C generators


clalancette:
Based on a comment at https://github.com/ros2/rosidl/pull/300#issuecomment-432038951, we should see if we can/should use the rcutils allocators when generating the C code for messages/services.  Then we could potentially create a custom allocator to test that messages are being properly freed on error paths.

allenh1:
@clalancette if it's alright with you, I'd be more than happy to look into this.

clalancette:
@allenh1 We should probably coordinate with @wjwwood a bit on this, as he has "adding allocators" as part of the larger memory management list here: https://github.com/ros2/ros2/issues/590 .  @wjwwood Do you have a plan to start working on that, or should we let @allenh1 loose on it?

wjwwood:
I’m hesitant to turn away help but I have already started working on this in part. Specifically I’ve been adding allocator calls to the typesupport functions responsible for converting ros types to dds types. Which would in turn pass that allocator to the rosidl functions described here. Let me see if there’s a reasonable place to do separation of labor in this and get back to you.

allenh1:
@wjwwood of course! Thank you very much. Keep me posted!

y-okumura-isp:
I'm interested in ROS2 realtype behavior, and checking malloc call(see Memory management in [Introduction to Real-time Systems](https://design.ros2.org/articles/realtime_background.html)).
I found some malloc calls such as `rosidl_generator_c__String__init` in `rosidl/rosidl_generator_c/src/string_functions.c`.

This is called at RCLCPP_INFO log macro, so called in run-time(ex. `demo_nodes_cpp/src/topics/talker.cpp`).
And this is also called at `convert_dds_to_ros` of `rosidl_typesupport_{DDS}`, so I wonder this may affect subscription.

@wjwwood How is the situation on this issue?
If the problem is unsolved, I'm thinking of writing a small patch.
(As rcl package uses `static rcl_allocator_t`, I plan to convey rcl allocator to this layer and replace malloc() with allocate() of rcl allocator.)

wjwwood:
> If the problem is unsolved, I'm thinking of writing a small patch.
(As rcl package uses static rcl_allocator_t, I plan to convey rcl allocator to this layer and replace malloc() with allocate() of rcl allocator.)

[...]

PR #584, opened 2021-04-26
Use RCUtils allocators in rosidl_generator_c

pablogs9:
This PR updates the usage of dynamic memory in `rosidl_generator_c` to rely on `rcutils` allocators.

As discussed in https://github.com/ros2/rosidl/issues/306 that had a not priority because `rosidl_generator_c` was mainly used in `rclpy` and some `rcl` tests.

But now **micro-ROS** relies completely on the C99 infrastructure of ROS 2 and, for us, in some scenarios is not feasible to use the common dynamic memory API (`malloc`, `free`,...).

### DIFF
--- a/rosidl_generator_c/CMakeLists.txt
+++ b/rosidl_generator_c/CMakeLists.txt
@@ -13,8 +13,10 @@ endif()
 
 find_package(ament_cmake_python REQUIRED)
 find_package(ament_cmake_ros REQUIRED)
+find_package(rcutils REQUIRED)
 
 ament_export_dependencies(rosidl_cmake)
+ament_export_dependencies(rcutils)
 ament_export_dependencies(rosidl_typesupport_interface)
 ament_index_register_resource("rosidl_generator_packages")
 
--- a/rosidl_generator_c/CMakeLists.txt
+++ b/rosidl_generator_c/CMakeLists.txt
@@ -13,10 +13,8 @@ endif()
 
 find_package(ament_cmake_python REQUIRED)
 find_package(ament_cmake_ros REQUIRED)
-find_package(rcutils REQUIRED)
 
 ament_export_dependencies(rosidl_cmake)
-ament_export_dependencies(rcutils)
 ament_export_dependencies(rosidl_typesupport_interface)
 ament_index_register_resource("rosidl_generator_packages")
 
--- a/rosidl_generator_c/resource/msg__functions.c.em
+++ b/rosidl_generator_c/resource/msg__functions.c.em
@@ -52,6 +52,8 @@ for member in message.structure.members:
 @#>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 @
 @#<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
+#include <rcutils/allocator.h>
+
 @[if includes]@
 
 // Include directives for member types
@@ -69,8 +71,6 @@ for member in message.structure.members:
 @[    end for]@
 @[end if]@
 @#>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
-
-#include <rcutils/allocator.h>
 
 @#######################################################################
 @# message functions
--- a/rosidl_generator_c/resource/msg__functions.c.em
+++ b/rosidl_generator_c/resource/msg__functions.c.em
@@ -70,6 +70,8 @@ for member in message.structure.members:
 @[end if]@
 @#>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 
+#include <rcutils/allocator.h>
+
 @#######################################################################
 @# message functions
 @#######################################################################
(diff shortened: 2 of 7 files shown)


pablogs9:
Please @clalancette @sloretz let us know your opinions. 
And if everything is ok, is there any possibility of backporting this to Foxy?

CC: @Acuadros95

rosidl_generator_c/CMakeLists.txt
```diff
@@ -13,8 +13,10 @@ endif()
 
 find_package(ament_cmake_python REQUIRED)
 find_package(ament_cmake_ros REQUIRED)
+find_package(rcutils REQUIRED)
```
sloretz, line 16: I don't think this needs to be found here. Instead I would `find_package()` it inside of `rosidl_generator_c_generate_interfaces.cmake` since that file creates the target that needs `rcutils` to build.

rosidl_generator_c/CMakeLists.txt
```diff
@@ -13,8 +13,10 @@ endif()
...
 
 ament_export_dependencies(rosidl_cmake)
+ament_export_dependencies(rcutils)
```
sloretz, line 19: Since this is adding a dependency to the generated code, I don't think `rcutils` needs to be exported. Instead it needs to be added as a dependency on the generated target.

https://github.com/ros2/rosidl/blob/dceaeb91f72a72e3867557690e2cc074e349230c/rosidl_generator_c/cmake/rosidl_generator_c_generate_interfaces.cmake#L131

rosidl_generator_c/resource/msg__functions.c.em
```diff
@@ -235,14 +237,15 @@ for line in lines:
 @(message_typename) *
 @(message_typename)__create()
 {
-  @(message_typename) * msg = (@(message_typename) *)malloc(sizeof(@(message_typename)));
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
```
sloretz, line 240: > in some scenarios is not feasible to use the common dynamic memory API (malloc, free,...).

I think the default allocator uses `malloc()` and `free()`, and this appears to give no way to change which allocator is used. The `__create()` function would need to accept an allocator as an argument for that to be changed.

I would also recommend adding an `rcutils_allocator_t` to the message struct so the `__destroy()` function is guaranteed to use the same allocator.

>  is there any possibility of backporting this to Foxy?

If we modify the function signature then this can't be backported because it breaks ABI. If hard coding the default allocator is good enough for your use case (maybe it's enough to be able to `LD_PRELOAD` a replacement for `rcutils_get_default_allocator()`?) then it seems like that could be backported.

sloretz (changes requested):
I think the `package.xml` needs a `<build_export_depend>rcutils</` so downstream projects get `rcutils` when they build the generated target and a `<exec_depend>rcutils</` so downstream projects have `librcutils.so` around when their targets run.

rosidl_generator_c/resource/msg__functions.c.em
pablogs9: Regarding the allocation functions that are retrieved with `rcutils_get_default_allocator` that's true that modifying the default allocators can be tricky.
But by now, in micro-ROS, we use a slightly modified version of the `rcutils` (stunned for embedded purposes) that has [`rcutils_set_default_allocator`](https://github.com/micro-ROS/rcutils/blob/1fdfd3cd2b645ce683fe3e150feeac7c4a1bf073/src/allocator.c#L87) functionality. We re planning to PR some of these changes to the `rcutils` mainline.

So, in that sense, in our opinion its ok to retrieve a global-wise allocator with `rcutils_get_default_allocator` and ABI/API is not modified so we can backport.

rosidl_generator_c/CMakeLists.txt
pablogs9: Addressed

rosidl_generator_c/CMakeLists.txt
pablogs9: Addressed

pablogs9:
@sloretz I also have added f51e1b7be0d8d8f8a1f95aae6cd885d55653526f with `rosidl_runtime_c` modifications

pablogs9:
Friendly ping @sloretz

rosidl_generator_c/resource/msg__functions.c.em
sloretz: Ah, that makes sense. Mind opening an issue to change these APIs to accept a non-default allocator in the future?

sloretz (approved):
LGTM with green CI

sloretz:
CI (build: `--packages-above-and-dependencies rosidl_generator_c rosidl_runtime_c` test: `--packages-above rosidl_generator_c rosidl_runtime_c`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14549/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9306/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12214/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14671/)

pablogs9:
Are these `unstable` labels a problem? What can I do for solving them?

If they are not, as soon as you merge I'm going to prepare a backport to `foxy`.

Thanks a lot for your reviews and time @sloretz.

sloretz:
@mergifyio backport galactic foxy

sloretz:
Thanks for the contribution @pablogs9 !

clalancette:
@sloretz @pablogs9 This PR added a regression on all CI jobs; one example is https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_debug/1594/testReport/junit/(root)/projectroot/cpplint_rosidl_generated_c/ .

The problem I think is that we re-use the `msg__functions.c.em` file twice when we build services.  And when we do that, this PR now unconditionally includes `rcutils/allocator.h` twice, which cpplint doesn't like.  I *think* have a solution in #590; please take a look.

sloretz:
>@sloretz @pablogs9 This PR added a regression on all CI jobs; one example is https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_debug/1594/testReport/junit/(root)/projectroot/cpplint_rosidl_generated_c/ .

Ah shoot. It looks like I put the repos file in the`branch` instead of the `repos_url` field when I ran CI, so this change didn't actually get tested before I merged it :no_mouth: 

```
10:51:24 II> Attempting to switch all repositories to the 'https://gist.githubusercontent.com/sloretz/f52090452ceb0a50e6fc954c8b896221/raw/15f9723dbeb8c2938f7d5be23589ac56566517b1/ros2.repos' branch
```

rosidl_generator_c/resource/msg__functions.c.em
jacobperron: I'm interested in being able to pass a non-default allocator. I think we can use this existing issue to track progress: https://github.com/ros2/rosidl/issues/306

PR #590, opened 2021-05-06
Fix a cpplint allocator regression.

clalancette:
We can't unconditionally include rcutils/allocator.h in the
msg__functions.c.em, because that file is used multiple times
when generating services.  That causes cpplint to complain
about a duplicate include.  Instead put the include in
idl__functions.c.em, and mark it as already included so we
won't try to include it again.

This should fix a regression we are seeing, like: https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_debug/1594/testReport/junit/(root)/projectroot/cpplint_rosidl_generated_c/ .  I'm going to run full CI on it to be sure.

Fixes regression introduced in #584

### DIFF
--- a/rosidl_generator_c/resource/idl__functions.c.em
+++ b/rosidl_generator_c/resource/idl__functions.c.em
@@ -16,7 +16,7 @@ include_parts = [package_name] + list(interface_path.parents[0].parts) + [
     'detail', convert_camel_case_to_lower_case_underscore(interface_path.stem)]
 include_base = '/'.join(include_parts)
 
-include_directives = {include_base + '__functions.h'}
+include_directives = {include_base + '__functions.h', 'rcutils/allocator.h'}
 }@
 #include "@(include_base)__functions.h"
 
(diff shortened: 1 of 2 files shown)


clalancette:
Full CI on all platforms:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14573/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9329/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12236/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14699/)

sloretz:
Linux regression is also in the nightly `test_action_communication__rmw_cyclonedds_cpp` , so not caused by this PR :tada: https://ci.ros2.org/view/nightly/job/nightly_linux_debug/1921/testReport/junit/(root)/projectroot/test_action_communication__rmw_cyclonedds_cpp/

sloretz (approved):

clalancette:
The failures on macOS are a known problem on lore, so again not related to this PR.

PR #591, opened 2021-05-11
Fix a cpplint allocator regression. (Backport #590)

pablogs9:


sloretz:
Ohhhh this targets the mergify branch. Oops, that would have been a good way to do it. Closing in favor of #604

PR #592, opened 2021-05-11
Fix a cpplint allocator regression. (Backport #590)

pablogs9:


sloretz:
Closing in favor of #603

PR #593, opened 2021-05-11
Generate documentation in generated C header files based on ROS interfaces comments

ivanpauno:
This PR does a bit more of what the title says, as I ended up fixing some old bugs.
I can create individual PRs for those if desired, but I'm creating only one now so there's more context.

* Generate documentation based on comments in ROS interface files.
  * This followed the current "parsing rules" in rosidl_adapter:
    ```
    # The top level comment is associated with the structure of the message.
    # The top level comment ends after the first new line.
    #
    # This is still part of the top level comment.
    
    # This comment is documentation for field1.

    # This is also documentation for field1.
    float field1;  # This is documentation for field1 as well.
    ```
    We might want to review those rules.
  * Done in commits: https://github.com/ros2/rosidl/commit/faa7c03b7ee291d3eecef4674d870856ccc7e67b, https://github.com/ros2/rosidl/commit/9b9d091f7746eaf3e4b5bada1e2bfca76cc9f3ea, https://github.com/ros2/rosidl/commit/ee86652d617940a57f47e7c1d440b3cb398dcbaf.
* Improve comment parsing in ROS interfaces:
  * Ignore multiple `#`.
  * Dedent comments.
  * Done in: https://github.com/ros2/rosidl/commit/3138141d36e0734dc50619c7cbe9714041c6c8b7, https://github.com/ros2/rosidl/commit/fd6de7a55a4152119c5febe2df354c47b1ca03da, https://github.com/ros2/rosidl/commit/b8360eeca11266ed569d8395748b79b860ac63c2.
* We weren't parsing escape sequences correctly in .idl files string literals.
  * Fixed in commit: https://github.com/ros2/rosidl/commit/7eb60df1308423381913ba05ea7481e9596299f6.
* rosidl_adapter was generating array default values that couldn't be correctly processed (the previous bug was hiding that).
  * The fact that we're using [ast.literal_eval](https://github.com/ros2/rosidl/blob/608761cd1bcf6f01e4d25f8fcc9745f1ea41d0b0/rosidl_generator_c/resource/msg__functions.c.em#L3) is extremely strange. I'm not sure what's the standard IDL way of specifying arrays defaults, it doesn't seem to be specified AFAIS.
  * Repeating that code in all generators also seems wrong, [Member](https://github.com/ros2/rosidl/blob/608761cd1bcf6f01e4d25f8fcc9745f1ea41d0b0/rosidl_parser/rosidl_parser/definition.py#L460) could use a `get_default_value`/`has_default_value` methods.
    I haven't addressed that here.
  * Fixed in https://github.com/ros2/rosidl/commit/daf38331f64b68001c392a0eb0c45b94b1864edd.
* We weren't passing ROS interface comments in constants to IDL, and also our IDL parser didn't support annotations in constants. Fixed in https://github.com/ros2/rosidl/commit/9bbd1ec3e384238a8fcb5cf8c053fdb363ab4d16.

### DIFF
--- a/rosidl_generator_c/resource/msg__struct.h.em
+++ b/rosidl_generator_c/resource/msg__struct.h.em
@@ -60,7 +60,11 @@ for member in message.structure.members:
 @[if comments]@
 /**
 @[  for line in comments]@
+@[    if line]@
   * @(line)
+@[    else]@
+  *
+@[    end if]@
 @[  end for]@
  */
 @[end if]@
@@ -142,7 +146,11 @@ enum
 @[if comments]@
 /**
 @[  for line in comments]@
+@[    if line]@
   * @(line)
+@[    else]@
+  *
+@[    end if]@
 @[  end for]@
  */
 @[end if]@
@@ -150,7 +158,11 @@ typedef struct @(idl_structure_type_to_c_typename(message.structure.namespaced_t
 {
 @[for member in message.structure.members]@
 @[  for line in member.get_comment_lines()]@
+@[    if line]@
   /// @(line)
+@[    else]@
+  ///
+@[    end if]@
 @[  end for]@
   @(idl_declaration_to_c(member.type, member.name));
 @[end for]@
--- a/rosidl_generator_c/resource/msg__struct.h.em
+++ b/rosidl_generator_c/resource/msg__struct.h.em
@@ -56,12 +56,11 @@ for member in message.structure.members:
 @[for constant in message.constants]@
 
 /// Constant '@(constant.name)'.
-@[if constant.annotations]@
+@{comments = constant.get_comment_lines()}@
+@[if comments]@
 /**
-@[  for x in constant.get_annotation_values('verbatim')]@
-@[    if x['language'] == 'comment']@
-  * @(x['text'])
-@[    end if]@
+@[  for line in comments]@
+  * @(line)
 @[  end for]@
  */
 @[end if]@
@@ -139,22 +138,19 @@ enum
 
 @#<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
 /// Struct defined in @(interface_path_to_string(interface_path)) in the package @(package_name).
-@[if message.structure.annotations]@
+@{comments = message.structure.get_comment_lines()}@
+@[if comments]@
 /**
-@[  for x in message.structure.get_annotation_values('verbatim')]@
-@[    if x['language'] == 'comment']@
-  * @(x['text'])
-@[    end if]@
+@[  for line in comments]@
+  * @(line)
 @[  end for]@
  */
 @[end if]@
 typedef struct @(idl_structure_type_to_c_typename(message.structure.namespaced_type))
 {
 @[for member in message.structure.members]@
-@[  for x in member.get_annotation_values('verbatim')]@
-@[    if x['language'] == 'comment']@
-  /// @(x['text'])
-@[    end if]@
+@[  for line in member.get_comment_lines()]@
+  /// @(line)
 @[  end for]@
   @(idl_declaration_to_c(member.type, member.name));
 @[end for]@
--- a/rosidl_generator_c/resource/msg__struct.h.em
+++ b/rosidl_generator_c/resource/msg__struct.h.em
@@ -56,6 +56,15 @@ for member in message.structure.members:
 @[for constant in message.constants]@
 
 /// Constant '@(constant.name)'.
+@[if constant.annotations]@
+/**
+@[  for x in constant.get_annotation_values('verbatim')]@
+@[    if x['language'] == 'comment']@
+  * @(x['text'])
+@[    end if]@
+@[  end for]@
+ */
+@[end if]@
 @[    if isinstance(constant.type, BasicType)]@
 @[        if constant.type.typename in (
                 *INTEGER_TYPES, *CHARACTER_TYPES, OCTET_TYPE
@@ -129,10 +138,24 @@ enum
 @#>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 
 @#<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
-// Struct defined in @(interface_path_to_string(interface_path)) in the package @(package_name).
+/// Struct defined in @(interface_path_to_string(interface_path)) in the package @(package_name).
+@[if message.structure.annotations]@
+/**
+@[  for x in message.structure.get_annotation_values('verbatim')]@
+@[    if x['language'] == 'comment']@
+  * @(x['text'])
+@[    end if]@
+@[  end for]@
+ */
+@[end if]@
 typedef struct @(idl_structure_type_to_c_typename(message.structure.namespaced_type))
 {
 @[for member in message.structure.members]@
+@[  for x in member.get_annotation_values('verbatim')]@
+@[    if x['language'] == 'comment']@
+  /// @(x['text'])
+@[    end if]@
+@[  end for]@
   @(idl_declaration_to_c(member.type, member.name));
 @[end for]@
 } @(idl_structure_type_to_c_typename(message.structure.namespaced_type));
--- a/rosidl_parser/rosidl_parser/definition.py
+++ b/rosidl_parser/rosidl_parser/definition.py
@@ -436,6 +436,22 @@ class Annotatable:
         """
         return [a.value for a in self.annotations if a.name == name]
 
+    def get_comment_lines(self):
+        """
+        Get the comment lines of the annotatable.
+
+        :param str name: the name of the annotation type
+        :returns: a list of comment lines
+        """
+        comments = [
+            x['text'] for x in self.get_annotation_values('verbatim') if
+            'language' in x and 'text' in x and x['language'] == 'comment'
+        ]
+        lines = []
+        for comment in comments:
+            lines.extend(comment.split('\n'))
+        return lines
+
     def has_annotation(self, name):
         """
         Check if there is exactly one annotation of a specific type.
--- a/rosidl_parser/rosidl_parser/definition.py
+++ b/rosidl_parser/rosidl_parser/definition.py
@@ -440,7 +440,6 @@ class Annotatable:
         """
         Get the comment lines of the annotatable.
 
-        :param str name: the name of the annotation type
         :returns: a list of comment lines
         """
         comments = [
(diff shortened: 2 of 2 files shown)


ivanpauno:
> This followed the current "parsing rules" in rosidl_adapter:

Unfortunatelly, we don't follow those rules correctly in many places.

* https://github.com/ros2/rcl_interfaces/blob/c5d0a20150533a6cb93d5dcbd6e27cea9300aa6e/rcl_interfaces/msg/Log.msg#L1-L12 this top level comment is talking about the constants, but we're associating it to the generated `struct`.
* https://github.com/ros2/rcl_interfaces/blob/c5d0a20150533a6cb93d5dcbd6e27cea9300aa6e/rcl_interfaces/msg/Log.msg#L31-L33 this comment is associated to the next field, though it shouldn't be associated to anything.
* For actions and srvs it's even more strange, I wouldn't expect [this](https://github.com/ros2/rosidl/blob/ivanpauno/pass-comments-to-generated-c-code/rosidl_adapter/test/data/action/Test.expected.idl) from [this](https://github.com/ros2/rosidl/blob/ivanpauno/pass-comments-to-generated-c-code/rosidl_adapter/test/data/action/Test.action).

We should probably need to define some rules of what we expect (which could be the current rules), and correct our interface definitions so the comments that are passed through look like we want.

rosidl_parser/rosidl_parser/parser.py
```diff
@@ -103,16 +103,19 @@ def extract_content_from_ast(tree):
     constants = {}
     const_dcls = tree.find_data('const_dcl')
     for const_dcl in const_dcls:
+        annotations = get_annotations(const_dcl)
         const_type = next(const_dcl.find_data('const_type'))
-        const_expr = next(const_dcl.find_data('const_expr'))
+        *_, const_expr_value = const_dcl.find_data('const_expr')
```
hidmic, line 108: @ivanpauno nit: isn't this still a `const_expr`:
```suggestion
        *_, const_expr = const_dcl.find_data('const_expr')
```
?

rosidl_adapter/rosidl_adapter/msg/__init__.py
```diff
@@ -61,10 +61,8 @@ def convert_msg_to_idl(package_dir, package_name, input_file, output_dir):
 
 def to_idl_literal(idl_type, value):
     if idl_type[-1] == ']' or idl_type.startswith('sequence<'):
-        elements = [repr(v) for v in value]
-        while len(elements) < 2:
-            elements.append('')
```
hidmic, line 66: @ivanpauno just to double check, is it guaranteed that `elements` will never be empty? I'd be curious about the original code if not though.

rosidl_adapter/test/data/action/Test.expected.idl
```diff
@@ -6,7 +6,7 @@
 module test_msgs {
   module action {
     @verbatim (language="comment", text=
-      " goal definition")
+      "goal definition")
```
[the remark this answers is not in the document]
ivanpauno: To be more clear, what I think it's strange is that this comment is associated with the goal, though the comments that I would expect to be associated with `result` and `feedback` are actually associated with their fields (`ok` and `sequence`).

If you agree with that, I will also address it (either here or in a new PR, this is already doing way too much)

rosidl_parser/rosidl_parser/parser.py
ivanpauno: Yes, it's the `const_expr` representing the value of the constant.
The annotations can also have some `const_expr` in the middle, so this was the way to get the one corresponding to the value.

rosidl_adapter/rosidl_adapter/msg/__init__.py
ivanpauno: > just to double check, is it guaranteed that elements will never be empty?

Maybe value can be `[]`, I think that can be an acceptable default value for an array.
IDK what that `while` is doing though, it doesn't seem to make sense.

rosidl_parser/rosidl_parser/parser.py
```diff
@@ -103,16 +103,19 @@ def extract_content_from_ast(tree):
...
         module_identifiers = get_module_identifier_values(tree, const_dcl)
         module_comments = constants.setdefault(
             module_identifiers[-1], [])
-        value = get_const_expr_value(const_expr)
-        module_comments.append(Constant(
-            get_first_identifier_value(const_dcl),
+        value = get_const_expr_value(const_expr_value)
+        constant = Constant(
+            get_child_identifier_value(const_dcl),
```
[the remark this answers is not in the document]
ivanpauno: I saw that worked for a `struct` and copy-pasted it here as well :joy:

https://github.com/ros2/rosidl/blob/ee86652d617940a57f47e7c1d440b3cb398dcbaf/rosidl_parser/rosidl_parser/parser.py#L144

But yes, it works because it only checks direct children and doesn't check recursively.

rosidl_parser/rosidl_parser/parser.py
ivanpauno: Actually, I think that I should only be checking childs like [here](https://github.com/ros2/rosidl/pull/593#discussion_r630455973).
@hidmic what do you think?

rosidl_adapter/test/data/action/Test.expected.idl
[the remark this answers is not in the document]
hidmic: > To be more clear, what I think it's strange is that this comment is associated with the goal, though the comments that I would expect to be associated with result and feedback are actually associated with their fields (ok and sequence).

Agreed.

> If you agree with that, I will also address it (either here or **in a new PR, this is already doing way too much**)

Agreed.

rosidl_adapter/rosidl_adapter/msg/__init__.py
hidmic: > IDK what that while is doing though, it doesn't seem to make sense.

IIUC it was ensuring you get `(value,)` if `len(elements) == 1` but you'd get `(,)` if `len(elements) == 0`, which is nonsense. 

> Maybe value can be [], I think that can be an acceptable default value for an array.

Maybe? Anyhow, your code will result in `()`, which is a bit more sensible if allowed.

rosidl_adapter/rosidl_adapter/parser.py
```diff
@@ -580,6 +581,9 @@ def process_comments(instance):
                 length -= 1
                 continue
             i += 1
+        if lines:
+            text = '\n'.join(lines)
+            instance.annotations['comment'] = textwrap.dedent(text).split('\n')
```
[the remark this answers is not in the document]
ivanpauno: That doesn't work.
It's getting `comment = '\n'.join(lines)`, bun then it's mutating lines and not doing `lines = comment.split('\n')`.

Some parts of the pre-existing code are really weird, e.g. what is it doing with units?
I didn't want to touch that awesome code :smiley:

rosidl_parser/rosidl_parser/parser.py
ivanpauno: > What you're getting is, strictly speaking, a const_expr object / thingie. That's why I pointed it out. It's just nit though,

Ah I see, I didn't notice that `const_expr_value` was also a thing.

> I don't have a strong opinion. Checking direct children is probably cheaper than a full scan though.

Simplified that in https://github.com/ros2/rosidl/pull/593/commits/16d079203c5f9c7d2877cb5b914407129dff4fde.

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -436,6 +436,22 @@ def get_annotation_values(self, name):
         """
         return [a.value for a in self.annotations if a.name == name]
 
+    def get_comment_lines(self):
+        """
+        Get the comment lines of the annotatable.
+
+        :param str name: the name of the annotation type
```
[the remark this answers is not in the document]
ivanpauno: Fixed in https://github.com/ros2/rosidl/pull/593/commits/f52e97dca2313abf9ef167f4dde676833525e483

ros-discourse:

https://discourse.ros.org/t/structured-metadata-for-message-files/20965/1

rosidl_generator_c/resource/msg__struct.h.em
```diff
@@ -129,10 +141,29 @@ enum
...
+ */
+@[end if]@
 typedef struct @(idl_structure_type_to_c_typename(message.structure.namespaced_type))
 {
 @[for member in message.structure.members]@
+@[  for line in member.get_comment_lines()]@
+@[    if line]@
+  /// @(line)
+@[    else]@
+  ///
```
jacobperron, line 164: IIUC, this is inserting an empty comment. Why do this versus inserting nothing?

If this is to satisfy linters, we could alternatively be explicit about there being no documentation.
E.g. `/// Not documented` or some identifier that in theory someone could use to detect something is not documented.

jacobperron: Same for the struct comment above.

jacobperron (approved):
LGTM with CI

One comment below that does not necessarily have to be addressed.

rosidl_generator_c/resource/msg__struct.h.em
ivanpauno: The idea is to respect the new lines in the original comment.
If I delete the `if/else` and just use `/// @(line)`, empty lines will generate `/// ` (note the space at the end), and linters will reject the generated header file.
For empty lines I want to leave an empty line in the comments, which equivalent is `///`.

We could also ommit new lines, but I think it's better to maintain them.

jacobperron: Thanks for explaining. That makes sense to me :+1:

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -436,6 +436,21 @@ def get_annotation_values(self, name):
...
+
+        :returns: a list of comment lines
+        """
+        comments = [
+            x['text'] for x in self.get_annotation_values('verbatim') if
+            'language' in x and 'text' in x and x['language'] == 'comment'
+        ]
+        lines = []
+        for comment in comments:
+            lines.extend(comment.split('\n'))
```
hidmic, line 451: @ivanpauno nit^N: 
```suggestion
            lines.extend(comment.splitlines())
```

hidmic (approved):

jacobperron:
@ivanpauno Friendly ping. This one has approvals; just needs CI I think.

rosidl_parser/rosidl_parser/definition.py
ivanpauno: Done in https://github.com/ros2/rosidl/pull/593/commits/1b09d4d546ef3b5f7c225a4a3eb16b436741ebe4

ivanpauno:
Rebased after https://github.com/ros2/rosidl/pull/630 and https://github.com/ros2/rosidl/pull/632, I think now we're all set

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15944/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10617/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16266/)

ivanpauno:
@jacobperron @hidmic only a last approval is pending to get this merged, could you take a look? Thanks!

hidmic (approved):

PR #596, opened 2021-05-21
Update function prefix

pablogs9:
I have found that this template will generate objects with same name if two packages have a .msg named equal.

This generate symbols named equal and in some cases, the linker will complain about this duplication.

In order to solve it, I propose to append also the package name that is guaranteed to be unique.

### DIFF
--- a/rosidl_typesupport_introspection_c/resource/msg__type_support.c.em
+++ b/rosidl_typesupport_introspection_c/resource/msg__type_support.c.em
@@ -27,7 +27,7 @@ header_files = [
     include_base + '__struct.h',
 ]
 
-function_prefix = message.structure.namespaced_type.name + '__rosidl_typesupport_introspection_c'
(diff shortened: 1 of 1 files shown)


clalancette (commented):
This idea seems totally reasonable to me.  My only concern here is that this may trigger cpplint warnings since it adds a lot of text to function names.  I guess we'll see when we run full CI, which I'll do next.

I would like to get another opinion from @sloretz as well.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14676/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9428/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12339/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14819/)

pablogs9:
Hello, @clalancette any update on this?

pablogs9:
@mergify backport galactic

pablogs9:
Thanks a lot, @clalancette. If you don't mind I have backported it to galactic.

PR #597, opened 2021-06-04
Support passing keyword arguments to rosidl CLI extensions

hidmic:
This patch builds on top of #567, adding support for passing keyword arguments for CLI extensions along with their spec. This allow for extension specific configuration.

CI up `rosidl_cli`, `rosidl_typesupport_c`, `rosidl_typesupport_cpp`, and `rosidl_generator_py`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14724/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9476/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12391/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14870/)

### DIFF
--- a/rosidl_cli/rosidl_cli/extensions.py
+++ b/rosidl_cli/rosidl_cli/extensions.py
@@ -26,7 +26,7 @@ logger = logging.getLogger(__name__)
 class Extension:
     """A generic extension point."""
 
-    def __init__(self, name, **kwargs):
+    def __init__(self, name):
         self.__name = name
 
     @property
--- a/rosidl_cli/rosidl_cli/extensions.py
+++ b/rosidl_cli/rosidl_cli/extensions.py
@@ -13,6 +13,9 @@
 # limitations under the License.
 
 import logging
+import re
+
+import yaml
 
 from rosidl_cli.entry_points import load_entry_points
 
@@ -23,7 +26,7 @@ logger = logging.getLogger(__name__)
 class Extension:
     """A generic extension point."""
 
-    def __init__(self, name):
+    def __init__(self, name, **kwargs):
         self.__name = name
 
     @property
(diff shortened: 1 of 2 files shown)


hidmic:
@clalancette @sloretz friendly ping

clalancette (commented):
I don't think I have enough context to review this.  Two questions that might help clarify things a bit:

1. How is this intended to be used in the end?  That is, what is a realistic example of something that might get passed through here?
2. Do we really have to define our own new string format for the SPECS_PATTERN?  Can't we use straight YAML (or XML, or JSON, or....)?

hidmic:
> How is this intended to be used in the end? That is, what is a realistic example of something that might get passed through here?

It really depends on each generator implementation. A realistic example, the one that pushed me to do this, is narrowing down the pool of typesupport implementations to be used when generating CPython extensions and C/C++ typesupport trampolines. Simply looking up the `ament_index` assumes the corresponding code has been generated and built, which is true in the typical `rosidl` CMake pipeline (where presence in the index implies ament extension registration) but not when manually recreating a subset of it in a different build system :)

> Do we really have to define our own new string format for the SPECS_PATTERN? Can't we use straight YAML (or XML, or JSON, or....)?

It is mostly YAML. We could go from `py[key: value]` to `{name: py, kwargs: {}}` plus `py` being a shortcut to `{name: py}`. I don't have a strong opinion.

hidmic:
@clalancette @sloretz friendly ping

rosidl_cli/rosidl_cli/extensions.py
```diff
@@ -13,44 +13,86 @@
...
+
 
 logger = logging.getLogger(__name__)
 
 
 class Extension:
     """A generic extension point."""
 
-    def __init__(self, name):
+    def __init__(self, name, **kwargs):
```
sloretz, line 29: I think the base class shouldn't have `**kwargs`. I think this in combination with how the PR in rosidl_python is passing remaining args means bogus arguments would be silently ignored. That could be confusing in case of typo's or keyword arguments being deprecated.

```
# The arguments here would be silently ignored
py[typosupport_implementations:[rosidl_typesupport_c]]
```

sloretz (changes requested):

rosidl_cli/rosidl_cli/extensions.py
hidmic: That's a very good point. See https://github.com/ros2/rosidl/pull/597/commits/db920d1e98bcfd899ea0d5a96845f4d03ae275e4.

hidmic:
Re-running CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14865/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9615/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12532/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15024/)

sloretz (approved):

hidmic:
Alright, thanks for the reviews! I'll get this in and release so we can have the Rolling PR job passing for downstream PRs.

PR #601, opened 2021-06-24
Bundle and ensure the exportation of rosidl generated targets 

hidmic:
Precisely what the title says. The `rosidl` pipeline was not exporting libraries and their transitive dependencies correctly, and neither was ensuring these libraries are later discovered in topological order by downstream consumers (and thus the need to bundle in a single export set).


rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
```diff
@@ -220,6 +220,7 @@ macro(rosidl_generate_interfaces target)
   )
 
   if(NOT _ARG_SKIP_INSTALL)
+    _ament_cmake_export_targets_register_package_hook()
```
ivanpauno, line 223: why do you call this manually?
That seems to be pretty odd.

rosidl_generator_c/cmake/rosidl_generator_c_generate_interfaces.cmake
```diff
@@ -152,10 +152,9 @@ if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
     )
   endif()
   ament_export_libraries(${rosidl_generate_interfaces_TARGET}${_target_suffix})
-  ament_export_targets(${rosidl_generate_interfaces_TARGET}${_target_suffix})
   install(
     TARGETS ${rosidl_generate_interfaces_TARGET}${_target_suffix}
-    EXPORT ${rosidl_generate_interfaces_TARGET}${_target_suffix}
+    EXPORT ${rosidl_generate_interfaces_TARGET}
```
ivanpauno, line 157: why the suffix isn't needed anymore?

ivanpauno: what this is doing is to only export "std_msgs" and not "std_msgs__rosidl_generator_c" (insert correct suffix there) , right? was the second target being used externally?
I guess we're repeating the export of `${rosidl_generate_interfaces_TARGET}` in all the extensions, I'm not sure if that's ok or not.

ivanpauno (commented):
> Precisely what the title says. The rosidl pipeline was not exporting libraries and their transitive dependencies correctly, and neither was ensuring these libraries are later discovered in topological order by downstream consumers (and thus the need to bundle in a single export set).

Can you explain what this patch is doing and how does this patch fix the issue?
It isn't obvious to me

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
hidmic: It is odd, and in retrospect I should've added a big comment right next to it.

In `ament_cmake`, hook macros conventionally register `ament_package` extensions. That's the case for [`_ament_cmake_export_targets_register_package_hook()`](https://github.com/ament/ament_cmake/blob/84719051ef2b26070de484e60b8f060ae7a70b06/ament_cmake_export_targets/ament_cmake_export_targets-extras.cmake#L19-L27). Typically, this extension is what actually performs the intended action e.g. `ament_export_targets()` collects them, but it is the [extension](https://github.com/ament/ament_cmake/blob/84719051ef2b26070de484e60b8f060ae7a70b06/ament_cmake_export_targets/cmake/ament_cmake_export_targets_package_hook.cmake) that actually installs the export set.

One normally does not invoke them directly -- [the user-facing macro does](https://github.com/ament/ament_cmake/blob/84719051ef2b26070de484e60b8f060ae7a70b06/ament_cmake_export_targets/cmake/ament_export_targets.cmake#L38). The big issue with it is that you lose control of _when_ the extension is registered, and extensions are invoked in registration order. That order is reflected in the `${PROJECT_NAME}_CONFIG_EXTRAS` variable, which is the same as the order in which the extra CMake files are included by downstream code.

Without this change, it is perfectly possible for [`rosidl_cmake_export_typesupport_targets-extras.cmake`](https://github.com/ros2/rosidl/blob/master/rosidl_cmake/cmake/rosidl_cmake_export_typesupport_targets-extras.cmake.in) to be loaded before [`ament_export_targets-extras.cmake`](https://github.com/ament/ament_cmake/blob/84719051ef2b26070de484e60b8f060ae7a70b06/ament_cmake_export_targets/cmake/ament_cmake_export_targets-extras.cmake.in) and fail. By invoking the hook earlier, I ensure targets are imported first.

Arguably, you can achieve the same by making sure that macros are invoked in an specific order in all generators (not the case right now) but it is fragile IMHO.

rosidl_generator_c/cmake/rosidl_generator_c_generate_interfaces.cmake
hidmic: > what this is doing is to only export "std_msgs" and not "std_msgs__rosidl_generator_c", right?

Exactly.

> I guess we're repeating the export of `${rosidl_generate_interfaces_TARGET}` in all the extensions, I'm not sure if that's ok or not.

So, every generator adds targets to the `${rosidl_generate_interfaces_TARGET}` export set, but the latter is exported once in [`rosidl_generate_interfaces()`](https://github.com/ros2/rosidl/pull/601/files#diff-ff51480653554f6ee5e7359a8c662bf8a46d2b6158fa576c8e2ce9cf1bfbf24aR305). 

---

As to why this is necessary, it boils down to the same `ament` extension order problem. Before this patch, every generator would install its own export set. The order in which each export is then loaded by downstream code is that of the registration, not topological. This comment:

https://github.com/ros2/rosidl/blob/01d1c3ca0274b8ddc0692a2f6045a0926e8e3031/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake#L268-L270

suggests this is somehow guaranteed by proper extension writing, but it fell apart as soon as I applied https://github.com/ros2/rosidl/pull/601/commits/ff6e53fecff62ebf93cf3c0f46c15cc1d42df741.

By aggregating all generator targets in a single export set, CMake takes care of the topological ordering.

ivanpauno: > So, every generator adds targets to the ${rosidl_generate_interfaces_TARGET} export set, but the latter is exported once in rosidl_generate_interfaces().

I'm not sure if this matters or not, but I think you only need one

```cmake
install(EXPORT ${rosidl_generate_interfaces_TARGET})
```

that could maybe be done in `cmake/rosidl_generate_interfaces.cmake`, though if this approach works I don't think it matters much.

> was the second target being used externally?

This question is important, were we using the suffixed targets that were previously exported somewhere else.
They aren't being exported anymore.

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
ivanpauno: > Without this change, it is perfectly possible for rosidl_cmake_export_typesupport_targets-extras.cmake to be included before ament_export_targets-extras.cmake and fail to collect any targets (which is what happened to me). By invoking the hook earlier, we ensure targets are imported first.

this is unfortunate, but I don't see a simple solutions ....
Please add a big comment :smiley:

rosidl_generator_c/cmake/rosidl_generator_c_generate_interfaces.cmake
hidmic: > I'm not sure if this matters or not, but I think you only need one

Right, that's what `ament_export_targets()` does. I could roll out a reduced, vanilla CMake form of `ament_export_targets()` just for `rosidl`, but I'd figured it'd be better to stick to `ament`.

> This question is important, were we using the suffixed targets that were previously exported somewhere else.

So, the distinction between target and export set is important. We are exporting all the same targets, but through a single export set. Community contributed `rosidl` generators can still export their own sets, we just can't assure it'll be included in the right order.

hidmic: One thing that may be creating confusion is `ament_export_targets()`. It doesn't actually take targets, it takes export set names (see [here](https://github.com/ament/ament_cmake/blob/84719051ef2b26070de484e60b8f060ae7a70b06/ament_cmake_export_targets/cmake/ament_cmake_export_targets_package_hook.cmake#L29) and [here](https://github.com/ament/ament_cmake/blob/84719051ef2b26070de484e60b8f060ae7a70b06/ament_cmake_export_targets/cmake/ament_cmake_export_targets-extras.cmake.in#L8)) which  almost always bear the same name as their corresponding target in our codebase.

ivanpauno: Ah thanks, I forgot how modern cmake actually worked :smile: .

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
hidmic: See https://github.com/ros2/rosidl/pull/601/commits/a10f1c8288525ad38809300ff9c1d8f0c9eb7a13.

ivanpauno: danke

ivanpauno (approved):

hidmic:
Full CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14825/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9575/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12491/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14980/)

PR #624, opened 2021-10-29
Update package maintainers

hidmic:
Precisely what the title says.


clalancette (approved):

hidmic:
@clalancette added you back as an author.

Issue #634, opened 2021-11-22
Galactic: set CMAKE_CXX_STANDARD to 17


oysstu:
Since rclcpp targets c++17 in galactic, rosidl should probably do the same.

https://github.com/ros2/rosidl/blob/7d95d643a7260ed37af4c7803422be9a205d9e51/rosidl_typesupport_introspection_cpp/CMakeLists.txt#L7

There's also a few places where C++14 is used in tests
https://github.com/ros2/rosidl/blob/7d95d643a7260ed37af4c7803422be9a205d9e51/rosidl_typesupport_interface/CMakeLists.txt#L26

https://github.com/ros2/rosidl/blob/7d95d643a7260ed37af4c7803422be9a205d9e51/rosidl_runtime_cpp/CMakeLists.txt#L30

https://github.com/ros2/rosidl/blob/7d95d643a7260ed37af4c7803422be9a205d9e51/rosidl_generator_cpp/CMakeLists.txt#L39

https://github.com/ros2/rosidl/blob/7d95d643a7260ed37af4c7803422be9a205d9e51/rosidl_runtime_c/CMakeLists.txt#L64

aprotyas:
Do you mind submitting a quick PR to that effect?

oysstu:
Done, see #635. Can open a backport PR to Galactic afterwards

PR #635, opened 2021-11-23
Set CXX standard to 17

oysstu:
Set CXX standard to 17. Valid for Galactic and later (#634).

### DIFF
--- a/rosidl_generator_cpp/CMakeLists.txt
+++ b/rosidl_generator_cpp/CMakeLists.txt
@@ -34,9 +34,9 @@ if(BUILD_TESTING)
     SKIP_INSTALL
   )
 
-  # Default to C++14
+  # Default to C++17
   if(NOT CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 14)
+    set(CMAKE_CXX_STANDARD 17)
   endif()
 
   if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
--- a/rosidl_runtime_c/CMakeLists.txt
+++ b/rosidl_runtime_c/CMakeLists.txt
@@ -59,9 +59,9 @@ if(BUILD_TESTING)
 
   # For gtest
   enable_language(CXX)
-  # Default to C++14
+  # Default to C++17
   if(NOT CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 14)
+    set(CMAKE_CXX_STANDARD 17)
   endif()
 
   find_package(ament_cmake_gtest REQUIRED)
--- a/rosidl_runtime_cpp/CMakeLists.txt
+++ b/rosidl_runtime_cpp/CMakeLists.txt
@@ -9,6 +9,11 @@ target_include_directories(${PROJECT_NAME} INTERFACE
   "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
   "$<INSTALL_INTERFACE:include>")
 
+if (MSVC)
+  target_compile_definitions(${PROJECT_NAME} INTERFACE
+  _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING)
+endif()
+
 ament_export_include_directories(include)
 ament_export_targets(${PROJECT_NAME})
 
@@ -25,9 +30,9 @@ if(BUILD_TESTING)
 
   find_package(ament_cmake_gtest REQUIRED)
 
-  # Default to C++14
+  # Default to C++17
   if(NOT CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 14)
+    set(CMAKE_CXX_STANDARD 17)
   endif()
 
   if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
--- a/rosidl_runtime_cpp/CMakeLists.txt
+++ b/rosidl_runtime_cpp/CMakeLists.txt
@@ -9,7 +9,7 @@ target_include_directories(${PROJECT_NAME} INTERFACE
   "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
   "$<INSTALL_INTERFACE:include>")
 
-if (MSVC)
+if(MSVC)
   target_compile_definitions(${PROJECT_NAME} INTERFACE
   _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING)
 endif()
--- a/rosidl_typesupport_interface/CMakeLists.txt
+++ b/rosidl_typesupport_interface/CMakeLists.txt
@@ -21,9 +21,9 @@ if(BUILD_TESTING)
   if(NOT CMAKE_C_STANDARD)
     set(CMAKE_C_STANDARD 11)
   endif()
-  # Default to C++14
+  # Default to C++17
   if(NOT CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 14)
+    set(CMAKE_CXX_STANDARD 17)
   endif()
 
   find_package(ament_cmake_gtest REQUIRED)
--- a/rosidl_typesupport_introspection_cpp/CMakeLists.txt
+++ b/rosidl_typesupport_introspection_cpp/CMakeLists.txt
@@ -2,9 +2,9 @@ cmake_minimum_required(VERSION 3.12)
 
 project(rosidl_typesupport_introspection_cpp)
 
-# Default to C++14
+# Default to C++17
 if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD 17)
 endif()
 
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
--- a/rosidl_typesupport_introspection_cpp/cmake/rosidl_typesupport_introspection_cpp_generate_interfaces.cmake
+++ b/rosidl_typesupport_introspection_cpp/cmake/rosidl_typesupport_introspection_cpp_generate_interfaces.cmake
@@ -98,7 +98,7 @@ if(WIN32)
 endif()
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   set_target_properties(${rosidl_generate_interfaces_TARGET}${_target_suffix} PROPERTIES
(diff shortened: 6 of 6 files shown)


sloretz (approved):
I'd rather this used `target_compile_features(target_name PUBLIC cxx_std_17)` instead, but I'd treat that as an enhancement rather than blocking this PR.


LGTM with green CI

sloretz:
CI
 * repos: `https://gist.githubusercontent.com/sloretz/6afda0358f48c644a0d292aae353e449/raw/e2f14016b0815e8c89af6233752f9a1cee1b8320/ros2.repos`
 * build: ` --packages-up-to rosidl_runtime_c rosidl_runtime_c rosidl_typesupport_introspection_cpp rosidl_typesupport_interface rosidl_runtime_cpp rosidl_generator_cpp`
 * test: `--packages-select rosidl_runtime_c rosidl_runtime_c rosidl_typesupport_introspection_cpp rosidl_typesupport_interface rosidl_runtime_cpp rosidl_generator_cpp`)
 
Jobs
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15655/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10333/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13321/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15876/)

aprotyas (changes requested):
Thank you for the PR! This looks good to me on a first pass, I think you missed one instance in [`rosidl_typesupport_introspection_cpp_generate_interfaces.cmake#L101`](https://github.com/oysstu/rosidl/blob/9c24e0e13b56f176eaa1c0c774b74529b06f004a/rosidl_typesupport_introspection_cpp/cmake/rosidl_typesupport_introspection_cpp_generate_interfaces.cmake#L101)

```cmake
if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
  set_target_properties(${rosidl_generate_interfaces_TARGET}${_target_suffix} PROPERTIES
    CXX_STANDARD 14
    COMPILE_OPTIONS -Wall -Wextra -Wpedantic
  )
```

aprotyas:
The warnings in the MSVC build come from the deprecation of [`std::wstring_convert`](https://en.cppreference.com/w/cpp/locale/wstring_convert) and of the [`<codecvt>`](https://en.cppreference.com/w/cpp/header/codecvt) header in C++17.

The deprecation message isn't benign because:
```
The C++ Standard doesn't provide equivalent non-deprecated functionality; consider using MultiByteToWideChar() and WideCharToMultiByte() from instead.
```

The suggestion here is not portable, being part of the Win32 API.

---

There's a couple of ways to approach this:

- According to https://isocpp.org/files/papers/p0636r0.html, the recommendation is "Users should use dedicated text-processing libraries instead."
  - [utfcpp](https://github.com/nemtrif/utfcpp) is a good library to do so, but that would introduce an external dependency in a core package.
- Silence this specific deprecation warning with `#define _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING` and hope that said classes/header are not removed before proper Unicode support is added to C++23.

oysstu:
> I'd rather this used `target_compile_features(target_name PUBLIC cxx_std_17)` instead, but I'd treat that as an enhancement rather than blocking this PR.
> 
> LGTM with green CI

Yes, that seems to be the modern approach to setting the C++ standard. Would also require upping the minimum cmake version to 3.10 in order to have support for MSVC. This is also a change that could be applied to all c++ packages.

> Thank you for the PR! This looks good to me on a first pass, I think you missed one instance in
> ...

Nice catch, I've pushed another commit that changes this one as well.

> The warnings in the MSVC build come from the deprecation of [`std::wstring_convert`](https://en.cppreference.com/w/cpp/locale/wstring_convert) and of the [`<codecvt>`](https://en.cppreference.com/w/cpp/header/codecvt) header in C++17 - these are used in [`rosidl_runtime_cpp/traits.hpp`](https://github.com/ros2/rosidl/blob/master/rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp).
> ...
According to the deprecation statement, the codecvt _"library component should be retired to Annex D, along side , until a suitable replacement is standardized"_ . I guess this means that it won't be removed for a long time.

aprotyas:
> According to the deprecation statement, the codecvt _"library component should be retired to Annex D, along side , until a suitable replacement is standardized"_. I guess this means that it won't be removed for a long time.

In the spirit of not fixing what isn't broken, I propose this deprecation be silenced with the `_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING` macro. It looks like @clalancette has a working branch that achieves this, check https://github.com/ros2/rosidl/commit/0c283d0db26c0c73f76aa20f624b0510c8766c5b.

```cmake
if(MSVC)
  ament_export_definitions(_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING)
endif()
```

aprotyas:
> Should I merge that branch into this PR?

Sure - you could also cherry-pick https://github.com/ros2/rosidl/commit/8f855b72f6a53cf90e5762a801442b947f79c43b and https://github.com/ros2/rosidl/commit/0c283d0db26c0c73f76aa20f624b0510c8766c5b onto your working branch. That might be less friction than a merge.

aprotyas (approved):
Thanks for iterating! LGTM with green CI.

Rerunning CI from above:
 * repos: `https://gist.githubusercontent.com/sloretz/6afda0358f48c644a0d292aae353e449/raw/e2f14016b0815e8c89af6233752f9a1cee1b8320/ros2.repos`
 * build: ` --packages-up-to rosidl_runtime_c rosidl_runtime_c rosidl_typesupport_introspection_cpp rosidl_typesupport_interface rosidl_runtime_cpp rosidl_generator_cpp`
 * test: `--packages-select rosidl_runtime_c rosidl_runtime_c rosidl_typesupport_introspection_cpp rosidl_typesupport_interface rosidl_runtime_cpp rosidl_generator_cpp`
 
Jobs
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15668/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10346/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13330/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15892/)

aprotyas:
It looks like MSVC throws the deprecation warnings despite https://github.com/ros2/rosidl/pull/635/commits/7ec237ede2b345e1edd4748452d774c05bbb9c5d. I'm not sure how the definition exporting mechanics is supposed to work. Any clues? May just have to define the silencing macro in the [`rosidl_runtime_cpp/traits.hpp`](https://github.com/ros2/rosidl/blob/master/rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp) header.

oysstu:
> It looks like MSVC throws the deprecation warnings despite [7ec237e](https://github.com/ros2/rosidl/commit/7ec237ede2b345e1edd4748452d774c05bbb9c5d). I'm not sure how the definition exporting mechanics is supposed to work. Any clues? May just have to define the silencing macro in the [`rosidl_runtime_cpp/traits.hpp`](https://github.com/ros2/rosidl/blob/master/rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp) header.

I don't think the 'ament_export_definitions' macro adds the preprocessor definition at all, it seems like it may have something to do with exporting definitions to downstream packages, but I can't find any documentation. For now, I think it's best to simply use add_compile_options (ROS hasn't begun to use the modern target-specific cmake style as far as I know). Not sure how this affects other packages that include this header, I'm not that familiar with the ament_cmake machinery (perhaps this is what ament_export_definitions does).

Defining it in the header isn't a good option, as it needs to be defined before the STL libraries are pulled in ([https://devblogs.microsoft.com/cppblog/c17-feature-removals-and-deprecations/](https://devblogs.microsoft.com/cppblog/c17-feature-removals-and-deprecations/)). 

e.g.
```C++
if (MSVC)
    add_compile_options(_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING)
endif()```

**Comment by aprotyas** [comment]
> ROS hasn't begun to use the modern target-specific cmake style as far as I know

Every PR can be a small step in that direction! ;-)

> Defining it in the header isn't a good option, as it needs to be defined before the STL libraries

Understood. Since `add_compile_options` appends to the project-scope `COMPILE_OPTIONS` variable, it might be better to limit the scope of the compile definition to direct consumers as such?

```cmake
if(MSVC)
  target_compile_definitions(${PROJECT_NAME} INTERFACE
  _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING)
endif()
```

oysstu:
Ok, I've added the proposed change and rebased/squashed the commits. Hopefully the current change successfully disables the warning on MSVC.

aprotyas:
Thanks for iterating @oysstu! There's a linter error in the CMakeLists.txt file, I think replacing `if (MSVC)` with `if(MSVC)` should fix it. Can you commit that change and then I'll run CI again?

aprotyas:
Rerunning CI:
 * repos: `https://gist.githubusercontent.com/sloretz/6afda0358f48c644a0d292aae353e449/raw/e2f14016b0815e8c89af6233752f9a1cee1b8320/ros2.repos`
 * build: ` --packages-up-to rosidl_runtime_c rosidl_runtime_c rosidl_typesupport_introspection_cpp rosidl_typesupport_interface rosidl_runtime_cpp rosidl_generator_cpp`
 * test: `--packages-select rosidl_runtime_c rosidl_runtime_c rosidl_typesupport_introspection_cpp rosidl_typesupport_interface rosidl_runtime_cpp rosidl_generator_cpp`
 
Jobs
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15718/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10400/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13383/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15962/)

hidmic:
@aprotyas @clalancette shall we get this moving?

aprotyas:
I think this PR is ready to go, feel free to merge it please!

hidmic:
It's been a while. Running full CI to be on the safe side:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15876/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10553/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13542/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16179/)

aprotyas:
Test failures look unrelated. The windows test failure is reported as flaky.

hidmic:
> Test failures look unrelated. The windows test failure is reported as flaky.

Indeed. Going in. Thanks @oysstu !

PR #641, opened 2022-01-04
Add tests for code generated by rosidl_typesupport_introspection_c/cpp code generators

gbiggs:
This PR adds a new package that contains tests for code generated by the `rosidl_typesupport_introspection_c` and `rosidl_typesupport_introspection_cpp` code generator packages.

The testing methodology is to test that the generated code behaves correctly. The tests do not check that the generated code matches some template or a priori prepared code, as that is a fragile approach to testing a code generator and would lead to a greater maintenance burden in the long term. Instead it is assumed that working code is correct even if the shape of that code may change over time.

The current status of this PR is:

- Only tests for `rosidl_typesupport_introspection_c` are done so far. However the C++ tests are going to be very similar. So much so that I think significant code reuse should be possible, with some cleanup to enable it.
- There are still some gaps (noted with comments in the source file)
- They're still very messy (dead code leftover from experimenting, redundant code, things like that)
- The tests are not really integrated into something to be called via `colcon test`; I've been working by executing `./build/rosidl_typesupport_introspection_tests/test_typesupport_c` directly. Ideally the tests would generate code, compile it, then execute the gtest tests in order to verify the complete pipeline. Currently it relies on the existing typesupport libraries.

### DIFF
--- a/rosidl_typesupport_introspection_tests/CMakeLists.txt
+++ b/rosidl_typesupport_introspection_tests/CMakeLists.txt
@@ -1,0 +1,33 @@
+cmake_minimum_required(VERSION 3.8)
+project(rosidl_typesupport_introspection_tests)
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+# find dependencies
+find_package(ament_cmake REQUIRED)
+find_package(rosidl_typesupport_introspection_c REQUIRED)
+find_package(rosidl_typesupport_introspection_cpp REQUIRED)
+find_package(rosidl_typesupport_interface REQUIRED)
+
+if(BUILD_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
+
+  find_package(ament_cmake_gtest REQUIRED)
+  find_package(ament_cmake_gtest REQUIRED)
+  find_package(test_msgs REQUIRED)
+  ament_add_gtest(
+    test_typesupport_c
+    test/test_typesupport_c.cpp
+  )
+  if(TARGET test_typesupport_c)
+    target_compile_features(test_typesupport_c PRIVATE cxx_std_17)
+    ament_target_dependencies(test_typesupport_c
+      "test_msgs"
+    )
+  endif()
+endif()
+
+ament_package()
--- a/rosidl_typesupport_introspection_tests/test/test_typesupport_c.cpp
+++ b/rosidl_typesupport_introspection_tests/test/test_typesupport_c.cpp
@@ -1,0 +1,2215 @@
+// Copyright 2021 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include <dlfcn.h>
+
+#include <cstdio>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "rosidl_runtime_c/message_type_support_struct.h"
+#include "rosidl_runtime_c/service_type_support_struct.h"
+#include "rosidl_typesupport_introspection_c/field_types.h"
+#include "rosidl_typesupport_introspection_c/message_introspection.h"
+#include "rosidl_typesupport_introspection_c/service_introspection.h"
+
+#include "test_msgs/msg/arrays.h"
+#include "test_msgs/msg/basic_types.h"
+#include "test_msgs/msg/bounded_plain_sequences.h"
+#include "test_msgs/msg/bounded_sequences.h"
+#include "test_msgs/msg/builtins.h"
+#include "test_msgs/msg/constants.h"
+#include "test_msgs/msg/defaults.h"
+#include "test_msgs/msg/empty.h"
+#include "test_msgs/msg/multi_nested.h"
+#include "test_msgs/msg/nested.h"
+#include "test_msgs/msg/strings.h"
+#include "test_msgs/msg/unbounded_sequences.h"
+#include "test_msgs/msg/w_strings.h"
+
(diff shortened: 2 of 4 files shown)


rosidl_typesupport_introspection_tests/test/test_typesupport_c.cpp
```diff
@@ -0,0 +1,2215 @@
...
+
+typedef const rosidl_message_type_support_t * (*get_message_ts_function)();
+typedef const rosidl_service_type_support_t * (*get_service_ts_function)();
+
+std::string make_message_typesupport_function_name(
+  std::string ns,
+  std::string type,
+  std::string interface)
+{
+  return "rosidl_typesupport_introspection_c__get_message_type_support_handle__" +
```
hidmic, line 244: @gbiggs hmm, we should probably use [rosidl_typesupport_interface](https://github.com/ros2/rosidl/blob/67bcca666d818823aba8ea08c12160a9badf4e41/rosidl_typesupport_interface/include/rosidl_typesupport_interface/macros.h) macros here. That won't do in macro form. Maybe it is a good time to add function equivalents.

rosidl_typesupport_introspection_tests/test/test_typesupport_c.cpp
```diff
@@ -0,0 +1,2215 @@
...
+      all_messages.end(),
+      action_messages.begin(),
+      action_messages.end());
+    all_services.insert(
+      all_services.end(),
+      action_services.begin(),
+      action_services.end());
+
+    auto typesupport_lib_name = "lib" + ns + "__rosidl_typesupport_introspection_c.so";
+    ts_lib = dlopen(typesupport_lib_name.c_str(), RTLD_LAZY);
```
hidmic, line 122: @gbiggs using [`rcpputils::SharedLibrary`](https://github.com/ros2/rcpputils/blob/master/include/rcpputils/shared_library.hpp) may be best.

rosidl_typesupport_introspection_tests/test/test_typesupport_c.cpp
```diff
@@ -0,0 +1,2215 @@
...
+    std::string ns,
+    std::string type,
+    std::string interface)
+  {
+    return ns + "__" + type + "__" + interface + "__destroy";
+  }
+
+  void populate_constructors()
+  {
+    auto typesupport_lib_name = "lib" + ns + "__rosidl_generator_c.so";
```
hidmic, line 606: @gbiggs meta: did you consider templating the hell of out this instead of dynamic loading interface representations (i.e. `__rosidl_generator_c*.so` libraries)? Few tests below benefit from the latter. And I'm a bit wary about name mangling when it's C++ and not C that we're loading.

I'm thinking that some type traits and fully specialized fixtures could come in handy.

hidmic: Actually, where are these constructors being used?

rosidl_typesupport_introspection_tests/test/test_typesupport_c.cpp
```diff
@@ -0,0 +1,2215 @@
...
+  EXPECT_EQ(members[13].size_function, nullptr);
+  EXPECT_EQ(members[13].get_const_function, nullptr);
+  EXPECT_EQ(members[13].get_function, nullptr);
+  EXPECT_EQ(members[13].resize_function, nullptr);
+
+  EXPECT_STREQ(members[16].name_, "defaults_values");
+  EXPECT_EQ(members[16].type_id_, rosidl_typesupport_introspection_c__ROS_TYPE_MESSAGE);
+  EXPECT_EQ(members[16].string_upper_bound_, 0u);
+  // TODO(gbiggs): Test the contained members are correct to verify this field was initialised
+  // correctly
```
hidmic, line 1868: @gbiggs hmm, perhaps it makes sense to have `*AreCorrect` test cases validate members only, and then add more test cases that, like `CanCallInitFunction` and `CanCallFiniFunction`, test that type-erased and non type-erased code can interoperate: initialize statically, write dynamically, read statically, finalize dynamically (and other meaningful permutations).

hidmic (commented):
@gbiggs it's looking good! I left some thoughts, would be great to ping pong ideas before you're out on vacation.

rosidl_typesupport_introspection_tests/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+cmake_minimum_required(VERSION 3.8)
+project(rosidl_typesupport_introspection_tests)
```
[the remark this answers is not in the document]
gbiggs: Ideally I'd prefer that, but I think there's a lot of code that would end up being duplicated that way.

rosidl_typesupport_introspection_tests/test/test_typesupport_c.cpp
```diff
@@ -0,0 +1,2215 @@
...
+#include "test_msgs/srv/arrays.h"
+#include "test_msgs/srv/basic_types.h"
+#include "test_msgs/srv/empty.h"
+
+#include "test_msgs/action/fibonacci.h"
+#include "test_msgs/action/nested_message.h"
+
+TEST(library_management, can_load_typesupport_library)
+{
+  auto typesupport_lib_name = "libtest_msgs__rosidl_typesupport_introspection_c.so";
```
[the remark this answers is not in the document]
gbiggs: I think there are already macros for this pattern, possibly, which for reasons I forget I didn't use. If not, then I agree with adding them.

rosidl_typesupport_introspection_tests/test/test_typesupport_c.cpp
gbiggs: Totally agree.

rosidl_typesupport_introspection_tests/test/test_typesupport_c.cpp
gbiggs: That's a good idea. The C-language tests don't have to avoid C++, so using an existing utility for managing the shared library would make them more robust.

rosidl_typesupport_introspection_tests/test/test_typesupport_c.cpp
gbiggs: I don't think they are being used. This is left-over code from an avenue I was exploring for testing the message introspection. I forgot to remove these functions and variables when I abandoned that idea.

rosidl_typesupport_introspection_tests/test/test_typesupport_c.cpp
gbiggs: Testing the interoperation of type-erased and non-type-erased code is a good idea. That would help confirm that the type-erased code is actually operating on the in-memory data correctly, which is otherwise hard to do.

hidmic:
@gbiggs FYI, to avoid overlap, I'm working on the C++ version of this.

j-rivero:
I think that we can close this one, right @hidmic ?

j-rivero:
Closing in favor of #651

Issue #642, opened 2022-01-05
nested for index overwrite in PyEM template


ibnHatab:
Same index name used in nested loop. Potentially causing problems fex: removing trailing ',' in struct with conditional '@(i < len(msg.fields) - 1 ? ',')@'

This code :
@[  for **i,** field in enumerate(msg.fields)]@
@[      for **i,** line in enumerate(field.annotations['comment'])]

https://github.com/ros2/rosidl/blob/2ec002300a41290ea69092ae883f4c5c02661939/rosidl_adapter/rosidl_adapter/resource/struct.idl.em#L76

clalancette:
Good catch.  Would you mind submitting a PR renaming at least one of them?

ibnHatab:
Sure, PR submitted : https://github.com/ros2/rosidl/pull/643
pytest passed on rosidl_adapter but I don't have dev env to isolate and recheck everything.

clalancette:
This was fixed by #643, so closing.  Thanks for the contribution!

PR #643, opened 2022-01-06
rename nested loop index

ibnHatab:
Fixing minor issue with loop index mentioned in:
https://github.com/ros2/rosidl/issues/642

rosidl_adapter  pytest passed and no changes in IDL output visible.

### DIFF
--- a/rosidl_adapter/rosidl_adapter/resource/struct.idl.em
+++ b/rosidl_adapter/rosidl_adapter/resource/struct.idl.em
@@ -68,14 +68,14 @@ else:
     struct @(msg.msg_name) {
 @[if msg.fields]@
 @[  for i, field in enumerate(msg.fields)]@
-@[if i > 0]@
+@[    if i > 0]@
 
-@[end if]@
+@[    end if]@
 @[    if field.annotations.get('comment', [])]@
       @@verbatim (language="comment", text=@
-@[      for i, line in enumerate(field.annotations['comment'])]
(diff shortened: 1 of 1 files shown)


clalancette:
@ibnHatab After looking at this for a bit, I think we actually don't even need the iterator in the outer loop.  We never use it.  So I opened up https://github.com/ibnHatab/rosidl/pull/1 , which just removes the iterator in the outer loop completely.  If you agree with that, please merge it in (which will update this PR).  Also, if there is a specific example problem that you ran into for this, it would be great if you could write a short test that shows the problem.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15970/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10641/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16292/)

clalancette:
Rebased, so here is another CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15979/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10649/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16301/)

clalancette:
The test failures on Windows are known.  Thanks for the contribution @ibnHatab !

Issue #598, opened 2021-06-14
Introspection bool arrays don't have get_function, get_const_function, resize_function and size_function 


StefanFabian:
On Foxy, bool arrays don't have any of the usual array functions in `rosidl_typesupport_introspection_c(pp)`.
This seems to be the case only for bool arrays, hence, I assume this is not intended.

hidmic:
This fell off my radar. Thanks for reporting @neko-suki.

nnmm:
In my experience this happens for all primitive values, and also for bounded sequences (but only on Foxy).

PR #647, opened 2022-01-20
Support nested types introspection in C/C++

hidmic:
Precisely what the title says. While working on typesupport introspection tests, I discovered that (a) the API for basic type arrays' introspection in C was completely missing (#598), and (b) the API for bool arrays' introspection in C++ was missing (and arguably broken due to `std::vector<bool>` quirks). 

This patch amends this by, among other things, introducing an `std::vector` subclass with its own specialization for `bool`. Note that since `std::vector<bool>` does not necessarily use contiguous storage, it makes type erased access to its elements via `void *` outright impossible (and that's exactly what introspection APIs do).

### DIFF
--- a/rosidl_generator_cpp/resource/idl__struct.hpp.em
+++ b/rosidl_generator_cpp/resource/idl__struct.hpp.em
@@ -27,8 +27,8 @@ include_directives = set()
 #include <array>
 #include <memory>
 #include <string>
-#include <vector>
 
+#include "rosidl_runtime_cpp/vector.hpp"
 #include "rosidl_runtime_cpp/bounded_vector.hpp"
 #include "rosidl_runtime_cpp/message_initialization.hpp"
 
--- a/rosidl_runtime_cpp/include/rosidl_runtime_cpp/vector.hpp
+++ b/rosidl_runtime_cpp/include/rosidl_runtime_cpp/vector.hpp
@@ -1,0 +1,179 @@
+// Copyright 2022 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ROSIDL_RUNTIME_CPP__VECTOR_HPP_
+#define ROSIDL_RUNTIME_CPP__VECTOR_HPP_
+
+#include <memory>
+#include <type_traits>
+#include <vector>
+
+namespace rosidl_runtime_cpp
+{
+namespace internal
+{
+/// A bool value wrapper to avoid std::vector<bool> specializations.
+/**
+ * For all practical purposes, an instance of this class is a bool value.
+ */
+class BoolWrapper
+{
+public:
+  constexpr BoolWrapper()
+  : value_(false) {}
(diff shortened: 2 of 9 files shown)


j-rivero:
Could the `#include <vector>` be removed from here https://github.com/ros2/rosidl/pull/647/files#diff-a05f4acc14bbcd36da0dfc3e0b606020ec48d5294b58f06b3cf9f9e593aa3cf1R21? 

A totally undocumented questions: the same that happen for `std::vector` could happen for `std::array`? There is one https://github.com/ros2/rosidl/blob/master/rosidl_generator_cpp/rosidl_generator_cpp/__init__.py#L130.

rosidl_generator_cpp/resource/idl__struct.hpp.em
```diff
@@ -29,6 +29,7 @@ include_directives = set()
 #include <string>
 #include <vector>
```
j-rivero, line 30: Would the include of vector now be superfluous?

j-rivero (approved):
I can not detect any error, although my knowledge of the package is very limited. A couple of minor questions before merging thought.

rosidl_generator_cpp/resource/idl__struct.hpp.em
hidmic: Good point.

hidmic:
> A totally undocumented questions: the same that happen for std::vector could happen for std::array? There is one https://github.com/ros2/rosidl/blob/master/rosidl_generator_cpp/rosidl_generator_cpp/__init__.py#L130.

@j-rivero fortunately, no. Only `std::vector<bool>` is special in an awkward sense.

gbiggs (approved):

hidmic:
Full CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15987/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10657/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16309/)

rosidl_runtime_cpp/include/rosidl_runtime_cpp/vector.hpp
```diff
@@ -0,0 +1,223 @@
...
+
+  using Base::vector;
+
+  // Implicit "copy" and "move" constructors from std::vector<bool> instances
+  Vector(const std::vector<bool, AllocatorT> & other)  // NOLINT(runtime/explicit)
+  : Base(other.begin(), other.end())
+  {
+  }
+
+  Vector(std::vector<bool, AllocatorT> && other)  // NOLINT(runtime/explicit)
```
ivanpauno, line 149: I think we should keep this implicit conversions, to make the transition smooth, but they are really bad performance wise.
IMHO, things that are costly should only be possible in an explicit way.

So I think we should add the implicit conversions with a deprecatd message, and provide explicit alternatives (e.g. a static method `from_std_vector_bool()`).
I don't mind much though.

hidmic: > I think we should keep this implicit conversions, to make the transition smooth, but they are really bad performance wise.

Yeah, I didn't want to cause breakage downstream but you're right.

> So I think we should add the implicit conversions with a deprecatd message, and provide explicit alternatives

That sounds good to me. Some `Vector::assign` overload and/or making constructors explicit would do too.

hidmic:
@ivanpauno @j-rivero I'm rolling back with this. Too many C++ template partial specializations that work with `std::vector` are breaking downstream (and it doesn't seem like I can seamlessly bend template resolution rules to stop that from happening). 

I'll re-open this PR with just the C stuff. We'll have to live with `std::vector<bool>`.

ivanpauno:
> Too many C++ template partial specializations that work with std::vector are breaking downstream

I'm curious what those look like.
Maybe introducing a `VectorBoolContinuous` and continue using `std::vector<T>` for everything else reduces the ammount of code being break in downstream packages.

Though that's maybe out of scope of what you're trying to do.

PR #648, opened 2022-01-20
Implement equality operator function for C messages.

hidmic:
Precisely what the title says. While working on typesupport introspection tests, I realized that while `operator==` is overloaded for C++ messages, there's no such functionality for C messages. Thus a lot of code would have to be written to compare C messages for equality. This patch amends this.


hidmic:
This is actually lacking some tests. I'll add them.

rosidl_runtime_c/test/test_string_functions.cpp
```diff
@@ -116,6 +116,32 @@ TEST(string_functions, resize_assignn) {
   rosidl_runtime_c__String__fini(&s);
 }
 
+TEST(string_functions, equality_comparison) {
+  rosidl_runtime_c__String empty_string, foo_string, bar_string;
+  EXPECT_TRUE(rosidl_runtime_c__String__init(&empty_string));
```
gbiggs, line 121: These could be asserts to break early if initialisation fails.

gbiggs (approved):
LGTM

rosidl_runtime_c/test/test_string_functions.cpp
hidmic: My understanding is that no tests in here do that to ensure `rosidl_runtime_c__String__fini` gets invoked. I could bring in `rcpputils::scope_exit` though. Thoughts?

hidmic:
CI up to `rosidl_generator_c` and `rosidl_runtime_c` (nothing uses this API just yet):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15990/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10659/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16313/)

hidmic:
Alright, going in!

PR #649, opened 2022-01-20
Add ROSIDL_TYPESUPPORT_INTERFACE__LIBRARY_NAME() macro

hidmic:
Precisely what the title says. Code that uses typesupport (generated) APIs often has to be able to find the corresponding shared library. The pattern `<package_name>__<typesupport_name>` is relied on and recreated in several places (in this repository and downstream). This patch amends this.


j-rivero (approved):

hidmic:
Thanks @j-rivero. No code is using this macro just yet, so I'll move forward with just the PR job.

PR #659, opened 2022-02-15
Update Quality declaration to level 1 in README for instrospection pkgs

j-rivero:
Thanks to @emersonknapp for catching the problem. Another alternative would be to remove the Quality level in the READMEs but I left them here to be consistent with the rest of the `rosidl` repository. I quick look into how things are in other repositories (rclcpp and rcl) seems like there is no homogeneous way to deal with this.

### DIFF
--- a/rosidl_typesupport_introspection_c/README.md
+++ b/rosidl_typesupport_introspection_c/README.md
@@ -8,4 +8,4 @@ The features provided by `rosidl_typesupport_introspection_c` are described in [
 
 ## Quality Declaration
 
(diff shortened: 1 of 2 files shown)


clalancette (approved):
Sigh, yes.  I argued against duplicating the information in these files at all for exactly this reason, but I was overridden on that decision.  I'd much rather this say "look at QUALITY_DECLARATION.md for the quality level".

That said, we don't need to do that here, so looks good to me.

gbiggs (approved):

PR #661, opened 2022-02-16
De-duplicate Quality Level from README and QUALITY_DECLARATION

j-rivero:
Follow up from #659


clalancette (approved):
I think this is good, but I'd like to hear from @ahcorde before we go ahead and merge this.

ahcorde (approved):

PR #662, opened 2022-02-17
Misc cleanup in the rosidl generator extensions

sloretz:
This PR cleans up a few things I've noticed while looking at the generator code for another task

* Make sure to `find_package()` dependencies needed by generated code
* Don't wrap the `target_compile_definitions(...BUILDING_DLL...)` line in an `if(WIN32)` block because it's unnecessary. The visibility header seems to do stuff on other platforms too.
* Instead of depending on all targets in a dependency interface package, depend on just the targets generated by the same generator using the suffixed `.._TARGETS${_target_suffix}` variable created by `rosidl_export_typesupport_targets()`.
  * also use `rosidl_export_typesupport_targets()` where it wasn't already
* Remove `if(NOT _generated_headers...` checks - because they don't seem to do anything. In all places I saw them there was another place further up or down that wasn't guarded and would have raised a CMake error  if that variable actually was empty (either the `add_library()` or `add_custom_target()` commands).
* Use `target_link_libraries()` instead of `ament_target_dependencies()` ament/ament_cmake#292 in the generator code (There are a few other places it's used in this repo still, but they weren't part of the generators so I didn't include them in this PR)
* Replace checks on `rosidl_generator_c_FOUND` with checks that the target that generator generates is found - because that's the important thing. 
   * Note to self: generators can control what order their extension is called by `find_packaging()` the generators they depend on before registering their own extensions in their "extras.cmake" file
* When a generator depends on another, make it `target_link_libraries()` to the generated target of the other rather than hard-coding the directory it's likely to create in the build folder
* Avoid calling `ament_export_dependencies(...)` when `SKIP_INSTALL` is used.


sloretz:
I don't see these failures in my focal container, but the PR job is the first one ever on Jammy https://build.ros2.org/job/Rpr__rosidl__ubuntu_jammy_amd64/1 . I'll try making a Jammy container and reproducing there.

sloretz:
I get the same test failures on Jammy, with and without this PR - so they're not caused by it. Still, those will need to be fixed for green CI.

sloretz:
CI (build: `--packages-up-to rcl_interfaces` test: `--packages-select  rosidl_generator_c  rosidl_generator_cpp rosidl_typesupport_introspection_cpp rosidl_typesupport_introspection_c rcl_interfaces`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16195/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10822/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16551/)

sloretz:
@ros-pull-request-builder retest this please

clalancette (commented):
I did a review of this, and I left two comments inline.  Both of them are really questions since I'm not that familiar with how all of this works.  It may be a good idea to get another review from someone like @hidmic before going ahead with this.

rosidl_generator_c/cmake/rosidl_generator_c_generate_interfaces.cmake
```diff
@@ -116,81 +118,81 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
...
   "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_c>"
   "$<INSTALL_INTERFACE:include>"
 )
 foreach(_pkg_name ${rosidl_generate_interfaces_DEPENDENCY_PACKAGE_NAMES})
-  ament_target_dependencies(
-    ${rosidl_generate_interfaces_TARGET}${_target_suffix}
-    ${_pkg_name})
-  if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
-    ament_export_dependencies(${_pkg_name})
-  endif()
```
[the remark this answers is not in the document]
sloretz: `rosidl_generate_interfaces()` [already calls `ament_export_dependencies()` on these](https://github.com/ros2/rosidl/blob/90a74c3158f1f5b5ccf334f63af0b855700c8a77/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake#L69). This removes `rosidl_generator_c` doing it an extra time.

rosidl_generator_cpp/cmake/rosidl_generator_cpp_generate_interfaces.cmake
```diff
@@ -112,53 +103,71 @@ target_include_directories(${rosidl_generate_interfaces_TARGET}${_target_suffix}
...
+# TODO(sloretz) make this target name less generic than "__cpp" when other
+# generators no longer depend on it
+add_dependencies(
+  ${rosidl_generate_interfaces_TARGET}${_target_suffix}
+  ${rosidl_generate_interfaces_TARGET}__cpp)
+
+# Make top level generation target depend on this generated library
+add_dependencies(
+  ${rosidl_generate_interfaces_TARGET}
+  ${rosidl_generate_interfaces_TARGET}${_target_suffix})
```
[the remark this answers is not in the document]
sloretz: When a package calls `rosidl_generate_interfaces(foobar` they're passing in a name to create for a top-level interface generation target. That target name is passed to the individual generators in the variable `${rosidl_generate_interfaces_TARGET}`. Each generator creates their own target to do message generation, and they all do the same pattern where they define a variable called `_target_suffix` and then make their generator specific target with the name `${rosidl_generate_interfaces_TARGET}${_target_suffix}`. The second `add_dependencies()` target is making the top level target depend on the generator specific target. That way, any other target that depends on the top level target will wait until all generators finish before being built. That's part of how targets can use generated messages in the same package.

`rosidl_generator_cpp` only generates headers, so it's library target is a header only library, which in CMake is an `INTERFACE` library. CMake doesn't provide a way for an `INTERFACE` library to directly depend on a file, so there's nothing to make targets who depend on the `INTERFACE` library wait for the header files to be generated. To solve this, `rosidl_generator_cpp` creates an extra custom target called `${rosidl_generate_interfaces_TARGET}__cpp`. CMake does allow custom targets to depend on generated files, and that's all this one does. The first `add_dependency()` makes the generator's `INTERFACE` library depend on that custom target so that anyone who depends on the library will wait to be built until the headers are generated.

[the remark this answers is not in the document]
sloretz: Added comments explaining these in 8a433e9

sloretz:
CI again (build: all test: `--packages-select rosidl_generator_c rosidl_generator_cpp rosidl_typesupport_introspection_cpp rosidl_typesupport_introspection_c rcl_interfaces rcl`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16302/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10922/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16669/)

rosidl_generator_c/cmake/rosidl_generator_c_generate_interfaces.cmake
```diff
@@ -116,81 +118,81 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     C_STANDARD 11
     COMPILE_OPTIONS -Wall -Wextra -Wpedantic)
 endif()
-if(WIN32)
-  target_compile_definitions(${rosidl_generate_interfaces_TARGET}${_target_suffix}
-    PRIVATE "ROSIDL_GENERATOR_C_BUILDING_DLL_${PROJECT_NAME}")
-endif()
+target_compile_definitions(${rosidl_generate_interfaces_TARGET}${_target_suffix}
+  PRIVATE "ROSIDL_GENERATOR_C_BUILDING_DLL_${PROJECT_NAME}")
```
hidmic, line 122: @sloretz nit: `ROSIDL_GENERATOR_C_BUILDING_DLL_*` macros are only used in visibility headers on Windows (e.g. see [here](https://github.com/ros2/rosidl/blob/67bcca666d818823aba8ea08c12160a9badf4e41/rosidl_generator_c/resource/rosidl_generator_c__visibility_control.h.in#L23)).

rosidl_generator_cpp/cmake/rosidl_generator_cpp_generate_interfaces.cmake
```diff
@@ -112,53 +108,72 @@ target_include_directories(${rosidl_generate_interfaces_TARGET}${_target_suffix}
...
+
+  # Export modern CMake targets
+  ament_export_targets(export_${rosidl_generate_interfaces_TARGET}${_target_suffix})
+  rosidl_export_typesupport_targets(${_target_suffix}
+    ${rosidl_generate_interfaces_TARGET}${_target_suffix})
 
   install(
     TARGETS ${rosidl_generate_interfaces_TARGET}${_target_suffix}
-    EXPORT ${rosidl_generate_interfaces_TARGET}${_target_suffix}
+    EXPORT export_${rosidl_generate_interfaces_TARGET}${_target_suffix}
```
hidmic, line 151: @sloretz nit: why add an `export_` suffix here but not for `rosidl_generator_c`?

hidmic (approved):
Only a couple nits. Overall LGTM

rosidl_generator_cpp/cmake/rosidl_generator_cpp_generate_interfaces.cmake
sloretz: I'm inconsistent. Added `export_` in 368d022

rosidl_generator_c/cmake/rosidl_generator_c_generate_interfaces.cmake
sloretz: Oh oops. I read that header wrong.

Interestingly CMake has a target property for this already called [`DEFINE_SYMBOL`](https://cmake.org/cmake/help/v3.10/prop_tgt/DEFINE_SYMBOL.html). It adds the compiler definition `<target_name>_EXPORTS` by default, setting the property changes it. I changed it to use that property instead in 519066b

sloretz:
CI re-run
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16334/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10951/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16703/)

rosidl_typesupport_introspection_c/cmake/rosidl_typesupport_introspection_c_generate_interfaces.cmake
```diff
@@ -108,41 +108,39 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     C_STANDARD 11
     COMPILE_OPTIONS -Wall -Wextra -Wpedantic)
 endif()
-if(WIN32)
-  target_compile_definitions(${rosidl_generate_interfaces_TARGET}${_target_suffix}
-    PRIVATE "ROSIDL_TYPESUPPORT_INTROSPECTION_C_BUILDING_DLL_${PROJECT_NAME}")
-endif()
+target_compile_definitions(${rosidl_generate_interfaces_TARGET}${_target_suffix}
+  PRIVATE "ROSIDL_TYPESUPPORT_INTROSPECTION_C_BUILDING_DLL_${PROJECT_NAME}")
```
hidmic, line 112: @sloretz shall we use `DEFINE_SYMBOL` here too?

rosidl_typesupport_introspection_cpp/cmake/rosidl_typesupport_introspection_cpp_generate_interfaces.cmake
```diff
@@ -92,84 +96,83 @@ if(rosidl_generate_interfaces_LIBRARY_NAME)
   set_target_properties(${rosidl_generate_interfaces_TARGET}${_target_suffix}
     PROPERTIES OUTPUT_NAME "${rosidl_generate_interfaces_LIBRARY_NAME}${_target_suffix}")
 endif()
-if(WIN32)
-  target_compile_definitions(${rosidl_generate_interfaces_TARGET}${_target_suffix}
-    PRIVATE "ROSIDL_TYPESUPPORT_INTROSPECTION_CPP_BUILDING_DLL")
-endif()
+target_compile_definitions(${rosidl_generate_interfaces_TARGET}${_target_suffix}
+  PRIVATE "ROSIDL_TYPESUPPORT_INTROSPECTION_CPP_BUILDING_DLL")
```
hidmic, line 100: @sloretz same about using `DEFINE_SYMBOL`

rosidl_generator_c/cmake/rosidl_generator_c_generate_interfaces.cmake
```diff
@@ -116,81 +118,81 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     C_STANDARD 11
     COMPILE_OPTIONS -Wall -Wextra -Wpedantic)
 endif()
-if(WIN32)
-  target_compile_definitions(${rosidl_generate_interfaces_TARGET}${_target_suffix}
-    PRIVATE "ROSIDL_GENERATOR_C_BUILDING_DLL_${PROJECT_NAME}")
-endif()
+set_property(TARGET ${rosidl_generate_interfaces_TARGET}${_target_suffix}
+  PROPERTY DEFINE_SYMBOL "ROSIDL_GENERATOR_C_BUILDING_DLL_${PROJECT_NAME}")
```
hidmic, line 122: @sloretz unconditionally? I'm OK with that, though talking about DLLs outside Windows is awkward.

sloretz: Looks that way. If you build `rcl` with `-DCMAKE_VERBOSE_MAKEFILE=ON` all the compiler invocations have `-Drcl_EXPORTS` passed to them; which is the default value of `DEFINE_SYMBOL`.

rosidl_typesupport_introspection_c/cmake/rosidl_typesupport_introspection_c_generate_interfaces.cmake
sloretz: [ea3a3ca](https://github.com/ros2/rosidl/pull/662/commits/ea3a3cae4eda7c65ce446627a24dd5f0278d10f5)

sloretz:
CI re-run
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16349/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10961/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16713/)

PR #650, opened 2022-01-20
Implement copy function for C messages

hidmic:
Precisely what the title says. While C++ messages overload the assignment operator, plain assignment of C messages results in shallow copy of pointers which in turn leads to ownership issues (e.g. double free crashes) down the line. This patch amends this.

Depends on #648.

PR #667, opened 2022-03-08
Foxy backport of #648 and #650

nnmm:
This is helpful for the ros2_rust project.

I also backported the equality functions, since without them there were merge conflicts.

### DIFF
--- a/rosidl_runtime_c/src/string_functions.c
+++ b/rosidl_runtime_c/src/string_functions.c
@@ -245,9 +245,9 @@ rosidl_runtime_c__String__Sequence__copy(
       }
     }
     output->data = data;
-    output->size = input->size;
     output->capacity = input->size;
   }
+  output->size = input->size;
   for (size_t i = 0; i < input->size; ++i) {
     if (!rosidl_runtime_c__String__copy(
         &(input->data[i]), &(output->data[i])))
--- a/rosidl_runtime_c/src/string_functions.c
+++ b/rosidl_runtime_c/src/string_functions.c
@@ -128,6 +128,18 @@ rosidl_runtime_c__String__assign(
 }
 
 bool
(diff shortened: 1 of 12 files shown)


nnmm:
@hidmic Could you review?

rosidl_runtime_c/src/string_functions.c
```diff
@@ -171,6 +197,67 @@ rosidl_runtime_c__String__Sequence__fini(
...
+        /* free currently allocated and return false */
+        for (; i-- > output->capacity; ) {
+          rosidl_runtime_c__String__fini(&data[i]);
+        }
+        free(data);
+        return false;
+      }
+    }
+    output->data = data;
+    output->size = input->size;
```
nnmm, line 248: In cases where the output capacity is greater than the input size, the output size is not set because this line is not executed.

It seems to me that this is a bug. There are two possible ways to fix it:
- Set the output size unconditionally
- Set the output size to the input size when the input size is larger.

hidmic: > It seems to me that this is a bug. 

It is. 

> Set the output size unconditionally

That gets us the closest to a regular copy constructor.

---

Would you be willing to contribute such a fix @nnmm ?

nnmm: Sure. I just pushed a new commit to the backport PRs and will make a new PR for master.

hidmic (approved):
Foxy CI up to `rosidl_generator_c` and `rosidl_runtime_c`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16579/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11162/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16966/)

PR #669, opened 2022-03-09
Set the output size unconditionally when copying sequences

nnmm:
Fixes the bug mentioned in https://github.com/ros2/rosidl/pull/667#discussion_r822925405

Signed-off-by: Nikolai Morin <nnmmgit@gmail.com>

PR #671, opened 2022-03-21
Fix error handling when copying C sequence messages

hidmic:
This is an issue I spotted while reviewing #669. We shouldn't hold reallocated (i.e. invalidated) data pointers if any new sequence item initialization fails. It's quite hard to run into this problem (e.g. RAM exhaustion), but it is theoretically possible.

CI up to `rosidl_generator_c` and `rosidl_runtime_c`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16390/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10999/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16762/)

### DIFF
--- a/rosidl_generator_c/resource/msg__functions.c.em
+++ b/rosidl_generator_c/resource/msg__functions.c.em
@@ -471,17 +471,16 @@ bool
     if (!data) {
       return false;
     }
+    output->data = data;
     for (size_t i = output->capacity; i < input->size; ++i) {
-      if (!@(message_typename)__init(&data[i])) {
+      if (!@(message_typename)__init(&output->data[i])) {
         /* free currently allocated and return false */
         for (; i-- > output->capacity; ) {
-          @(message_typename)__fini(&data[i]);
+          @(message_typename)__fini(&output->data[i]);
         }
-        allocator.deallocate(data, allocator.state);
         return false;
       }
     }
-    output->data = data;
     output->capacity = input->size;
   }
   output->size = input->size;
--- a/rosidl_generator_c/resource/msg__functions.c.em
+++ b/rosidl_generator_c/resource/msg__functions.c.em
@@ -471,10 +471,14 @@ bool
     if (!data) {
       return false;
     }
+    // If reallocation succeeded, memory may or may not have been moved
+    // to fulfill the allocation request, invalidating output->data.
     output->data = data;
     for (size_t i = output->capacity; i < input->size; ++i) {
       if (!@(message_typename)__init(&output->data[i])) {
-        /* free currently allocated and return false */
+        // If initialization of any new item fails, roll back
+        // all previously initialized items. Existing items
+        // in output are to be left unmodified.
         for (; i-- > output->capacity; ) {
           @(message_typename)__fini(&output->data[i]);
         }
--- a/rosidl_runtime_c/src/string_functions.c
+++ b/rosidl_runtime_c/src/string_functions.c
@@ -243,17 +243,16 @@ rosidl_runtime_c__String__Sequence__copy(
     if (!data) {
       return false;
     }
+    output->data = data;
     for (size_t i = output->capacity; i < input->size; ++i) {
-      if (!rosidl_runtime_c__String__init(&data[i])) {
+      if (!rosidl_runtime_c__String__init(&output->data[i])) {
         /* free currently allocated and return false */
         for (; i-- > output->capacity; ) {
-          rosidl_runtime_c__String__fini(&data[i]);
+          rosidl_runtime_c__String__fini(&output->data[i]);
         }
-        allocator.deallocate(data, allocator.state);
         return false;
       }
     }
-    output->data = data;
     output->capacity = input->size;
   }
   output->size = input->size;
--- a/rosidl_runtime_c/src/string_functions.c
+++ b/rosidl_runtime_c/src/string_functions.c
@@ -243,10 +243,14 @@ rosidl_runtime_c__String__Sequence__copy(
     if (!data) {
       return false;
     }
+    // If reallocation succeeded, memory may or may not have been moved
+    // to fulfill the allocation request, invalidating output->data.
     output->data = data;
     for (size_t i = output->capacity; i < input->size; ++i) {
       if (!rosidl_runtime_c__String__init(&output->data[i])) {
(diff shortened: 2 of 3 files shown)


rosidl_generator_c/resource/msg__functions.c.em
```diff
@@ -471,17 +471,16 @@ bool
...
+          @(message_typename)__fini(&output->data[i]);
         }
-        allocator.deallocate(data, allocator.state);
         return false;
       }
     }
-    output->data = data;
     output->capacity = input->size;
   }
   output->size = input->size;
```
clalancette, line 486: This isn't directly related to what you are fixing here, but should we move this *after* the loop below that copies the individual fields?  That way if one of them fails, we aren't updating `output->size` erroneously (this is probably never a problem in practice, but I still thought I would point it out).

rosidl_generator_c/resource/msg__functions.c.em
```diff
@@ -471,17 +471,16 @@ bool
...
+    output->data = data;
     for (size_t i = output->capacity; i < input->size; ++i) {
-      if (!@(message_typename)__init(&data[i])) {
+      if (!@(message_typename)__init(&output->data[i])) {
         /* free currently allocated and return false */
         for (; i-- > output->capacity; ) {
-          @(message_typename)__fini(&data[i]);
+          @(message_typename)__fini(&output->data[i]);
         }
-        allocator.deallocate(data, allocator.state);
```
clalancette, line 480: I'm not sure I agree with the error handling here.  Let's assume for a minute that we started copying the elements here one at a time, and one of them failed, so this method returned false.  In that case, the most reasonable thing for a caller of these functions to do is to call the `__fini` method on this array.  The problem with that sequence, though, is that we've already freed up some of the elements here, and the `__fini` method will try to free those elements again, leading to a double-free.  What I *think* we should do here is probably set `output->size` to 0 on error, and then have the `__fini` method free only up to `output->size`; that way we won't double-free things.  What do you think?

clalancette (commented):
I've left a couple of comments on how to clean things up here.  Let me know what you think.

rosidl_generator_c/resource/msg__functions.c.em
sloretz: > In that case, the most reasonable thing for a caller of these functions to do is to call the __fini method on this array. 

`fini` the input or the output?

> What I think we should do here is probably set output->size to 0 on error

I think this loop is only finalizing the things beyond `output`'s capacity before `copy` was called, so if size was set to 0 then all of output needed to be finalized in this function too.

> and then have the __fini method free only up to output->size; that way we won't double-free things. What do you think?

It looks like all elements up to `capacity` can be initialized at the moment. Say there are two arrays `foo` and `bar`. `foo` has size and capacity 5 and `bar` has size and capacity 7. Say `foo` is copied into `bar`. Because `bar` has excess capacity, the whole `if(output->capacity < input->size)` block is skipped. `foo`'s 5 elements get copied into `bar` and `bar` now has size 5 and capacity 7. If `fini` only free's up to `output->size` then the last two elements in `bar` will be leaked because `copy` didn't free them.

Finalizing up to `size` seems like it could be made to work, but everything needs to make sure only elements up to `size` are initialized. There would need to be another loop in copy to fini `output`'s elements that go beyond the new size when `input->size < output->size`.

rosidl_generator_c/resource/msg__functions.c.em
hidmic: @clalancette even more so, we should probably be incrementing it as we copy.

rosidl_generator_c/resource/msg__functions.c.em
hidmic: > I think this loop is only finalizing the things beyond output's capacity before copy was called, so if size was set to 0 then all of output needed to be finalized in this function too.

That's correct.

> The problem with that sequence, though, is that we've already freed up some of the elements here, and the __fini method will try to free those elements again, leading to a double-free.

A double-free isn't a possible failure mode, as message finalization functions are typically idempotent. We're not consistent about that across core packages though. 

> What I think we should do here is probably set output->size to 0 on error, and then have the __fini method free only up to output->size; that way we won't double-free things.

> Finalizing up to size seems like it could be made to work, but everything needs to make sure only elements up to size are initialized.

`__fini` functions operate under the assumption that elements up to capacity will have been default initialized. `__copy` functions simply follow suit. We can revisit that assumption though, to allocate up to capacity and initialize up to size instead.

Thoughts?

clalancette: @hidmic was kind enough to work through this with me, and I now understand a lot more about what we are doing here.  In short, if the `realloc` on line 469 succeeds, we *must* assign it to `output->data`, because that memory may have been moved and now the original pointer is bogus.  So that explains the first line of the patch.

The rest falls out of that.   `__fini` expects that up to `->capacity` is valid memory that has been default initialized.  If we failed to initialize one of the elements during the copy, then we roll back the *new* elements we've copied, leaving the existing ones alone.  We also do not update `->capacity`.  That fulfills `__fini`'s requirements (we end up wasting the additional memory that we just allocated, but we don't lose it or leak it).

So with all of that said, I think this refactor is the correct thing.  The only thing I'll request is a small comment somewhat explaining the situation, since it is not obvious (at least, it wasn't obvious to me).

I'm still going to approve, since the code itself seems correct.

clalancette (approved):

rosidl_generator_c/resource/msg__functions.c.em
hidmic: > @hidmic was kind enough to work through this with me

Thank you @clalancette for sparing the time to look at this. 

Added comments in https://github.com/ros2/rosidl/pull/671/commits/3a09837b0a9126ea513ead0e4a91de6ba30de4f3. PTAL!

clalancette (approved):

hidmic:
Since https://github.com/ros2/rosidl/pull/671/commits/3a09837b0a9126ea513ead0e4a91de6ba30de4f3 is comments only and the Rpr job passed, I'll merge.

PR #672, opened 2022-03-24
Make rosidl_get_typesupport_target return -NOTFOUND instead of FATAL_ERROR

sloretz:
This is in support of conditionally using typesupport targets: https://github.com/ros2/common_interfaces/pull/183

It allows using `rosidl_get_typesupport_target` even if the target doesn't exist. If it doesn't, the function sets the output variable to the expected target name ending in `-NOTFOUND`. That allows it to be checked with [if()](https://cmake.org/cmake/help/latest/command/if.html) which evaluates any string ending with that suffix as a false constant.

### DIFF
--- a/rosidl_cmake/cmake/rosidl_get_typesupport_target.cmake
+++ b/rosidl_cmake/cmake/rosidl_get_typesupport_target.cmake
@@ -33,7 +33,8 @@ function(rosidl_get_typesupport_target var generate_interfaces_target typesuppor
   set(output_target "${generate_interfaces_target}__${typesupport_name}")
 
   if(NOT TARGET ${output_target})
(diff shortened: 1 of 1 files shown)


pablogs9 (approved):
LGTM

sloretz:
CI (build: `--packages-above-and-dependencies rosidl_cmake` test: `--packages-select rosidl_cmake`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16429/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11034/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16797/)

gbiggs:
Re-run of Linux CI (now that the failing CI node has been excised): [Build Status](https://ci.ros2.org/job/ci_linux/16449/)

PR #683, opened 2022-06-06
Fix comment in camel case conversion function

sloretz:
This fixes a comment in `convert_camel_case_to_lower_case_underscore`. I reimplemented this function in bazel based on the comment and had a bug. In my bazel function `ColorRGBA.msg` became `color_rgb_a.msg` because `.` isn't an upper case letter. The correct result is `color_rgba.msg`.


sloretz:
Since this is only fixing a comment, I think the PR job is enough coverage for CI.

clalancette (approved):

PR #697, opened 2022-08-05
feat: add idl to rosidl spec converter

russkel:
The tooling within eProsima's *Integration Service ROS2 Handle* uses MessageSpecifications to perform code-gen. I have some messages that are IDL format and have added support by using the `rosidl_parser` output and transforming to `rosidl_adapter` Specification objects. I was surprised this didn't already exist within the code base so I have created a PR. There isn't a whole lot of code here.

## Uncertainties that I need help with:
- There is no `wchar` in rosidl? I have added a hack to convert it to `wstring<=1`. I can't help but think this isn't a good solution.
- There is no `long double` type in rosidl either. I have mapped it to a `float64` as a stop gap solution and print a warning, but this probably isn't good enough.

## TODO:
- Services, Actions
- Comments
- More tests
- CLI converter from IDL to msg/srv

### DIFF
--- a/rosidl_parser/rosidl_parser/idl_to_message_spec.py
+++ b/rosidl_parser/rosidl_parser/idl_to_message_spec.py
@@ -1,0 +1,157 @@
+from pathlib import Path
+from types import NoneType
+from typing import Any, List, Type
+
+from ament_index_python.packages import get_package_share_directory
+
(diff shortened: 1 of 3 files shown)


russkel:
Should this be moved to a separate package? I thought it would make sense to have it here.

PR #696, opened 2022-08-02
Move rosidl_cmake Python module to a new package rosidl_pycommon

jacobperron:
This decouples the common Python module from the package providing the cmake entrypoint.

Part of an effort to de-tangle the rosidl dependency graph and remove a dependency loop wrt to `action_msgs`.

PR #700, opened 2022-08-09
[service introspection] generate service_event messages

ihasdapie:
This PR is part of a prototype implementation for https://github.com/ros-infrastructure/rep/pull/360 https://github.com/ros2/ros2/issues/1285. 

 The rosidl generators are extended to generate a third message type for services, service events. Macro for generating interfaces in `rosidl_cmake` are also updated to `find_package` service_msgs introduced in https://github.com/ros2/rcl_interfaces/pull/143 and the new dependency on `service_msgs` by `action_msgs`.

Changes:
- Add <service_event>_Event message generation
- Add member methods to `rosidl_service_type_support_t` for building service messages and to provide the publisher typesupport handle.

This PR depends on https://github.com/ros2/rosidl/pull/696

Service introspection: https://github.com/ros2/ros2/issues/1285


rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -30,13 +32,18 @@ msvc_common_macros = ('DELETE', 'ERROR', 'NO_ERROR')
...
 for member in message.structure.members:
     type_ = member.type
     if isinstance(type_, AbstractNestedType):
         type_ = type_.value_type
     if isinstance(type_, NamespacedType):
+        if (
+            type_.name.endswith(SERVICE_REQUEST_MESSAGE_SUFFIX) or
+            type_.name.endswith(SERVICE_RESPONSE_MESSAGE_SUFFIX)
+        ):
+            continue
```
jacobperron, line 46: To assist with review, can you explain why we should check for service suffixes here and ignore them?

jacobperron (commented):
I think this PR can be rebased on `rolling` now.

rosidl_generator_cpp/resource/msg__struct.hpp.em
ihasdapie: Service suffixes are checked here in order to avoid generating `#include`s for generated service request, response, and event messages [`__trait`](https://github.com/ros2/rosidl/blob/46753d73dd1fc6e9208330ee8e6f340db72157de/rosidl_generator_cpp/resource/msg__traits.hpp.em#L35-L34)s and `__struct`s. This is because, unlike how a message that embeds another message defined "traditionally" i.e. creating a .msg file would need to `#include` the correspoinding traits/struct/etc, the service request/response/events already live together in the produced traits/struct/etc.

This is also how it is done with action messages in the same file.

jacobperron: It make sense that we have a special case for these service types, but why not handle them similar to what we're doing for actions. Instead of including nothing, we include the header that actually contains the declaration (e.g. at L52, we lop off the action suffix).

TBH, I'm not sure under what case we have a message that contains a nested action type, but I guess it doesn't hurt to allow it here.

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
```diff
@@ -137,22 +137,54 @@ macro(rosidl_generate_interfaces target)
...
+      ament_export_dependencies(service_msgs)
+      set(_action_msgs_added TRUE)
+      break()
+    elseif("${_parent_dir}" STREQUAL "srv" AND NOT _service_msgs_added)
+      list_append_unique(_ARG_DEPENDENCIES "service_msgs")
+      ament_export_dependencies(service_msgs)
+      set(_service_msgs_added TRUE)
+    endif()
+
+    if("${_parent_dir}" STREQUAL "action" AND NOT DEFINED action_msgs_FOUND)
```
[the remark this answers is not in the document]
ihasdapie: Should be resolved in https://github.com/ros2/rosidl/pull/700/commits/51071fbabf53c46d0d1f9be21c7bca4bf57a548f

rosidl_generator_cpp/resource/msg__struct.hpp.em
ihasdapie: Yeah, that'd make sense in the case of nested action or service types (which are still unsupported outside of injecting them into the rosidl pipeline as we're doing in this PR). But for currently supported usecases these `#include`s are referencing the file itself and it's a little circular (though harmless because of the header guards) For example `example_interfaces/AddTwoInts.srv` would create a `rosidl_generator_cpp/detail/add_two_ints__struct.hpp` which includes the `_Request`, `_Response`, and `Event` messages -- the declarations for which already live inside `rosidl_generator_cpp/detail/add_two_ints__struct.hpp`, and the same would apply to actions. I _think_ we could even just change that entire block to

```cpp
        if (
            type_.name.endswith(SERVICE_REQUEST_MESSAGE_SUFFIX) or
            type_.name.endswith(SERVICE_RESPONSE_MESSAGE_SUFFIX) or
            type_.name.endswith(ACTION_GOAL_SUFFIX) or
            type_.name.endswith(ACTION_RESULT_SUFFIX) or
            type_.name.endswith(ACTION_FEEDBACK_SUFFIX)
        ):
            continue
```

But if that nesting is something we want then it's just 

```suggestion
            typename = type_.name.rsplit('_', 1)[0]
```

Either should be fine

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
```diff
@@ -137,22 +137,50 @@ macro(rosidl_generate_interfaces target)
   # afterwards all remaining interface files are .idl files
   list(APPEND _idl_tuples ${_idl_adapter_tuples})
 
-  # to generate action interfaces, we need to depend on "action_msgs"
+  set(_service_msgs_added FALSE)
+  set(_action_msgs_added FALSE)
```
ivanpauno, line 141: instead of the new flags, can't you use `service_msgs_FOUND` and `action_msgs_FOUND` directly?

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
```diff
@@ -137,22 +137,50 @@ macro(rosidl_generate_interfaces target)
...
   foreach(_tuple ${_interface_tuples})
     string(REGEX REPLACE ".*:([^:]*)$" "\\1" _tuple_file "${_tuple}")
     get_filename_component(_parent_dir "${_tuple_file}" DIRECTORY)
-    if("${_parent_dir}" STREQUAL "action")
+
+    if("${_parent_dir}" STREQUAL "action" AND NOT _action_msgs_added)
```
ivanpauno, line 146:
```suggestion
    if(NOT _action_msgs_added AND "${_parent_dir}" STREQUAL "action")
```

may be more "efficient", as the STREQUAL part doesn't need to be evaluated if the flag is true

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
```diff
@@ -137,22 +137,50 @@ macro(rosidl_generate_interfaces target)
...
+        if(NOT ${service_msgs_FOUND})
+          message(FATAL_ERROR
+            "Unable to generate service interface for '${_tuple_file}'. "
+            "In order to generate service interfaces you must add a depend tag "
+            "for 'service_msgs' in your package.xml.")
+        endif()
+        ament_export_dependencies(service_msgs)
+        list_append_unique(_ARG_DEPENDENCIES "service_msgs")
+        set(_service_msgs_added TRUE)
+      endif()
```
ivanpauno, line 158: you don't need to duplicate this if you use in L171:

```cmake
    elseif(NOT _service_msgs_added AND "${_parent_dir}" MATCHES "srv|action")
```

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
```diff
@@ -137,22 +137,50 @@ macro(rosidl_generate_interfaces target)
...
+        message(FATAL_ERROR
+          "Unable to generate service interface for '${_tuple_file}'. "
+          "In order to generate service interfaces you must add a depend tag "
+          "for 'service_msgs' in your package.xml.")
+      endif()
+      ament_export_dependencies(service_msgs)
+      list_append_unique(_ARG_DEPENDENCIES "service_msgs")
+      set(_service_msgs_added TRUE)
     endif()
+
```
ivanpauno, line 183: you could add here an `if()` that breaks the loop if both service_msgs and action_msgs were already FOUND

rosidl_generator_tests/CMakeLists.txt
```diff
@@ -0,0 +1,113 @@
+cmake_minimum_required(VERSION 3.8)
+project(rosidl_generator_tests)
```
ivanpauno, line 2: why is this new package needed?

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -595,6 +596,20 @@ def __init__(
             namespaced_type.name + SERVICE_RESPONSE_MESSAGE_SUFFIX
         self.response_message = response
 
+        self.event_message = Message(
```
ivanpauno, line 599: shouldn't the event message be passed as an argument?
similar to request and response

rosidl_generator_cpp/resource/msg__struct.hpp.em
jacobperron: I see. The circular includes _are_ odd (I recall being confused because of them when looking at generated code). Maybe it is best to just ignore them like you suggest.

rosidl_cmake/rosidl_cmake/__init__.py
```diff
@@ -12,192 +12,11 @@
...
-        content = h.read()
-    try:
-        interpreter.string(content, str(template_path), kwargs)
-    except Exception as e:  # noqa: F841
-        print(f"{e.__class__.__name__} in template '{template_path}': {e}",
-              file=sys.stderr)
-        raise
-    interpreter.invoke('afterInclude')
+warnings.warn(
+    "The 'rosidl_cmake' Python module is deprecated. Use 'rosidl_pycommon' instead.",
```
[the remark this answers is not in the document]
ihasdapie: See: https://github.com/ros2/rosidl/pull/696

This was done to remove a dependency loop with action_msgs and service_msgs

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
ihasdapie: They aren't declared until the first `find_package` so it would be a little messy. But we could definitely do that if requested.

rosidl_generator_tests/CMakeLists.txt
ihasdapie: This removes a dependency loop between rosidl_cmake and the generators 

Reference: https://github.com/ros2/rosidl/pull/701

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
ivanpauno: you don't need to declare variables in cmake

```cmake
if (service_msgs_FOUND)
  message("A")
else()
  message("B")
endif()
```

will print `B`, if service_msgs_FOUND is undefined or it's set to a false constant.
From [cmake docs](https://cmake.org/cmake/help/latest/command/if.html#basic-expressions):

> if(\<variable\>)
True if given a variable that is defined to a value that is not a false constant. False otherwise, including if the variable is undefined. Note that macro arguments are not variables. [Environment Variables](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-environment-variables) also cannot be tested this way, e.g. if(ENV{some_var}) will always evaluate to false.

jacobperron:
rebased on rolling

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
jacobperron: I've refactored the loop, how's this? [91ea5a9](https://github.com/ros2/rosidl/pull/700/commits/91ea5a9062cecf44e8d77fc730c06ea89e922593)

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
jacobperron: See [91ea5a9](https://github.com/ros2/rosidl/pull/700/commits/91ea5a9062cecf44e8d77fc730c06ea89e922593)

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
jacobperron: I think this is not relevant with [91ea5a9](https://github.com/ros2/rosidl/pull/700/commits/91ea5a9062cecf44e8d77fc730c06ea89e922593)

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
jacobperron: [91ea5a9](https://github.com/ros2/rosidl/pull/700/commits/91ea5a9062cecf44e8d77fc730c06ea89e922593)

ivanpauno (approved):
LGTM!

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
```diff
@@ -137,21 +137,48 @@ macro(rosidl_generate_interfaces target)
   # afterwards all remaining interface files are .idl files
   list(APPEND _idl_tuples ${_idl_adapter_tuples})
 
-  # to generate action interfaces, we need to depend on "action_msgs"
+  # Check for any action or service interfaces
+  # Which have implicit dependencies that need to be found
   foreach(_tuple ${_interface_tuples})
+    # We use the parent directory name to identify if the interface is an action or service
     string(REGEX REPLACE ".*:([^:]*)$" "\\1" _tuple_file "${_tuple}")
     get_filename_component(_parent_dir "${_tuple_file}" DIRECTORY)
```
jacobperron, line 145: Turns out this logic for using the parent directory is quite brittle :roll_eyes: 

With the introduction of the `service_msgs` dependency, this PR has caused a regression in `rqt_py_common`, which doesn't try to `find_package(service_msgs)` because the parent directory is `test/srv` (not `srv`): https://github.com/ros-visualization/rqt/blob/d0234e0c3a96e6815e0a650e48fae97c7d877f92/rqt_py_common/CMakeLists.txt#L35

Now might be a good time to rethink this heuristic.

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
```diff
@@ -137,21 +137,48 @@ macro(rosidl_generate_interfaces target)
...
           "for 'action_msgs' in your package.xml.")
       endif()
-      list_append_unique(_ARG_DEPENDENCIES "action_msgs")
       ament_export_dependencies(action_msgs)
+      list_append_unique(_ARG_DEPENDENCIES "action_msgs")
+
+      # It is safe to break out of the loop since services only depend on service_msgs
+      # Which has already been found above
       break()
+    elseif(NOT service_msgs_FOUND AND "${_parent_dir}" STREQUAL "srv")
```
jacobperron: Ah, this is a bug. If the caller manually calls `find_package(service_msgs)` _before_ calling `rosidl_generate_interfaces`, then `service_msgs` will not be added to the list to the dependency list for linking against the target.

See [3168e55](https://github.com/ros2/rosidl/pull/700/commits/3168e55019442027a2dc99df93ac8c11a0e43486)
I just removed the use of `service_msgs_FOUND`. It's less efficient, but I don't think it should impact performance too badly.

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
jacobperron: See [4409291](https://github.com/ros2/rosidl/pull/700/commits/4409291176aada4e79ac249e60f6a386769e8418)
It proposes only relying on the immediate parent directory name for detecting action and service interfaces.

ivanpauno: that seems fine to me

ivanpauno (approved):
LGTM!
I have left some comments for myself to double check before merging.

rosidl_typesupport_introspection_c/resource/srv__type_support.c.em
```diff
@@ -58,10 +70,13 @@ static rosidl_service_type_support_t @(function_prefix)__@(service.namespaced_ty
```
ivanpauno, line 70: `service_introspection_message__create` and `service_introspection_message__destroy` need to be implemented and added to `rosidl_service_type_support_t` above to complete the implementation

ivanpauno (changes requested):
same for typesupport introspection cpp

rosidl_generator_cpp/resource/msg__struct.hpp.em
ivanpauno: This is to avoid recursive inclusion of the same file, it isn't a problem but it looks weird.
Same change needs to be applied to `rosidl_generator_c`.

rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
```diff
@@ -139,21 +139,49 @@ macro(rosidl_generate_interfaces target)
...
+      # Actions depend on the packages service_msgs and action_msgs
+      find_package(service_msgs QUIET)
+      if(NOT ${service_msgs_FOUND})
+        message(FATAL_ERROR
+          "Unable to generate action interface for '${_tuple_file}'. "
+          "In order to generate action interfaces you must add a depend tag "
+          "for 'service_msgs' in your package.xml.")
+      endif()
+      ament_export_dependencies(service_msgs)
+      list_append_unique(_ARG_DEPENDENCIES "service_msgs")
```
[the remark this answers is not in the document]
ivanpauno: This is not a good idea, see https://github.com/ros2/rosidl/pull/700#discussion_r986321213

rosidl_generator_cpp/resource/msg__struct.hpp.em
ivanpauno: I have done some modifications of this in https://github.com/ros2/rosidl/pull/700/commits/8e66f38e476d27ebc4cf5b1a9932e5620c010cb7

rosidl_typesupport_introspection_c/resource/srv__type_support.c.em
ivanpauno: Done in https://github.com/ros2/rosidl/pull/700/commits/f07ad3027ce2c1a854b938b81da7ecb311571f97

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17649/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12204/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18224/)

ivanpauno:
There's a bug with actions, I have to double check what's going on

ivanpauno:
Hopefully got it right now:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17668/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12217/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18237/)

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17669/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12218/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18238/)

ivanpauno:
windows error seems unrelated, though I don't see that happening in the nightlies ....
But there's no way these PR can make `mimick_vendor` fail.

There are also new linter failres, but those should be easy to fix ...

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17761/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12301/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18406/)

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17823/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12390/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18491/)

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17825/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12392/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18494/)

ivanpauno:
@wjwwood friendly ping.
Now there's green CI for:

- https://github.com/ros2/rosidl/pull/700
- https://github.com/ros2/rosidl_typesupport/pull/127
- https://github.com/ros2/rosidl_typesupport_fastrtps/pull/92
- https://github.com/ros2/rosidl_defaults/pull/24
- https://github.com/ros2/rcl_interfaces/pull/143

rosidl_typesupport_introspection_c/resource/srv__type_support.c.em
```diff
@@ -46,22 +57,42 @@ static rosidl_typesupport_introspection_c__ServiceMembers @(function_prefix)__@(
...
+    @(',\n    '.join(service.namespaced_type.namespaced_name()))
+  ),
+  ROSIDL_TYPESUPPORT_INTERFACE__SERVICE_DESTROY_EVENT_MESSAGE_SYMBOL_NAME(
+    rosidl_typesupport_c,
+    @(',\n    '.join(service.namespaced_type.namespaced_name()))
+  ),
+  &@(message_function_prefix)__@(service.namespaced_type.name)@(SERVICE_EVENT_MESSAGE_SUFFIX)_message_type_support_handle,
 };
+// rcl_interfaces__srv__GetParameters_Event__rosidl_typesupport_introspection_c__GetParameters_Event_message_type_support_handle
+// rcl_interfaces__srv__detail__get_parameters__rosidl_typesupport_introspection_c__GetParameters_Event_message_type_support_handle,
```
wjwwood, line 85: What are these comments for?

rosidl_runtime_c/include/rosidl_runtime_c/service_type_support_struct.h
```diff
@@ -29,6 +33,46 @@ typedef struct rosidl_service_type_support_t rosidl_service_type_support_t;
...
+  int64_t sequence_number;
+} rosidl_service_introspection_info_t;
+
+/// Creates a ServiceEvent message for the service.
+/**
+ * Instantiates a ServiceEvent message with the given info and request/response message.
+ * The message is allocated using the given allocator and must be deallocated using
+ * the rosidl_service_introspection_destroy_handle
+ *
+ * \param[in] POD fields of service_msgs/msg/ServiceEventInfo to be passed from rcl
```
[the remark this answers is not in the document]
ivanpauno: Fixed in https://github.com/ros2/rosidl/pull/700/commits/621facc30f273fc393f215706166b00b1ad812ed

rosidl_runtime_c/include/rosidl_runtime_c/service_type_support_struct.h
```diff
@@ -29,6 +33,46 @@ typedef struct rosidl_service_type_support_t rosidl_service_type_support_t;
...
+ * \param[in] allocator The allocator to use for allocating the ServiceEvent message
+ * \param[in] request_message type-erased handle to request message from rcl. Can be NULL.
+ * \param[in] response_message type-erased handle to request message from rcl. Can be NULL.
+ * \param[in] enable_message_payload Whether to enable message payloads in the ServiceEvent message
```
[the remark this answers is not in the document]
ivanpauno: Yes, it was removed.
Updated the docs in https://github.com/ros2/rosidl/pull/700/commits/621facc30f273fc393f215706166b00b1ad812ed.

rosidl_runtime_c/include/rosidl_runtime_c/service_type_support_struct.h
```diff
@@ -38,6 +82,12 @@ struct rosidl_service_type_support_t
   const void * data;
   /// Pointer to the service type support handler function
   rosidl_service_typesupport_handle_function func;
+  /// Pointer to function to create the introspection message
+  rosidl_event_message_create_handle event_message_create_handle;
```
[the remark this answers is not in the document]
ivanpauno: Corrected in https://github.com/ros2/rosidl/pull/700/commits/621facc30f273fc393f215706166b00b1ad812ed

rosidl_typesupport_introspection_c/resource/srv__type_support.c.em
ivanpauno: It was an old alternative that I commented out and I forgot to delete.
Fixed in https://github.com/ros2/rosidl/pull/700/commits/621facc30f273fc393f215706166b00b1ad812ed.

ivanpauno:
There's still [one comment](https://github.com/ros2/rcl_interfaces/pull/143#discussion_r1054821369) to be addressed, but running CI again to make sure the latest commits didn't introduce any regression:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17863/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12431/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18534/)

ivanpauno:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18541/)

ivanpauno:
Forgot to update branches correctly :man_facepalming:.
Windows CI again:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18548/)

mjcarroll (approved):

clalancette:
CI for this is in https://github.com/ros2/ros2/issues/1285#issuecomment-1404070715

PR #702, opened 2022-08-15
to support complete for rosidl and avoid calling multiple compinit

iuhilnehc-ynos:
Signed-off-by: Chen Lihui <lihui.chen@sony.com>

### DIFF
--- a/rosidl_cli/colcon.pkg
+++ b/rosidl_cli/colcon.pkg
@@ -1,0 +1,6 @@
+{
+    "hooks": [
+        "share/rosidl_cli/environment/rosidl-argcomplete.bash",
+        "share/rosidl_cli/environment/rosidl-argcomplete.zsh"
+    ]
+}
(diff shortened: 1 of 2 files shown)


iuhilnehc-ynos:
to fix https://github.com/ros2/ros2_documentation/issues/2944

mjbogusz:
Cross-ref https://github.com/ros2/ros2cli/issues/534

mjbogusz:
If I understand correctly, this solution relies on checking whether functions `compdef` and `complete` exists and assumes that if so, then they must have been called.
This MIGHT conflict with other systems/scripts that either utilize completion or define functions with such names. For example I have NVM installed (note: the nvm script merely calls `bashcompinit`):
```
$ type compdef
compdef is a shell function from /usr/share/zsh/functions/Completion/compinit
$ type complete
complete is a shell function from /home/mjbogusz/.nvm/bash_completion
```

Note however that this might not be a problem at all, I'm just documenting my observations.

PR #708, opened 2022-09-27
Add option to rosidl_generate_interfaces to skip dependency export.

clalancette:
When a dependency is declared in rosidl_generate_interfaces(), ament_export_dependencies() for that dependency is automatically called.  That is the correct default behavior, but it isn't always the behavior that is wanted.  In particular, if a package has messages that are only used for tests, we don't want to export those dependencies downstream.  Add an option to skip dependency export for rosidl_generate_interfaces.

This came out of https://github.com/ros-tooling/libstatistics_collector/pull/54 , and `libstatistics_collector` will be the first user of it.

### DIFF
--- a/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
+++ b/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake
@@ -47,14 +47,12 @@
 # :param ADD_LINTER_TESTS: if set lint the interface files using
 #   the ``ament_lint`` package
 # :type ADD_LINTER_TESTS: option
-# :param SKIP_EXPORT_DEPENDENCIES: if set don't ament export the dependencies listed
-# :type SKIP_EXPORT_DEPENDENCIES: option
 #
 # @public
 #
 macro(rosidl_generate_interfaces target)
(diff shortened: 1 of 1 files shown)


clalancette:
CI for this is over in https://github.com/ros-tooling/libstatistics_collector/pull/145#issuecomment-1261028442

sloretz (changes requested):
I think this should use the existing `SKIP_INSTALL` variable. It's already used to skip exporting targets, libraries, and dependencies in downstream generators.

https://github.com/ros2/rosidl_typesupport/blob/9fe935124b2a4f5c662f241850c03223855e8fc6/rosidl_typesupport_c/cmake/rosidl_typesupport_c_generate_interfaces.cmake#L140-L161

clalancette:
That's fair.  Made that change in 9013e83f176a4d7e66e288e0d8aa102597139417; please take another look.

sloretz (approved):

clalancette:
Latest CI for this is over in https://github.com/ros-tooling/libstatistics_collector/pull/145#issuecomment-1265545356 (and it is green).  So going ahead and merging this; thanks for the review!

PR #721, opened 2023-01-17
ZSH: fix argcomplete clearing previous registered completions

mjbogusz:
This is an alternative to https://github.com/ros2/rosidl/pull/702, up for reviews/discussion

More discussion in https://github.com/ros2/ros2cli/issues/534


mjbogusz:
As https://github.com/ros2/ros2cli/pull/750 was merged I'm removing the draft status.

PR #730, opened 2023-03-22
Return reference from BoundedVector::emplace_back

ahans:
Using `auto` made `BoundedVector::emplace_back()` return a copy instead of a reference. Code such as the following would not compile in that case:
```
auto & value = bounded_vector.emplace_back();
```
Due to the `auto` usage an rvalue was returned which couldn't be bound to an lvalue reference.

### DIFF
--- a/rosidl_runtime_cpp/include/rosidl_runtime_cpp/bounded_vector.hpp
+++ b/rosidl_runtime_cpp/include/rosidl_runtime_cpp/bounded_vector.hpp
@@ -474,7 +474,7 @@ public:
    * \param args Arguments to be forwarded to the constructor of Tp
    */
   template<typename ... Args>
-  auto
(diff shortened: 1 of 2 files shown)


clalancette (approved):
Yep, this seems reasonable to me.  It also gets rid of an `auto`, which I think is generally a good thing.  Thanks for the fix, will run CI on it next.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18285/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12812/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18953/)

clalancette:
The flaky tests on Windows are a known issue, so going ahead and merging.

PR #732, opened 2023-03-24
Copied type_description_interfaces structs (rep2011)

emersonknapp:
Part of ros2/ros2#1159

Reviewers please see [scripts/copy_type_description_generated_sources.bash]() which creates these copies - the manual changes in this review are
* `rosidl_runtime_c/CMakeLists.txt`
* `rosidl_runtime_c/docs/FEATURES.md`
* `rosidl_runtime_cpp/CMakeLists.txt`
* `rosidl_runtime_cpp/docs/FEATURES.md`
* `scripts/copy_type_description_generated_sources.bash`


scripts/copy_type_description_generated_sources.bash
```diff
@@ -0,0 +1,71 @@
...
+mkdir -p $C_SRC_DEST
+rm -f $C_SRC_DEST/*.c
+
+cp $C_DETAIL/*__struct.h $C_INCLUDE_DEST/
+cp $C_DETAIL/*__functions.h $C_INCLUDE_DEST/
+cp $C_DETAIL/*__functions.c $C_SRC_DEST/
+
+# add copy notice
+sed -i '1s/^/\/\/ DO NOT EDIT MANUALLY - this copied file managed by copy_type_description_generated_sources.bash\n/' $C_INCLUDE_DEST/*.h $C_SRC_DEST/*.c
+# remove unecessary includes (before doing replacements)
```
clalancette, line 36:
```suggestion
# remove unnecessary includes (before doing replacements)
```

scripts/copy_type_description_generated_sources.bash
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+set -euxo pipefail
+
+# Description:
+# Copy the structs defined in type_description_interfaces to rosidl_runtime_c and rosidl_runtime_cpp
+# so that they can be used in code generation to describe interface types
```
clalancette, line 6:
```suggestion
# so that they can be used in code generation to describe interface types.
```

Also, I think it is worthwhile to repeat here that this is necessary to break the loop between type_description_interfaces and the generators.

scripts/copy_type_description_generated_sources.bash
```diff
@@ -0,0 +1,71 @@
...
+sed -i '/#include "rosidl_runtime_c\/type_description\/type_description__struct.h/d' $C_INCLUDE_DEST/*.h $C_SRC_DEST/*.c
+# include guards
+sed -i -e 's/TYPE_DESCRIPTION_INTERFACES__MSG__DETAIL__/ROSIDL_RUNTIME_C__TYPE_DESCRIPTION__/g' $C_INCLUDE_DEST/*.h $C_SRC_DEST/*.c
+# visibility macros
+sed -i -e 's/ROSIDL_GENERATOR_C_PUBLIC_type_description_interfaces/ROSIDL_GENERATOR_C_PUBLIC/g' $C_INCLUDE_DEST/*.h $C_SRC_DEST/*.c
+# namespace prefixes
+sed -i -e 's/type_description_interfaces__msg__/rosidl_runtime_c__type_description__/g' $C_INCLUDE_DEST/*.h $C_SRC_DEST/*.c
+# includes
+sed -i -e 's/type_description_interfaces\/msg\/detail/rosidl_runtime_c\/type_description/g' $C_INCLUDE_DEST/*.h $C_SRC_DEST/*.c
+# sed -i -e 's/extern const rosidl_runtime_c__type_description__TypeDescription/\/\/ extern const rosidl_runtime_c__type_description__TypeDescription/' $C_INCLUDE_DEST/*.h $C_SRC_DEST/*.c
```
clalancette, line 47: Do we need this line?

scripts/copy_type_description_generated_sources.bash
```diff
@@ -0,0 +1,71 @@
...
+sed -i -e 's/namespace type_description_interfaces/namespace rosidl_runtime_cpp/g' *.hpp
+# msg namespace
+sed -i -e 's/namespace msg/namespace type_description/g' *.hpp
+# includes
+sed -i -e 's/type_description_interfaces\/msg\/detail/rosidl_runtime_cpp\/type_description/g' *.hpp
+# type references
+sed -i -e 's/type_description_interfaces::msg::/rosidl_runtime_cpp::type_description::/g' *.hpp
+# macros
+sed -i -e 's/type_description_interfaces__msg__/rosidl_runtime_cpp__type_description__/g' *.hpp
+# sed -i -e 's/static const rosidl_runtime_cpp::type_description::TypeDescription/\/\/ static const rosidl_runtime_cpp::type_description::TypeDescription/' *.hpp
```
clalancette, line 70: Do we need this line?

rosidl_runtime_c/docs/FEATURES.md
```diff
@@ -27,3 +27,16 @@ They are defined for the three main ROS 2 rosidl types:
 * Messages (`message_type_support_struct.h`)
 * Services (`service_type_support_struct.h`)
 * Actions (`action_type_support_struct.h`)
+
+## Type Description Interface Structs
+
+A copy of the output of `rosidl_generator_c` for `type_description_interfaces` messages.
```
clalancette, line 33: I think it is worthwhile to describe why this copy is here in the first place (to break the cycle).  Otherwise it might be hard for someone to understand why it is here.

rosidl_runtime_cpp/docs/FEATURES.md
```diff
@@ -25,3 +25,16 @@ These type traits are useful for providing compile-time checks of rosidl C++ typ
 
 Provided in `bounded_vector.hpp` is an implementation of a std::vector like container class, that has a bounded capacity.
 This class is utilized by rosidl types that declare a bounded sequence.
+
+## Type Description Interface Structs
+
+A copy of the output of `rosidl_generator_cpp` for `type_description_interfaces` messages.
```
clalancette, line 31: Same comment; we should add in why the copy is needed at all.

clalancette (changes requested):
Besides the changes inline, I'm wondering if we should add a git hook or github action to check if someone manually changes these files. I'd prefer the git hook (that way it will inform them locally when they are committing), but if we can't make that happen then a github action to check this would be nice.

scripts/copy_type_description_generated_sources.bash
emersonknapp: Well, we will when the type description code gen is in, but that's a followup PR and I can add the line when that change is actually made. I'll remove from this one.

rosidl_runtime_cpp/docs/FEATURES.md
emersonknapp: Done

emersonknapp:
@clalancette I'm thinking about how a hook could check on this. At the end of the day we don't know where the built code being copied actually came from so it can't be fully proofed against fools or ne'erdowells.

Maybe some "fingerprint" separate file that's also checked in - and that file can be just a big mean warning and one piece of information like the git commit of `type_description_interfaces` that the generated sources came from

```
# DO NOT EDIT MANUALLY - managed by scripts/copy_type_description_generated_sources.bash
# INTENTIONALLY CHANGING THIS FILE OUTSIDE THE SCRIPT 
# WILL RESULT IN UNDEFINED BEHAVIOR FOR ALL OF ROS 2 CORE
TYPE_DESCRIPTION_INTERFACES_GIT_COMMIT=deadbeef
```

And then the git hook could simply verify that this file is also changed alongside changes to any of the copied sources?

emersonknapp:
OK - I've checked in a github action that looks for changed files, and fails immediately if "fingerprint" is changed without sources, or vice versa. If both are changed, then it looks for the latest commit on `rcl_interfaces:rolling` and checks for equality on the value in the fingerprint. Given that this PR necessarily changes both the fingerprint and the sources, I've just tested the "not equal commit" case, and then pushed another commit with the correct fingerprint value to show. You can view https://github.com/ros2/rosidl/pull/732/checks?check_run_id=12319035002 vs https://github.com/ros2/rosidl/pull/732/checks?check_run_id=12319175638

clalancette (changes requested):

emersonknapp:
OK, second pass. I've changed the fingerprint file to be an output of hashes of the files that got copied. The builtin check here only checks that the fingerprint file changes when the copied sources do, but the big mean warning should be what warns folks off from changing it manually.

Cases testing:
* No protected things changed: https://github.com/emersonknapp/rosidl/pull/8
* Fingerprint changed, but not copied sources:  https://github.com/emersonknapp/rosidl/pull/6
* Sources changed, but not fingerprint: https://github.com/emersonknapp/rosidl/pull/7

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/583f883f05c5405351ac612c1c628d1f/raw/5d9dbe9f73ad23684d544bf0ace2955160cf2340/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11701
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18346/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12870/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19014/)

clalancette (approved):

PR #739, opened 2023-04-12
Add #ifndef NDEBUG to asserts

Acuadros95:
This PR disables assertions with undefined variables on Release mode:

- The `#ifndef NDEBUG` guards are messing with the build process with some assert implementations: 
https://github.com/ros2/rosidl/blob/7cbb1161e8d3fa9aff29a481bce0e6c2b56715b5/rosidl_runtime_c/src/type_description/field__description.c#L29-L37

- And the later assert, which should not use the variable if `NDEBUG` is defined:
https://github.com/ros2/rosidl/blob/7cbb1161e8d3fa9aff29a481bce0e6c2b56715b5/rosidl_runtime_c/src/type_description/field__description.c#L101

### A little explanation

This behaviour should be OK on all platforms, but some assert implementations still use the variable when they are dissabled, as its happening on our cross-compiled systems:
```
src/rosidl/rosidl_runtime_c/src/type_description/individual_type_description__description.c:104:25: error: 'rosidl_runtime_c__type_description__FieldType__EXPECTED_HASH' undeclared (first use in this function); did you mean 'rosidl_runtime_c__type_description__FieldType__TYPE_NAME'?
     assert(0 == memcmp(&rosidl_runtime_c__type_description__FieldType__EXPECTED_HASH, rosidl_runtime_c__type_description__FieldType__get_type_hash(NULL), sizeof(rosidl_type_hash_t)));
```

You can check what is happening on the assert implementation we are forces to use : [assert.h](https://github.com/espressif/esp-idf/blob/master/components/newlib/platform_include/assert.h).

If `NDEBUG` is defined, the assert macro still uses the variable and cast it to void, leading to the variable undefined error:
`#define assert(__e) ((void)(__e))`

### DIFF
--- a/rosidl_runtime_c/src/type_description/field__description.c
+++ b/rosidl_runtime_c/src/type_description/field__description.c
@@ -100,9 +100,9 @@ rosidl_runtime_c__type_description__Field__get_type_description(
   if (!constructed) {
     #ifndef NDEBUG
     assert(0 == memcmp(&rosidl_runtime_c__type_description__FieldType__EXPECTED_HASH, rosidl_runtime_c__type_description__FieldType__get_type_hash(NULL), sizeof(rosidl_type_hash_t)));
+    #endif
     description.referenced_type_descriptions.data[0].fields = rosidl_runtime_c__type_description__FieldType__get_type_description(NULL)->type_description.fields;
     constructed = true;
-    #endif
   }
   return &description;
 }
--- a/rosidl_runtime_c/src/type_description/field__description.c
+++ b/rosidl_runtime_c/src/type_description/field__description.c
@@ -98,9 +98,11 @@ rosidl_runtime_c__type_description__Field__get_type_description(
     {rosidl_runtime_c__type_description__Field__REFERENCED_TYPE_DESCRIPTIONS, 1, 1},
   };
   if (!constructed) {
+    #ifndef NDEBUG
     assert(0 == memcmp(&rosidl_runtime_c__type_description__FieldType__EXPECTED_HASH, rosidl_runtime_c__type_description__FieldType__get_type_hash(NULL), sizeof(rosidl_type_hash_t)));
     description.referenced_type_descriptions.data[0].fields = rosidl_runtime_c__type_description__FieldType__get_type_description(NULL)->type_description.fields;
     constructed = true;
+    #endif
   }
   return &description;
 }
--- a/rosidl_generator_c/resource/full__description.c.em
+++ b/rosidl_generator_c/resource/full__description.c.em
@@ -150,7 +150,9 @@ const rosidl_runtime_c__type_description__TypeDescription *
 c_typename = typename_to_c(ref_td['type_name'])
 }@
 @[    if ref_td['type_name'] not in full_type_names]@
+    #ifndef NDEBUG
     assert(0 == memcmp(&@(c_typename)__EXPECTED_HASH, @(c_typename)__@(GET_HASH_FUNC)(NULL), sizeof(rosidl_type_hash_t)));
+    #endif
 @[    end if]@
     description.referenced_type_descriptions.data[@(idx)].fields = @(c_typename)__@(GET_DESCRIPTION_FUNC)(NULL)->type_description.fields;
 @[  end for]@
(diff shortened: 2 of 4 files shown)


iuhilnehc-ynos:
I think this compilation issue should be reported at https://github.com/espressif/esp-idf/issues?
As https://en.wikipedia.org/wiki/Assert.h#cite_note-FOOTNOTEISO/IEC_JTC_1/SC_22/WG141999169-3 said, `If NDEBUG is defined as a macro name, the assert macro is defined as #define assert(ignore) ((void)0) `.

The `assert` is updated in https://github.com/espressif/esp-idf/commit/9afdf547481

```
-# define assert(__e) ((void)0)
+# define assert(__e) ((void)(__e))

-        assert(send_get_state(hal) == STATE_SENDING);
+        HAL_ASSERT(send_get_state(hal) == STATE_SENDING);
```

I think the `assert` macro should be reverted, and some logic needs to update.

Acuadros95:
> I think this compilation issue should be reported at https://github.com/espressif/esp-idf/issues?

Indeed, the implementation needs to be changed on their side. But without this changes we cannot offer compatibility with previous or the current release of this framework.

rosidl_runtime_c/src/type_description/field__description.c
```diff
@@ -98,9 +98,11 @@ rosidl_runtime_c__type_description__Field__get_type_description(
     {rosidl_runtime_c__type_description__Field__REFERENCED_TYPE_DESCRIPTIONS, 1, 1},
   };
   if (!constructed) {
+    #ifndef NDEBUG
     assert(0 == memcmp(&rosidl_runtime_c__type_description__FieldType__EXPECTED_HASH, rosidl_runtime_c__type_description__FieldType__get_type_hash(NULL), sizeof(rosidl_type_hash_t)));
     description.referenced_type_descriptions.data[0].fields = rosidl_runtime_c__type_description__FieldType__get_type_description(NULL)->type_description.fields;
     constructed = true;
+    #endif
```
iuhilnehc-ynos, line 105:
```suggestion
    #ifndef NDEBUG
    assert(0 == memcmp(&rosidl_runtime_c__type_description__FieldType__EXPECTED_HASH, rosidl_runtime_c__type_description__FieldType__get_type_hash(NULL), sizeof(rosidl_type_hash_t)));
    #endif
    description.referenced_type_descriptions.data[0].fields = rosidl_runtime_c__type_description__FieldType__get_type_description(NULL)->type_description.fields;
    constructed = true;
```

The same goes below.

iuhilnehc-ynos (commented):
It seems some `#endif` are not in the correct location.

rosidl_runtime_c/src/type_description/field__description.c
Acuadros95: My bad, fixed it here: https://github.com/ros2/rosidl/pull/739/commits/8a70ee16fd5d2a8e6d80a1264fba0b86bf6ac434.

Not sure why the CI is failing thought.

iuhilnehc-ynos: LGTM. Let's wait for others' opinions.

pablogs9:
Any update on this?

wjwwood:
@ros-pull-request-builder test this please

methylDragon:
Note that these files are copied from the generated files from rosidl

(i.e. the .msg files are used to generate files in the build space using rosidl, and then copied to this repository)

So it's kind of strange that the issues here don't pop up across all the message packages..

Still, if we do fix this, we need some documentation to remember to fix this again if we ever change the message build pipeline (since we'll need to regenerate the .msg files and copy them again, in which case the fixes in this PR will get overriden)

wjwwood:
Yeah, I'd like to understand why the generated version of this doesn't exhibit this issue first.

Also, if we do this, we should make an assert macro for this package, rather than wrapping all the asserts in the ifndef.

clalancette:
Just to follow-up further, another way to workaround this as an integration step is to define your own assert.h, redefining assert so it follows the standards.

sloretz:
It looks like this is working around a compiler bug for a compiler this repo doesn't officially support. I'll close this for now. Patching the toolchain to fix the bug, or replacing `assert.h` seems like a better path forward.

PR #703, opened 2022-09-16
Generate typesupport declarations for actions, messages and services

StefanFabian:
As discussed in ros2/rosidl_typesupport#114, this PR adds the generation of typesupport getter declarations that is currently missing.

For this, I have also added package-specific visibility control analogous to `rosidl_generator_c`.

Here's an example of the generated `fibonacci__type_support.hpp` from the `action_tutorials_interfaces` package:
```cpp
// generated from rosidl_generator_cpp/resource/idl__type_support.hpp.em
// with input from action_tutorials_interfaces:action/Fibonacci.idl
// generated code does not contain a copyright notice

#ifndef ACTION_TUTORIALS_INTERFACES__ACTION__DETAIL__FIBONACCI__TYPE_SUPPORT_HPP_
#define ACTION_TUTORIALS_INTERFACES__ACTION__DETAIL__FIBONACCI__TYPE_SUPPORT_HPP_

#include "rosidl_typesupport_interface/macros.h"

#include "action_tutorials_interfaces/msg/rosidl_generator_cpp__visibility_control.hpp"

#include "rosidl_typesupport_cpp/action_type_support.hpp"

#ifdef __cplusplus
extern "C"
{
#endif
// Forward declare the get type support functions for this type.
ROSIDL_GENERATOR_CPP_PUBLIC_action_tutorials_interfaces
const rosidl_action_type_support_t *
ROSIDL_TYPESUPPORT_INTERFACE__ACTION_SYMBOL_NAME(
  rosidl_typesupport_cpp,
  action_tutorials_interfaces,
  action,
  Fibonacci
)();
#ifdef __cplusplus
}
#endif

#include "rosidl_typesupport_cpp/message_type_support.hpp"

#ifdef __cplusplus
extern "C"
{
#endif
// Forward declare the get type support functions for this type.
ROSIDL_GENERATOR_CPP_PUBLIC_action_tutorials_interfaces
const rosidl_message_type_support_t *
ROSIDL_TYPESUPPORT_INTERFACE__MESSAGE_SYMBOL_NAME(
  rosidl_typesupport_cpp,
  action_tutorials_interfaces,
  action,
  Fibonacci_Goal
)();
#ifdef __cplusplus
}
#endif

// already included above
// #include "rosidl_typesupport_cpp/message_type_support.hpp"

#ifdef __cplusplus
extern "C"
{
#endif
// Forward declare the get type support functions for this type.
ROSIDL_GENERATOR_CPP_PUBLIC_action_tutorials_interfaces
[...]

PR #747, opened 2023-05-26
export for the event type support function

iuhilnehc-ynos:
There is no reason to ignore `event_message`.

Please see https://github.com/ros2/rosidl/blob/a63f32e7c425da1a8ea300227ac82fb7b3a0bd28/rosidl_generator_c/resource/srv__type_support.h.em#L19

### DIFF
--- a/rosidl_generator_cpp/resource/srv__type_support.hpp.em
+++ b/rosidl_generator_cpp/resource/srv__type_support.hpp.em
@@ -37,3 +37,9 @@ TEMPLATE(
     include_directives=include_directives)
 }@
 
+@{
+TEMPLATE(
+    'msg__type_support.hpp.em',
+    package_name=package_name, message=service.event_message,
+    include_directives=include_directives)
+}@


iuhilnehc-ynos:
`rosidl_runtime_cpp` only provides the template method to get the type support
https://github.com/ros2/rosidl/blob/034e526b3295108b7618cfd511660b44a76fc6b2/rosidl_runtime_cpp/include/rosidl_typesupport_cpp/message_type_support.hpp#L24-L25
which is different from `rosidl_runtime_c` using a macro to call the function
https://github.com/ros2/rosidl/blob/7583b95e13b9b0ca22f4f166da732ac73eace945/rosidl_runtime_c/include/rosidl_runtime_c/message_type_support_struct.h#L98-L100
, I wonder if it's necessary to add [the declaration](https://github.com/ros2/rosidl/blob/0c34a8e7bc45774bb5bcc0b1a262dbaa6867dc3d/rosidl_generator_cpp/resource/srv__type_support.hpp.em#L26-L38) for each message/service in the header file by `rosidl_generator_cpp`.

- e.g., build/example_interfaces/rosidl_generator_cpp/example_interfaces/srv/detail/add_two_ints__type_support.hpp

```cpp

#include "rosidl_typesupport_cpp/message_type_support.hpp"

#ifdef __cplusplus
extern "C"
{
#endif
// Forward declare the get type support functions for this type.
ROSIDL_GENERATOR_CPP_PUBLIC_example_interfaces
const rosidl_message_type_support_t *
  ROSIDL_TYPESUPPORT_INTERFACE__MESSAGE_SYMBOL_NAME(
  rosidl_typesupport_cpp,
  example_interfaces,
  srv,
  AddTwoInts_Request
)();
#ifdef __cplusplus
}
#endif

// already included above
// #include "rosidl_typesupport_cpp/message_type_support.hpp"

#ifdef __cplusplus
extern "C"
{
#endif
// Forward declare the get type support functions for this type.
ROSIDL_GENERATOR_CPP_PUBLIC_example_interfaces
const rosidl_message_type_support_t *
  ROSIDL_TYPESUPPORT_INTERFACE__MESSAGE_SYMBOL_NAME(
  rosidl_typesupport_cpp,
  example_interfaces,
  srv,
  AddTwoInts_Response
)();
#ifdef __cplusplus
}
#endif
```

As we know, the definition of messages
https://github.com/ros2/rosidl_typesupport/blob/9062d91c09a79c3b9902a8f06bca4e994121cc52/rosidl_typesupport_cpp/resource/srv__type_support.cpp.em#L3-L24
can be used for loading dynamically without header file,

https://github.com/ros2/rclcpp/blob/95adde2a19426c23d6ff6e13fab21d62e5e2ae69/rclcpp/src/rclcpp/typesupport_helpers.cpp#L124-L129
```cpp
    std::string symbol_name = typesupport_identifier + "__get_message_type_support_handle__" +
      package_name + "__" + (middle_module.empty() ? "msg" : middle_module) + "__" + type_name;

    const rosidl_message_type_support_t * (* get_ts)() = nullptr;
    // This will throw runtime_error if the symbol was not found.
    get_ts = reinterpret_cast<decltype(get_ts)>(library.get_symbol(symbol_name));
```

My question is,
do we need to add the declaration in the header file by `rosidl_generator_cpp` or just remove the declaration because it seems unnecessary and unuseful?

iuhilnehc-ynos:
I think it's not good to add `rosidl_typesupport_interface`
```
#define ROSIDL_GET_MSG_TYPE_SUPPORT_CPP(PkgName, MsgSubfolder, MsgName) \
  ROSIDL_TYPESUPPORT_INTERFACE__MESSAGE_SYMBOL_NAME( \
    rosidl_typesupport_cpp, PkgName, MsgSubfolder, MsgName)()
```
in `rosidl_runtime_cpp`, but it seems it's the follow-up PR of https://github.com/ros2/rosidl/pull/703.

fujitatomoya (approved):
I think this can be added along with response and request message types.

fujitatomoya:
> can be used for loading dynamically without the header file,
> do we need to add the declaration in the header file by rosidl_generator_cpp, or remove https://github.com/ros2/rosidl/pull/703 because it seems unnecessary and unuseful?

it is doable, but the difference is between compile time and runtime? i might be mistaken.

iuhilnehc-ynos:
> it is doable, but the difference is between compile time and runtime?

Yes, but no.

Let me take a generated service typesupport in cpp as an example,

- build/example_interfaces/rosidl_typesupport_cpp/example_interfaces/srv/add_two_ints__type_support.cpp

```cpp
namespace rosidl_typesupport_cpp
{

template<>
ROSIDL_TYPESUPPORT_CPP_PUBLIC
const rosidl_service_type_support_t *
get_service_type_support_handle<example_interfaces::srv::AddTwoInts>()
{
  return &::example_interfaces::srv::rosidl_typesupport_cpp::AddTwoInts_service_type_support_handle;
}

}  // namespace rosidl_typesupport_cpp

#ifdef __cplusplus
extern "C"
{
#endif

ROSIDL_TYPESUPPORT_CPP_PUBLIC
const rosidl_service_type_support_t *
ROSIDL_TYPESUPPORT_INTERFACE__SERVICE_SYMBOL_NAME(rosidl_typesupport_cpp, example_interfaces, srv, AddTwoInts)() {
  return ::rosidl_typesupport_cpp::get_service_type_support_handle<example_interfaces::srv::AddTwoInts>();
}

#ifdef __cplusplus
}
#endif
```

Is there anybody want to use `ROSIDL_TYPESUPPORT_INTERFACE__SERVICE_SYMBOL_NAME(rosidl_typesupport_cpp,...)` instead of `get_service_type_support_handle<>`?

Please notice:
1. C++ template is also built in the compile time 
2. the `ROSIDL_TYPESUPPORT_INTERFACE__SERVICE_SYMBOL_NAME(rosidl_typesupport_cpp,...)` is calling the template inside.
3. use in runtime (load dynamically) doesn't need the declaration in the header file

This is why I was confused that adding the declaration `ROSIDL_TYPESUPPORT_INTERFACE__SERVICE_SYMBOL_NAME(rosidl_typesupport_cpp,...)` in the header file?

PR #748, opened 2023-06-01
fix: Generate correct code for nested arrays of own type

jmachowinski:
Part of https://github.com/ros2/ros2/issues/1445

This fixes the code generation for the case that you have a message 
A :
A[] arrayOfA

### DIFF
--- a/rosidl_generator_c/resource/full__description.c.em
+++ b/rosidl_generator_c/resource/full__description.c.em
@@ -77,8 +77,16 @@ static const rosidl_type_hash_t @(c_typename)__EXPECTED_HASH = @(type_hash_to_c_
 
 @#<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
 @# Names for all types
+@{
+added_itype_names = set()
+}@
 @[for itype_description in all_type_descriptions]@
+@[  if itype_description['type_name'] not in added_itype_names]@
 static char @(typename_to_c(itype_description['type_name']))__TYPE_NAME[] = "@(itype_description['type_name'])";
+@{
+added_itype_names.add(itype_description['type_name'])
+}@
+@[  end if]@
 @[end for]@
 @#>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 
--- a/rosidl_generator_c/resource/msg__struct.h.em
+++ b/rosidl_generator_c/resource/msg__struct.h.em
@@ -151,6 +151,22 @@ enum
 @#>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 
 @#<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
+@# We must define the sequence type before the type itself
+@# for the special case, that the type contains a sequence
+@# of the type itself.
+// forward declare type
+struct @(idl_structure_type_to_c_typename(message.structure.namespaced_type));
+// Struct for a sequence of @(idl_structure_type_to_c_typename(message.structure.namespaced_type)).
+typedef struct @(idl_structure_type_sequence_to_c_typename(message.structure.namespaced_type))
+{
+  struct @(idl_structure_type_to_c_typename(message.structure.namespaced_type)) * data;
+  /// The number of valid items in data
+  size_t size;
+  /// The number of allocated items in data
+  size_t capacity;
+} @(idl_structure_type_sequence_to_c_typename(message.structure.namespaced_type));
+
+@#<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
 /// Struct defined in @(interface_path_to_string(interface_path)) in the package @(package_name).
 @{comments = message.structure.get_comment_lines()}@
 @[if comments]@
@@ -179,13 +195,3 @@ typedef struct @(idl_structure_type_to_c_typename(message.structure.namespaced_t
 } @(idl_structure_type_to_c_typename(message.structure.namespaced_type));
 @#>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 
-@#<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
-// Struct for a sequence of @(idl_structure_type_to_c_typename(message.structure.namespaced_type)).
-typedef struct @(idl_structure_type_sequence_to_c_typename(message.structure.namespaced_type))
-{
-  @(idl_structure_type_to_c_typename(message.structure.namespaced_type)) * data;
-  /// The number of valid items in data
-  size_t size;
-  /// The number of allocated items in data
-  size_t capacity;
-} @(idl_structure_type_sequence_to_c_typename(message.structure.namespaced_type));
--- a/rosidl_generator_type_description/rosidl_generator_type_description/__init__.py
+++ b/rosidl_generator_type_description/rosidl_generator_type_description/__init__.py
@@ -471,7 +471,7 @@ def calculate_type_hash(serialized_type_description):
         del field['default_value']
     for referenced_td in hashable_dict['referenced_type_descriptions']:
         for field in referenced_td['fields']:
-            del field['default_value']
+            field.pop('defaut_value', None)
 
     hashable_repr = json.dumps(
         hashable_dict,
--- a/rosidl_typesupport_introspection_cpp/resource/msg__type_support.cpp.em
+++ b/rosidl_typesupport_introspection_cpp/resource/msg__type_support.cpp.em
@@ -19,6 +19,9 @@ include_parts = [package_name] + list(interface_path.parents[0].parts) + [
     'detail', convert_camel_case_to_lower_case_underscore(interface_path.stem)]
 include_base = '/'.join(include_parts)
 
+message_namespace = '::'.join([package_name] + list(interface_path.parents[0].parts))
+message_namespaced_name = '::'.join([message_namespace, message.structure.namespaced_type.name])
+
 header_files = [
     'array',
     'cstddef',  # providing offsetof()
@@ -45,6 +48,19 @@ header_files = [
 @[    end if]@
 #include "@(header_file)"
 @[end for]@
+
+
+namespace rosidl_typesupport_introspection_cpp
+{
+
+// forward declare template, as it may be used in MessageMember
+template<>
+ROSIDL_TYPESUPPORT_INTROSPECTION_CPP_PUBLIC
+const rosidl_message_type_support_t *
+get_message_type_support_handle<@(message_namespaced_name)>();
+
+}  // namespace rosidl_typesupport_introspection_cpp
+
 @[for ns in message.structure.namespaced_type.namespaces]@
 
 namespace @(ns)
@@ -57,12 +73,12 @@ namespace rosidl_typesupport_introspection_cpp
 void @(message.structure.namespaced_type.name)_init_function(
   void * message_memory, rosidl_runtime_cpp::MessageInitialization _init)
 {
-  new (message_memory) @('::'.join([package_name] + list(interface_path.parents[0].parts) + [message.structure.namespaced_type.name]))(_init);
+  new (message_memory) @(message_namespaced_name)(_init);
 }
 
 void @(message.structure.namespaced_type.name)_fini_function(void * message_memory)
 {
-  auto typed_message = static_cast<@('::'.join([package_name] + list(interface_path.parents[0].parts) + [message.structure.namespaced_type.name])) *>(message_memory);
+  auto typed_message = static_cast<@(message_namespaced_name) *>(message_memory);
   typed_message->~@(message.structure.namespaced_type.name)();
 }
 
@@ -241,7 +257,7 @@ static const ::rosidl_typesupport_introspection_cpp::MessageMembers @(message.st
   "@('::'.join([package_name] + list(interface_path.parents[0].parts)))",  // message namespace
   "@(message.structure.namespaced_type.name)",  // message name
   @(len(message.structure.members)),  // number of fields
-  sizeof(@('::'.join([package_name] + list(interface_path.parents[0].parts) + [message.structure.namespaced_type.name]))),
+  sizeof(@(message_namespaced_name)),
   @(message.structure.namespaced_type.name)_message_member_array,  // message members
   @(message.structure.namespaced_type.name)_init_function,  // function to initialize message memory (memory has to be allocated)
   @(message.structure.namespaced_type.name)_fini_function  // function to terminate message instance (will not free memory)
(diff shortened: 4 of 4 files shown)


ros-discourse:

https://discourse.ros.org/t/use-ros-message-in-parameter-interface/31269/19

rosidl_typesupport_introspection_cpp/resource/msg__type_support.cpp.em
```diff
@@ -45,6 +45,18 @@ header_files = [
 @[    end if]@
 #include "@(header_file)"
 @[end for]@
+
+
+namespace rosidl_typesupport_introspection_cpp
+{
+
+template<>
```
emersonknapp, line 53: This is just a forward decl for a function defined lower down, right? If you could you add a comment to that effect? Is it not already declared in the `.hpp.em` header, and if so could we just include that instead? Why is the forward decl needed?

If we do need the forward declaration here - then I want to point out this expression

```
'::'.join([package_name] + list(interface_path.parents[0].parts))
```

Which is repeated many times in this file, and I think could be factored out into a variable at the top level block for easier reading and refactoring. Maybe factor out

```
message_namespace = '::'.join([package_name] + list(interface_path.parents[0].parts))
message_namespaced_name = '::'.join([message_namespace, message-structure.namespaced_type.name])
```

And that would really simplify some of the expressions farther down, make them more readable

rosidl_generator_c/resource/msg__struct.h.em
```diff
@@ -150,6 +150,19 @@ enum
 @[end if]@
 @#>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 
+@#<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
+// forward declare type
+struct @(idl_structure_type_to_c_typename(message.structure.namespaced_type));
+// Struct for a sequence of @(idl_structure_type_to_c_typename(message.structure.namespaced_type)).
+typedef struct @(idl_structure_type_sequence_to_c_typename(message.structure.namespaced_type))
```
emersonknapp, line 157: I think I understand that you are putting the Sequence struct before the single struct, so that the single struct could contain as members a Sequence of itself, correct?

If so, maybe add that to the comment for future readers. Could even put it in a template comment `@#` just so it's only for people modifying the template in the future

rosidl_generator_c/resource/full__description.c.em
```diff
@@ -77,8 +77,16 @@ static const rosidl_type_hash_t @(c_typename)__EXPECTED_HASH = @(type_hash_to_c_
 
 @#<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
 @# Names for all types
+@{
+added_itype_names = set()
+}@
 @[for itype_description in all_type_descriptions]@
+@[  if itype_description['type_name'] not in added_itype_names]@
```
emersonknapp, line 84: Did you hit a case where there was a multiple of a type in here? `all_type_descriptions` _should_ be a unique set to begin with

emersonknapp (commented):
Are there other PRs or issues that this is related to? Some cross-links between them would help understand the context for this change

jmachowinski:
> Are there other PRs or issues that this is related to? Some cross-links between them would help understand the context for this change

This commit would be the follow up. 
https://github.com/ros2/rcl_interfaces/commit/d9731c388ef6c55e5dc1d534af4e5575c9221f64

The goal here is to support complex parameters.

rosidl_generator_type_description/rosidl_generator_type_description/__init__.py
```diff
@@ -471,7 +471,8 @@ def calculate_type_hash(serialized_type_description):
         del field['default_value']
     for referenced_td in hashable_dict['referenced_type_descriptions']:
         for field in referenced_td['fields']:
-            del field['default_value']
+            if 'default_value' in field:
+                del field['default_value']
```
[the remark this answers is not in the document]
jmachowinski: Updated

rosidl_generator_c/resource/msg__struct.h.em
jmachowinski: Added a comment

rosidl_typesupport_introspection_cpp/resource/msg__type_support.cpp.em
jmachowinski: We are declaring that a special instance of the template function (for the given template argument) is present, and will be defined later. If we don't do this, we will get a compile error. (Usage of template function before instantiation)

I refactored the rest of the code as suggested.

rosidl_generator_c/resource/full__description.c.em
jmachowinski: Yes, I ran into this problem. 

//test gen rcl_interfaces/msg/ParameterValue {'type_name': 'rcl_interfaces/msg/ParameterValue', 'fields': [{'name': 'type', 'type': {'type_id': 3, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'bool_value', 'type': {'type_id': 15, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'integer_value', 'type': {'type_id': 8, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'double_value', 'type': {'type_id': 11, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'string_value', 'type': {'type_id': 17, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'byte_array_value', 'type': {'type_id': 160, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'bool_array_value', 'type': {'type_id': 159, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'integer_array_value', 'type': {'type_id': 152, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'double_array_value', 'type': {'type_id': 155, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'string_array_value', 'type': {'type_id': 161, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'value_array_value', 'type': {'type_id': 145, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': 'rcl_interfaces/msg/ParameterValue'}, 'default_value': ''}]}

static char rcl_interfaces__msg__ParameterValue__TYPE_NAME[] = "rcl_interfaces/msg/ParameterValue";
//test gen rcl_interfaces/msg/ParameterValue {'type_name': 'rcl_interfaces/msg/ParameterValue', 'fields': [{'name': 'type', 'type': {'type_id': 3, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'bool_value', 'type': {'type_id': 15, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'integer_value', 'type': {'type_id': 8, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'double_value', 'type': {'type_id': 11, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'string_value', 'type': {'type_id': 17, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'byte_array_value', 'type': {'type_id': 160, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'bool_array_value', 'type': {'type_id': 159, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'integer_array_value', 'type': {'type_id': 152, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'double_array_value', 'type': {'type_id': 155, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'string_array_value', 'type': {'type_id': 161, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': ''}, 'default_value': ''}, {'name': 'value_array_value', 'type': {'type_id': 145, 'capacity': 0, 'string_capacity': 0, 'nested_type_name': 'rcl_interfaces/msg/ParameterValue'}, 'default_value': ''}]}

static char rcl_interfaces__msg__ParameterValue__TYPE_NAME[] = "rcl_interfaces/msg/ParameterValue";

For an unknown reason, this is not an unique set....

emersonknapp: Oh, yeah - because you have 

```
# ParameterValue.msg
ParameterValue[] value_array_value
```

Then `toplevel_msg['referenced_type_descriptions']` contains as one of its members the `toplevel_msg['type_description']`. So, there can only be this one possible case where any type gets defined twice, and that's if it is self-referencing. I think this change is fine, given that. No need to add a special-case code when you can detect it generically like this.

emersonknapp:
I don't see any real problem in changing this code - but I'm curious if you've found other problems with introducing recursive message types. I imagine that this has unintended consequences in other parts of the codebase. Also, is supporting cyclic message definitions a change in the inherent design of ROS 2? Have you logged an issue somewhere to track the development of this feature? Probably https://github.com/ros2/ros2/issues would be a good place to track such a feature change.

jmachowinski:
> I don't see any real problem in this particular diff - but I'm curious if you've found other problems with introducing recursive message types. I imagine that this has unintended consequences in other parts of the codebase. 

Nothing until now, but I haven't really started the code building up on this yet, as it would be pointless if this merge request would not go through.

> Also, is supporting cyclic message definitions a change in the inherent design of ROS 2? 

Hm, interesting point, until now I just thought it was a weakness in the implementation, but not by design.

> Have you logged an issue somewhere to track the development of this feature? Probably https://github.com/ros2/ros2/issues would be a good place to track such a feature change.

I still haven't found out the 'correct' way how to contribute to ros... 
According to https://docs.ros.org/en/iron/The-ROS2-Project/Contributing.html
one should contact Open Robotics as early as possible, but sadly it is not stated how.
From the rest of documentation it looked like discourse is the way to go.
I will go save and add a discourse, and an issue and link them.

Thanks for the review by the way.

ros-discourse:

https://discourse.ros.org/t/cyclic-message-definitions/32075/1

emersonknapp:
>  If you open an issue, there is a hint, that general discussion should go to discourse.

"General discussion" should go to discourse, yes. But a specific feature/design that is being implemented via PRs, is what issue trackers are for. Sorry I stated it as a "probably" - given that this feature would span multiple repositories, then the issue should definitely go to ros2/ros2 metarepo. I've opened that issue here https://github.com/ros2/ros2/issues/1445. Maybe it turns out the limitation is by design, in which case we can close it, but for now that's going to be better than discourse for tracking work.

Also, just as a general practice - if something is labeled as "fix" such as this PR, then it makes it much easier to future historians/implementers if there is a detailed bug ticket that describes the problem, that the PR then fixes. Just a generalized process tip going forward, not saying you did anything wrong here.

jmachowinski:
So what is the way forward here ? We got no response at all for two weeks, judging from this, I would say it is not intentional that needed arrays do not work.

emersonknapp:
There was some offline conversation about this last week, but we're still uncertain as to the use case, whether this _should_ be supported. It was not specifically considered in the design either way originally, but seems like it opens a big can of complexity worms, the downstream effects are really uncertain.

This really seems like the sort of thing where a structured design proposal considering all effects and alternative approaches, highlighting what is actually trying to be accomplished would be the most valuable approach. The discussion is a bit scattered at this point, and I'm not seeing how self-referential nested types is helping to accomplish a real need that other approaches can't accomplish just as well.

Sorry for the slow progress - but this is unfortunately not a straightforward topic due to the potential downstream ramifications.

jmachowinski:
>The discussion is a bit scattered at this point, and I'm not seeing how self-referential nested types is helping to accomplish a real need that other approaches can't accomplish just as well.

My goal here is to have complex structured parameters. As rcl_interfaces::msg::ParameterDescriptor, rcl_interfaces::msg::ParameterValue etc are part of the rclcpp interface API, I can't extend the parameters without breaking the API.

Therefore the nicest way forward would be to use nested messages, as we would have full backward compatibility.

What are the potential downstream problems that you would expect ? (Only from this commit, not from the change to the parameters later on)

emersonknapp:
I think I'm going to need @clalancette @tfoote opinion on this. Is complex nested ParameterDescriptor something that ROS 2 should support? That downstream decision is what necessitates nested types as a general feature, and therefore whether this PR should move forward.

jmachowinski:
@emersonknapp  any progress ?

jmachowinski:
@emersonknapp ping ?

ros-discourse:

https://discourse.ros.org/t/simplifying-how-to-declare-parameters-in-ros-2/33272/5

jmachowinski:
@emersonknapp What is the status of this ? 
If possible, I would like to see this merged for jazzy if we decide to go ahead with it.

clalancette:
I'm going to retarget this one to `rolling`, and then we can discuss whether we actually want to take this.

ros-discourse:

https://discourse.ros.org/t/next-client-library-wg-meeting-friday-22nd-november-2024/40703/1

MichaelOrlov:
@jmachowinski I am a bit lost with the target use case and whether it is supposed to be a support for the recursive type declaration or just forward declaration in the same message definition file.

On today's PMC meeting you mentioned that you want to use it to define a list of ROS 2 parameters in the same ParameterValue.msg.
I am curious if it is supposed to be like just a forward declaration in the same file? like:
```
struct Foo; // Forward declaration
typedef sequence<Foo> FooSeq;
FooSeq ArrayOfFoo;
struct Foo {
  long value;
};
```
or recursive declaration like
```
struct Foo; // Forward declaration
typedef sequence<Foo> FooSeq;
struct Foo {
  long value;
FooSeq chain; // Recursive
};
```
Could you please clarify?

jmachowinski:
Note, for now we implemented our own parameter loading, 
but want to move back to 'pure' ros2 parameters in the long run.

A typical configuration in our system looks like this :
```
  lightbarriers:
    - pose:
        frame: conveyor_0/lightbarrier_0
        orientation:
          w: 0.707
          x: 0
          y: 0
          z: -0.707
        parent: conveyor_0
        position:
          x: 2.50
          y: 0.9
          z: 0.0555
      reflector_distance: 1.6
      update_frequency: 20
    - pose:
        frame: conveyor_0/lightbarrier_1
        orientation:
          w: 0.707
          x: 0
          y: 0
          z: -0.707
        parent: conveyor_0
        position:
          x: 2.25
          y: 0.9
          z: 0.0555
      reflector_distance: 1.6
      update_frequency: 20
``` 

Therefore we are interested in having support for arrays of complex nested structures.
**One** way of reaching this goal would be to modify the ParameterType and ParameterValue messages in this way : 

```
# These types correspond to the value that is set in the ParameterValue message.

# Default value, which implies this is not a valid parameter.
uint8 PARAMETER_NOT_SET=0

uint8 PARAMETER_BOOL=1
uint8 PARAMETER_INTEGER=2
uint8 PARAMETER_DOUBLE=3
uint8 PARAMETER_STRING=4
uint8 PARAMETER_BYTE_ARRAY=5

uint8 PARAMETER_ARRAY=10
uint8 PARAMETER_STRUCT
```

```
# Used to determine which of the next *_value fields are set.
# ParameterType.PARAMETER_NOT_SET indicates that the parameter was not set
# (if gotten) or is uninitialized.
# Values are enumerated in `ParameterType.msg`.

# The type of this parameter, which corresponds to the appropriate field below.
uint8 type

# "Variant" style storage of the parameter value. Only the value corresponding
# the type field will have valid information.

# Boolean value, can be either true or false.
bool bool_value

# Integer value ranging from -9,223,372,036,854,775,808 to
# 9,223,372,036,854,775,807.
int64 integer_value

# A double precision floating point value following IEEE 754.
float64 double_value

# A textual value with no practical length limit.
string string_value

# An array of bytes, used for non-textual information.
byte[] byte_array_value

# If type is PARAMETER_STRUCT, this is the name of the entry in the struct
string struct_entry_name

# If type is PARAMETER_STRUCT this is an array of struct members
# If type is PARAMETER_ARRAY this will be an array of parameters
ParameterValue[] value_array_value
``` 

@MichaelOrlov to answer your question, the current idea would be to use a recursive approach.

I would like to point out, that I am not settled on this approach. If anyone has other ideas on how to support parameters as shown above, I would love to hear them !

jmachowinski:
The goal of this PR was to support structured parameters.
It was now decided, that we want to implement this feature rather by stuffing a serialized blob (that can be recursive in itself) into a parameter msg than making msgs recursive themself. 
Therefore I am going to close this PR.

PR #754, opened 2023-06-15
Add rosidl_find_package_idl helper function

mikepurvis:
This is an initial effort meant to spur discussion— it's a first step in [making ament and ROS 2 friendlier to split-output building](https://discourse.ros.org/t/proposal-for-gnuinstalldirs-use-to-enable-multiple-installspaces/29182), which is ultimately accomplished by explicitly passing the values of the GNUInstallDirs variables into the configure step (there isn't really an equivalent for Python, since all of a Python package is required at runtime): 

```
cmake .. -DCMAKE_INSTALL_LIBDIR=$out -DCMAKE_INSTALL_DATADIR=$out -DCMAKE_INSTALL_INCLUDEDIR=$dev [....]
```

Part of the follow-on to that will be having ament (and catkin) generate variables like `fancy_package_DATADIR` and `fancy_package_LIBDIR` into their package find modules, and then working through the package tree to break the assumption that runtime files will always be located relative to the `find_package`-supplied `fancy_package_DIR` variable (which points to the location of the CMake module, in the dev tree— which in the Nix case can be disjoint from the runtime/out tree).

One of our biggest pain points around this is the finding of IDL files at configure time, since this is done by name, relative to the CMake module location, and occurs in many packages in a cut-and-pasted snippet that looks always like this:

```
foreach(_pkg_name ${rosidl_generate_interfaces_DEPENDENCY_PACKAGE_NAMES})
  foreach(_idl_file ${${_pkg_name}_IDL_FILES})
    set(_abs_idl_file "${${_pkg_name}_DIR}/../${_idl_file}")
    normalize_path(_abs_idl_file "${_abs_idl_file}")
    rosidl_find_package_idl(_abs_idl_file "${_pkg_name}" "${_idl_file}")
    list(APPEND _dependency_files "${_abs_idl_file}")
    list(APPEND _dependencies "${_pkg_name}:${_abs_idl_file}")
  endforeach()
endforeach()
```

This change seeks to centralize a small part of this blob into a `rosidl_cmake` helper function where it will be more convenient to iterate on the logic used in locating these things, ahead of some more ambitious changes to other parts of ament. In the immediate-term, if you don't want to include the `${${pkg_name}_DATADIR}/${idl_file}` part of this, at least adopting the framework makes it possible for us to patch this in in a single location rather than 7+ of them.

My initial questions are:

- Does the problem statement and overall approach seem to make sense here?
- Is `rosidl_cmake` the proper place for the suggested helper function?
- Should the helper be expanded in scope to include either or both of the surrounding loops, so basically just taking in the list of package names and spitting back `_dependencies` and `_dependency_files`?
- Is it likely we can merge something like this in the relatively near term, and then make companion changes to other repos (`rosidl_typesupport`, `rosidl_typesupport_fastrtps`, `rosidl_dds`, `rosidl_python`, etc) which adopt the new helper function?

FYI @iwanders @guillaumeautran @lopsided98 @gbiggs


rosidl_generator_c/cmake/rosidl_generator_c_generate_interfaces.cmake
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 find_package(rcutils REQUIRED)
+find_package(rosidl_cmake REQUIRED)
```
clalancette, line 16: Since we are adding this here, we should also add a dependency in `package.xml`.  I think that one needs to be a `<build_export_depend>`.

rosidl_generator_cpp/cmake/rosidl_generator_cpp_generate_interfaces.cmake
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+find_package(rosidl_cmake REQUIRED)
```
clalancette, line 15: Same comment here.

rosidl_typesupport_introspection_c/cmake/rosidl_typesupport_introspection_c_generate_interfaces.cmake
```diff
@@ -18,6 +18,7 @@ if(NOT TARGET ${rosidl_generate_interfaces_TARGET}__rosidl_generator_c)
     "'rosidl_typesupport_introspection_c' extension.")
 endif()
 
+find_package(rosidl_cmake REQUIRED)
```
clalancette, line 21: Same comment here.

rosidl_typesupport_introspection_cpp/cmake/rosidl_typesupport_introspection_cpp_generate_interfaces.cmake
```diff
@@ -18,6 +18,7 @@ if(NOT TARGET ${rosidl_generate_interfaces_TARGET}__rosidl_generator_cpp)
     "'rosidl_typesupport_introspection_cpp' extension.")
 endif()
 
+find_package(rosidl_cmake REQUIRED)
```
clalancette, line 21: Same comment here.

clalancette (changes requested):
Overall, I think this is an OK thing to do.  I've left some feedback inline on things to fix.

I do want to make one comment about the overall goal of switching to `GNUInstallDirs`.  After discussing with some other people, one thing that came up was that the switch to `GNUInstallDirs` will probably have knock-on effects, particularly on RHEL/Fedora.  We *think* that our infrastructure handles this, but we aren't sure about this aspect.  So there will need to be careful testing on both our default platforms (Ubuntu and Windows), as well as on RHEL.

rosidl_generator_c/cmake/rosidl_generator_c_generate_interfaces.cmake
mikepurvis: Fixed in a0dca0c6781a9c23618ee1383febccc37d957f95.

mikepurvis:
@clalancette Thanks for looking at this. I've added the missing dependencies; let me know if there's anything else I can do.

Regarding GNUInstallDirs, my impression on an initial survey is that it would _largely_ be a no-op for current use cases. Basically we'd be replacing what is currently hard-coded `lib`, `include`, etc with the `CMAKE_INSTALL_<thing>` variables, for example stuff like this:

https://github.com/ceres-solver/ceres-solver/pull/706/commits/4998c549396d36a491f1c0638fe57824a40bcb0d

But I'd be glad to participate in a discussion about what are some of the other barriers peoples people are foreseeing.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19332/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13860/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20051/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/20052/)
* RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/458/)

mikepurvis:
Fantastic! 🎉 

FYI @guillaumeautran

PR #755, opened 2023-06-19
Fixed visibility control file added to wrong header list variable.

StefanFabian:
See [#703](https://github.com/ros2/rosidl/pull/703#discussion_r1203804332)

### DIFF
--- a/rosidl_generator_cpp/cmake/rosidl_generator_cpp_generate_interfaces.cmake
+++ b/rosidl_generator_cpp/cmake/rosidl_generator_cpp_generate_interfaces.cmake
@@ -104,7 +104,7 @@ configure_file(
   "${_visibility_control_file}"
   @ONLY
 )
-list(APPEND _generated_msg_headers "${_visibility_control_file}")
+list(APPEND _generated_headers "${_visibility_control_file}")
 
 
 # INTERFACE libraries can't have file-level dependencies in CMake,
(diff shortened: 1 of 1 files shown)


rosidl_generator_cpp/cmake/rosidl_generator_cpp_generate_interfaces.cmake
```diff
@@ -104,7 +104,7 @@ configure_file(
   "${_visibility_control_file}"
   @ONLY
 )
-list(APPEND _generated_msg_headers "${_visibility_control_file}")
+list(APPEND _generated_headers "${_visibility_control_file}")
```
clalancette, line 107: Actually, looking at this I think this whole block is in the wrong location.  That is, the `add_custom_command` in the previous block is the one that defines the outputs from the Python command.  But since this block is added *after* that, this will never be part of `_generated_headers` and thus will not be tracked when something changes.

So I'd say that we need this fix, but we also need to move this entire block up above the `add_custom_commands`.

clalancette (changes requested):

rosidl_generator_cpp/cmake/rosidl_generator_cpp_generate_interfaces.cmake
StefanFabian: You're right. I've also moved it to the appropriate place (right after the other headers were generated).

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19007/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13532/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19761/)

calderpg-tri:
Regrettably, this change broke all of the prior work done to decouple IDL target generation from CMake by creating a new file that is only generated by CMake and not through one of the rosidl generators.

The generation of `rosidl_generator_cpp__visibility_control.hpp` should have be treated like `rosidl_generator_c__visibility_control.h` which is generated in `rosidl_generator_c`.

PR #756, opened 2023-06-30
Convert markdown to rosidl msg, srv or action content

zflat:
Adds a parser function to be used as part of the process to convert Markdown to `msg`, `srv` or `action` content. The conversion allows Markdown files to be treated like source files in a literate programming approach to IDL processing.

Most message files have comments providing documentation (for example [common_interfaces/Plane.msg](https://github.com/ros2/common_interfaces/blob/rolling/shape_msgs/msg/Plane.msg) ) and using a literate programming approach would allow for nice looking documentation when browsing source on github and gitlab hosted files. This would help with SEO and could help improve the quality of documentation. The readme at [common_interfaces/README.md](https://github.com/ros2/common_interfaces/blob/rolling/shape_msgs/README.md) already links to `.msg` files and the experience would be better if those files could also be rendered in HTML by github instead of as a raw text document. Even when viewing the raw files in a text editor, it will be easier to read documentation when it is not commented out with `#` at the start of each line.

Literate programming is to be achieved by parsing Markdown files for code blocks and separators used by `srv` and `action` messages, then concatenating the code blocks and separators to convert the input to `msg`, `srv` or `action` content as applicable. The conversion process is known as tangling in literate programming terms (for another example of literate programming see the guide [How to Use Emacs Org-Babel Mode to Write Literate Programming Document in R Language](https://orgmode.org/worg/org-contrib/babel/how-to-use-Org-Babel-for-R.html)).

Here is an example Markdown file that could be saved with the name `my_message.msg.md`. The `.msg.md` extension indicates that the file contains Markdown content that needs to be converted to `msg` content:

````markdown
# My Message
## Fields
* The `data` member holds an integer.
  ```
  std_msgs/msg/Int32 data
  ```
* The `str` member holds an optional string.
  ```
  std_msgs/msg/String str
  ```
## See also
The `result_val` in `MyService` also has the same semantic meaning as the `data` field in `MyMessage`.
```srv
# ...
---
# ...
std_msgs/msg/Int32 result_val
```
````

When the above `my_message.msg.md` is processed it is converted to create the following content which is then further processed as any other `msg` file content for message codegen.
```
std_msgs/msg/Int32 data
std_msgs/msg/String str
```
Note that the `srv` code block is not part of the converted output. Other code blocks like python or c++ could also be included in the Markdown and will not be part of the converted output.

### DIFF
--- a/rosidl_adapter/rosidl_adapter/parser.py
+++ b/rosidl_adapter/rosidl_adapter/parser.py
@@ -910,3 +910,31 @@ def parse_action_string(pkg_name, action_name, action_string):
 
     return ActionSpecification(
         pkg_name, action_name, goal_message, result_message, feedback_message)
+
+
+def tangle_markdown_to_rosidl(markdown_content, adapter_type):
+    if adapter_type == 'msg':
+        section_separator = None
+    elif adapter_type == 'srv':
+        section_separator = SERVICE_REQUEST_RESPONSE_SEPARATOR
+    elif adapter_type == 'action':
+        section_separator = ACTION_REQUEST_RESPONSE_SEPARATOR
+    tangle_yes_pattern = r'^```[^`]*tangle-yes'
+    output_lines = []
+    inside_codeblock = False
+    for line in [line.strip() for line in markdown_content.splitlines()]:
+        if inside_codeblock is True:
+            is_end_codefence = line == '```'
+            if is_end_codefence is True:
+                inside_codeblock = False
+            else:
+                output_lines.append(line)
+        else:
+            is_start_codefence = (line == '```' or re.search(tangle_yes_pattern, line))
+            if is_start_codefence is True:
+                inside_codeblock = True
+            else:
+                is_separator_line = (line == section_separator)
+                if is_separator_line is True:
+                    output_lines.append(section_separator)
+    return '\n'.join(output_lines)
--- a/rosidl_adapter/test/data/action/Test.action.md
+++ b/rosidl_adapter/test/data/action/Test.action.md
@@ -1,0 +1,38 @@
+# Foo
+## Goal
+```
+# goal definition
+# foo
+
+# bar
+bool bool_value
+# baz
+byte byte_value
+# asd
+char char_value  # bsd
+float32 float32_value
+float64 float64_value
+int8 int8_value
+uint8 uint8_value
+int16 int16_value
+uint16 uint16_value
+int32 int32_value
+uint32 uint32_value
+int64 int64_value
+uint64 uint64_value
+string string_value
+```
+---
+## Result and feedback
+```
+# result definition
+
+# ok docs
+bool ok
+---
+# feedback definition
+# more
+
+# sequence docs
+int32[] sequence
+```
--- a/rosidl_adapter/test/data/msg/Test.msg.md
+++ b/rosidl_adapter/test/data/msg/Test.msg.md
@@ -1,0 +1,42 @@
+# Test Msg
+
+Msg for testing parsing done by the `rosidl_adapter` package
+
+
+It is typical for a comment to describe the entire message as a whole and then to have a comment right above  single field to describe that field as follows:
+
+```
+# msg level doc
+
+# field level doc
+bool bool_value
+```
+
+The test message definition has multiple fields declared in varying types. The following field is commented on the same line as the field instead of the line right before the field.
+
+```
+byte byte_value  # field level doc, style 2
+```
+
+Some fields are commented with two different styles
+```
+# combined styles
+char char_value # combined styles, part 2
+```
+
+The remaining fields are not commented
+
+```
+float32 float32_value
+float64 float64_value
+int8 int8_value
+uint8 uint8_value
+int16 int16_value
+uint16 uint16_value
+int32 int32_value
+uint32 uint32_value
+int64 int64_value
+uint64 uint64_value
+```
+
+See the [test_tangle_markdown_to_rosidl](../../test_tangle_markdown_to_rosidl.py) test case.
--- a/rosidl_adapter/test/test_tangle_markdown_to_rosidl.py
+++ b/rosidl_adapter/test/test_tangle_markdown_to_rosidl.py
@@ -1,0 +1,44 @@
+# Copyright 2014 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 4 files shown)

### CODE REVIEW

File: rosidl_adapter/rosidl_adapter/parser.py [hunk 1]
```diff
@@ -910,3 +910,31 @@ def parse_action_string(pkg_name, action_name, action_string):
 
     return ActionSpecification(
         pkg_name, action_name, goal_message, result_message, feedback_message)
+
+
+def tangle_markdown_to_rosidl(markdown_content, adapter_type):
```
emersonknapp, line 915: Intentions aside: is there another linked PR somewhere else? Is this function used by anything, or is the intention just to put it in this library as a utility?

zflat: Thanks @emersonknapp for taking a look. The intention of this PR is to have a concrete implementation for allowing Markdown files to be treated like source files in a literate programming approach to IDL processing. With the changes in this branch it is possible to see with examples (test cases) just what is proposed.

These changes are not the full feature because the method `convert_to_idl` in the file `rosidl_adapter/rosidl_adapter/__init__.py` (https://github.com/ros2/rosidl/blob/c72b4545894ca99b58b73c5b8f3e232541353862/rosidl_adapter/rosidl_adapter/__init__.py#L16) will have to be updated to support the `*.msg.md`, `*.srv.md`, and `*.action.md` file extensions for example. However, I would not want to make more changes in this branch until the Markdown conversion has been more fully considered.

emersonknapp: I guess what I'm saying is I don't understand the basic goal/benefit of providing this utility. I understand from an abstract level what you mean by "literate programming" - meaning I've read the wikipedia page :wink:  - but I don't see the concrete benefit to providing this capability here in `rosidl_adapter`. If you have a specific need/desire for this approach in your applications, it should be perfectly possible to build a standalone `literate_markdown_adapter` (or whatever) that files can go through before being passed to `rosidl_adapter`.

I'm not the final word here, but that would be my suggestion to move forward with this.

clalancette: My personal opinion here is that we should not have this in the core.

I have two main reasons for this:
1.  Having support for too many formats effectively splits the community.  We already support two formats (`.msg` files and `.idl` files), and in order to look at all of ROS 2 you have to know both.  If we add a third format here, then you have to know all three formats.  The more formats we add, the harder it is for anyone to pick up a ROS 2 package and understand it.
2. This seems like something that can easily be done outside of the core.  That is, I can easily see a preprocessing step that takes the `.msg.md` files and converts them to `.msg`, and then invokes the rest of the ROSIDL pipeline (this is similar to @emersonknapp 's comment above).

But I'd like to hear other opinions from @ros2/team

sloretz: This PR has a side effect of the markdown files getting added to the ament index. I'm not sure that it would cause problems, but maybe [`ros2 interface show`](https://docs.ros.org/en/rolling/Tutorials/Beginner-CLI-Tools/Understanding-ROS2-Topics/Understanding-ROS2-Topics.html#ros2-interface-show) would show markdown on the command line?

> This seems like something that can easily be done outside of the core. That is, I can easily see a preprocessing step that takes the .msg.md files and converts them to .msg, and then invokes the rest of the ROSIDL pipeline (this is similar to @emersonknapp 's comment above).

This could be a macro in the build file that happens before `rosidl_generate_interfaces()`, though it might have to happen at configure time (via calling a script through `execute_process`) as I think the generators expect the list of interfaces to be known by then.

I think the documentation strategy for ROS 2 has moved towards packages using rosdoc2 to generate web pages. Markdown files in the repo would be nice for viewing them in Github, but maybe not for the generated package documentation. I'd recommend going the direction of turning comments on interface files into something pretty via rosdoc2.

zflat: Thanks for bringing up some good points. I think that the point about generating documentation is really important. If `rosdoc2` was further along then we would already have nice HTML to browse and that really is the aim at what I am trying to do. So yes, there are other ways to get good documentation, but I still think that literate programming has benefits and can get us there sooner and with tooling that will encourage developers to write good documentation.

I want to point out that in the `rosdoc2` readme, handling documentation of message files is still a TODO (see https://github.com/ros-infrastructure/rosdoc2/blob/47c4683dcd92063cc28bfae763ea64dfb96ada87/README.md?plain=1#L114). Since `rosdoc2` is mainly a wrapper around sphinx and doxygen, I think that it would make sense to use those tools and work with what they support. For example, sphinx can handle markdown with https://www.sphinx-doc.org/en/master/usage/markdown.html. If we want to go with the sphinx default, this branch can be changed to work with [reStructuredText](https://www.sphinx-doc.org/en/master/usage/restructuredtext/index.html) instead. So, there is a gap right now in what `rosdoc2` can do and the approach I am proposing can help to bridge that gap and get to better documentation.

Regarding the point about too many formats, I want to point out that launch files can be python, XML and YAML. There are reasons to use python versus XML and likewise I can see reasons to use the different formats for interfaces definitions. Additionally, the syntax for `.msg` `.action` and `.srv` files is very basic so we are not asking people to learn a new programming language or even deal with strange concepts like YAML anchors. Also if the main concern is people being able to pick up a ROS package and understand it, as @clalancette brings up, then really good documentation is only going to help.

I think that everybody here understands the benefits of using markdown, but I will restate that being able to use bullet lists, include diagrams and links to other documents and other features that markdown supports will make the documentation much more rich. Take a look at https://github.com/ros2/common_interfaces/blob/rolling/sensor_msgs/msg/CameraInfo.msg for example. The poor authors are trying really hard to provide good documentation but it is still pretty hard on the eyes to read. I think that parsing the `CameraInfo.msg` comments and generating nice HTML will be a big effort to add to `rosdoc2` but converting it to a markdown file `CameraInfo.msg.md` and then using what sphinx already supports will be much easier and will provide the authors exactly what they need to describe their messages right out of the box. Supporting math formulas would be in reach when using markdown, but a real rabbit-hole to pursue if a new markup language is invented and embedded into comment blocks in `.msg` files.

Overall, tooling should be easy to use and encourage adding more description and not just the bare minimum. With good tooling, a developer would want to add more detailed information and make edits to improve what has been written. Markdown can be previewed while editing, so again, we would take advantage of existing tooling to have a much better developer experience.

I think that turning comments on interface files into something pretty via `rosdoc2` could be explored more as @sloretz recommends. I acknowledge that there are drawbacks of introducing another interface file type that uses this literate programming approach, but I think the benefits will make it a better choice than processing comments in  interface files.

sloretz:
Sorry for the long delay. I don't think this is a feature we want to add to this repo, so I'll close this for now. I'd recommend making a separate package that translates markdown to `.msg` and then feeding that output into `rosidl_generate_interfaces()` instead.

PR #759, opened 2023-07-21
Fix same named types overriding typesources

emersonknapp:
When two interface types in the same package had the same name stem (e.g. `msg/Empty.msg` and `srv/Empty.srv`), both types would receive the same `TypeSource`s in their generated description. Can be considered undefined behavior, but it's processed alphabetically so the `msg` consistently received the `srv` sources.

This fixes it, adds a regression test, and adds a new basic test for `.idl` files just to make sure that continues to work OK.
Bug exists in Iron, this would need to be backported.

### DIFF
--- a/rosidl_generator_tests/CMakeLists.txt
+++ b/rosidl_generator_tests/CMakeLists.txt
@@ -34,6 +34,7 @@ if(BUILD_TESTING)
     ${test_interface_files_ACTION_FILES}
     ${test_interface_files_MSG_FILES}
     ${test_interface_files_SRV_FILES}
+    msg/BasicIdl.idl
     ADD_LINTER_TESTS
     SKIP_INSTALL
   )
--- a/rosidl_pycommon/rosidl_pycommon/__init__.py
+++ b/rosidl_pycommon/rosidl_pycommon/__init__.py
@@ -70,7 +70,9 @@ def generate_files(
     ros_interface_files = {}
     for ros_interface_file in args.get('ros_interface_files',  []):
         p = pathlib.Path(ros_interface_file)
-        ros_interface_files[p.stem] = p
+        # e.g. ('Empty', 'msg')
+        key = (p.stem, p.suffix[1:])
+        ros_interface_files[key] = p
 
     for idl_tuple in args.get('idl_tuples', []):
         idl_parts = idl_tuple.rsplit(':', 1)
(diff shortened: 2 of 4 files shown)


rosidl_pycommon/rosidl_pycommon/__init__.py
```diff
@@ -70,7 +70,9 @@ def generate_files(
     ros_interface_files = {}
     for ros_interface_file in args.get('ros_interface_files',  []):
         p = pathlib.Path(ros_interface_file)
-        ros_interface_files[p.stem] = p
+        # e.g. ('Empty', 'msg')
+        key = (p.stem, p.suffix[1:])
```
clalancette, line 74: This is very nit-picky, but we are used to thinking about things in terms of `pkg/type/name`.  So this tuple being (`name`/`type`) seems a little odd.  Should we invert the order?  If we do this, we'll also have to fix `type_source_key` below.

rosidl_generator_tests/CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ if(BUILD_TESTING)
     ${test_interface_files_ACTION_FILES}
     ${test_interface_files_MSG_FILES}
     ${test_interface_files_SRV_FILES}
+    msg/BasicIdl.idl
```
clalancette, line 37: Just to be totally clear, this is an unrelated test, correct?  I'm actually fine with adding it in here for testing purposes, but I just want to make sure I'm not missing something.

clalancette (changes requested):

rosidl_generator_tests/CMakeLists.txt
emersonknapp: I just wanted to make extra sure, here in `rosidl` repo, that the new changes didn't break anything for `.idl` files, which are handled slightly differently as they don't go through `rosidl_adapter`. The intention for this extra file is to serve that purpose, since there were not yet any `.idl` files being tested by this package.

rosidl_pycommon/rosidl_pycommon/__init__.py
emersonknapp: No problem, I'll invert it. I had written it both ways and just gone this way, but no strong opinion on it

clalancette (approved):
Looks good with green CI.

emersonknapp:
Pulls: ros2/rosidl#759
Gist: https://gist.githubusercontent.com/emersonknapp/67c942266b93571ebb90a33aef5d2ddc/raw/a35bc75291715af8472773be5c927306a1474c62/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12433
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19233/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13788/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19970/)

emersonknapp:
@mergifyio backport iron

PR #766, opened 2023-09-04
Fixed C++20 warning implicit capture of this in lambda

AiVerisimilitude:


AiVerisimilitude:
Let me know if you prefer this or for me to open a PR that simply does [ =, this] that gets merged when it's time for C++20. 

The advantage of using the approach of this PR is that it compiles on both C++17 & C++20 without warnings, but I'm not so happy about it.

clalancette:
> The advantage of using the approach of this PR is that it compiles on both C++17 & C++20 without warnings, but I'm not so happy about it.

For the foreseeable future, I think we are going to have to maintain compatibility with C++17.  Even if we were to make the core C++20 clean, we are unlikely to force that on all downstream packages for Jazzy.  Granted that this is just a test, but I'd still prefer not to have PRs hanging around for a long time waiting for a change to C++20.

All of that is to say that I think a slightly uglier approach is fine to maintain C++17 compatibility.  I'll take a look at this one and run CI on it.

rosidl_typesupport_introspection_tests/include/rosidl_typesupport_introspection_tests/gtest/message_introspection_test.hpp
```diff
@@ -76,12 +76,14 @@ class MessageIntrospectionTest : public SharedLibraryTest
   MakeTypeErasedMessage() const
   {
     std::allocator<MessageT> allocator;
-    std::function<void(void *)> type_erased_message_deleter = [ = ](void * ptr) mutable {
-        finalize_message(ptr, message_descriptor_);
+    auto message_descriptor = GetMessageDescriptor();
+    std::function<void(void *)> type_erased_message_deleter =
+      [allocator, message_descriptor](void * ptr) mutable {
```
clalancette, line 81: I'm certainly not an expert on C++20, but is there some reason we can't do something like:

```
    std::function<void(void *)> type_erased_message_deleter =
      [this, allocator](void * ptr) mutable {
```

?  `message_descriptor_` is part of the class, so capturing `this` should be enough, no?

AiVerisimilitude: Yep, for some reason my brain assumed a similar warning would be thrown, so I went full on explicit declaration, but that works out, and it's a lot cleaner. Sold! Thanks for the review :)

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19457/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13980/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20180/)

PR #768, opened 2023-09-11
Fix zero-variadic-macro-arguments warning when using clang.

clalancette:
When building this package with clang, we get a lot of warnings like the following:

must specify at least one argument for '...' parameter of variadic macro

That comes mostly from the TYPED_TEST_SUITE macros in gtest, the signature of which looks like:

Prior to C++20, C++ technically did not allow macros with variadic arguments to be called with nothing for the last variadic part. This is allowed in C++20, but we aren't ready to transition to it yet.

So just add in an unused third "empty" argument anytime we use these variadic arguments.  This will quiet the warning on clang and have no effect on operation.

### DIFF
--- a/rosidl_typesupport_introspection_tests/test/test_arrays_message_introspection.cpp
+++ b/rosidl_typesupport_introspection_tests/test/test_arrays_message_introspection.cpp
@@ -38,7 +38,7 @@ class ArraysMessageIntrospectionTest
 using ArraysMessageTypes = ::testing::Types<
   rosidl_typesupport_introspection_tests__msg__Arrays,
   rosidl_typesupport_introspection_tests::msg::Arrays>;
-TYPED_TEST_SUITE(ArraysMessageIntrospectionTest, ArraysMessageTypes);
+TYPED_TEST_SUITE(ArraysMessageIntrospectionTest, ArraysMessageTypes, );
 
 // NOTE(hidmic): cppcheck complains about gtest macros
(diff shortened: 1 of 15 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19500/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14020/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20219/)

emersonknapp (approved):
🤷

clalancette:
So `cppcheck` on Windows is unhappy with this change; it thinks it is a "syntax error".  We only see this on Windows since we effectively disabled `cppcheck` on Linux (the performance with version 2 of cppcheck, which ships in Ubuntu 22.04, is just terrible).

There are two paths forward here:
1.  I could disable cppcheck on Windows for this package.  Someday we are going to disable it for the core entirely anyway.
2.  Instead of this patch, I could just disable the warning when using clang.  It is kind of a silly error anyway, and since this is only effecting tests I would not feel bad about it.

I think I'm going to go for 2; this will stop being a warning at all in C++20, so I really don't see a downside to doing it.

mjcarroll (approved):
After some offline deliberating, I'm good with this flag.  It's an artifact of `-Wpedantic` and `clang` and will be resolved in C++20.

Since this is all test code, there is no potential of this leaking to other packages.

PR #772, opened 2023-10-10
Fix constant generation for C++ floats

clalancette:
While working on a package, I came across a build warning on Windows that said:

```
warning C4305: 'initializing': truncation from 'double' to 'const float'
```

After looking into it, I realized that the problem was that when we generate float constants for C++, the value we assign to it is a double, not a float.  For instance, given the [Constants.msg](https://github.com/ros2/test_interface_files/blob/rolling/msg/Constants.msg) file, we currently generate (truncated for brevity):

```
  // constant declarations
  static constexpr float FLOAT32_CONST =
    1.125;
  static constexpr double FLOAT64_CONST =
    1.125;
```

But that is not quite right; that first line should be:

```
  static constexpr float FLOAT32_CONST =
    1.125f;
```

Interestingly, the warning only shows up when assigning certain constants to the float, like 0.05.

So this PR does three things:
1.  Does a small refactoring of the tests, removing bits we don't need.
2.  Adds a test for the problematic constants (by default, the warning will only show up on Windows).
3.  Fixes the code generation to append the `f` as appropriate.

### DIFF
--- a/rosidl_generator_cpp/resource/msg__struct.hpp.em
+++ b/rosidl_generator_cpp/resource/msg__struct.hpp.em
@@ -280,10 +280,16 @@ non_defaulted_zero_initialized_members = [
   static const @(MSG_TYPE_TO_CPP['wstring']) @(constant.name);
 @[ else]@
   static constexpr @(MSG_TYPE_TO_CPP[constant.type.typename]) @(constant.name) =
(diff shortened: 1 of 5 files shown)


mjcarroll (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19642/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14150/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20366/)

Issue #775, opened 2023-11-15
`char` type in Python messages is of type uint8_t not str.


arjo129:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04 and windows 10
- Installation type:
  - Binaries
- Version or commit hash:
  - Humble
- DDS implementation:
  - FastDDS
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue

The [documentation](https://docs.ros.org/en/humble/Concepts/Basic/About-Interfaces.html) page states that `char`s should corresponf to `builtins.str` but when I try to set a field of type `char` to a string in python like so: 
```
my_awesome_msg.char_field = "L"
```
I get the error
```
AssertionError: The 'my_awesome_msg.char_field' field must be of type 'int'
```

Is the documentation wrong or is this unexepected behaviour?

I am willing to submit a fix for this.

sloretz:
Thank you for the bug report! It seems like you're correct that the type should be a `str` field.

* ROS msg files to OMG IDL maps char to char: https://design.ros2.org/articles/mapping_dds_types.html
* OMG IDL char maps to python string with length 1: https://design.ros2.org/articles/idl_interface_definition.html


It seems like the bug is in `rosidl_adapter`. It maps ROS msg char to OMG IDL uint8: https://github.com/ros2/rosidl/blob/5d8191e07627a8c394e42690d6a090d2f63c016a/rosidl_adapter/rosidl_adapter/msg/__init__.py#L46 . 

Would you be willing to open a PR with a fix on the rolling branch? I don't think we could backport it, but we could fix it going forward.

arjo129:
I think updating the documentation is sufficient. Changing this is not worth the effort as users who use this bug will likely have to go through  a very painful process of updating their code base for very little value-add.

PR #781, opened 2024-01-04
Address #775: Really make `char` in python a `str`

arjo129:
This commit addresses #775 for rolling going forward.

Note: This commit breaks a lot of down stream tests. I will be pushing fixes one by one as I discover errors.


sloretz:
The fix here looks good to me! Let me know when they're ready and I'll review them all at once.

clalancette:
While I think this is the right thing to do, this is probably going to cause some heartache on downstream consumers.

To that end, I think we should be loud about the change.  So I'll request that you please open up a PR to update http://docs.ros.org/en/rolling/Releases/Release-Jazzy-Jalisco.html with details of this change so that users who are switching to Jazzy are aware.

arjo129:
I don't think this change is worth it as downstream users will be left with a lot of "busy work"

PR #791, opened 2024-03-20
Add types to definition.py in `rosidl_parser`

InvincibleRMC:
Adding types to defintion.py to solve warnings raised in https://github.com/ros2/rosidl_python/pull/206 and improve readability.


fujitatomoya (approved):
lgtm with green CI. I would like to have 2nd review.

rosidl_parser/rosidl_parser/definition.py
```diff
@@ -742,31 +803,38 @@ class IdlLocator:
...
+    def __init__(self) -> None:
         super().__init__()
-        self.elements = []
+        self.elements: List[IdlContentElement] = []
 
-    def get_elements_of_type(self, type_):
+    def get_elements_of_type(
+            self,
+            type_: Type[IdlContentElementT]
+    ) -> List[IdlContentElementT]:
```
clalancette, line 837: I could be wrong, but shouldn't this be:

```suggestion
    ) -> List[IdlContentElement]:
```

clalancette (commented):
I've left one comment with a possible problem; otherwise this looks good to me.

rosidl_parser/rosidl_parser/definition.py
InvincibleRMC: Due to the `isinstance` check we know that they have to match `type_` so using IdlContentElementT is correct.

```python
return [e for e in self.elements if isinstance(e, type_)]
```

This make it so when you use `get_elements_of_type` it knows that returned List only has objects of type `Action`

```python
foo = IdlContent().get_elements_of_type(Action)
reveal_type(foo)  # List[Action]
```

clalancette: I guess I don't understand then.  This method is essentially returning a subset of `self.elements`, which is typed as `List[IdlContentElement]`.  Thus it seems to me that the type of `self.elements` and the return type of `get_elements_of_type` should be the same, no?

InvincibleRMC: Ok so in line 821 is a Union of all possible entries into the `elements` list. I got these from the `extract_content_from_ast` in ros_idl.parser. These are all the possible values that can go into the `elements` list.

InvincibleRMC: Line 822 is a TypeVar that is upper bound by the TypeAlias from 821.  Because of the `isinstance` check inside the list interpolation we know that the returned List will be a subset of `self.elements` which all match `type_`. Since we know the the list in a subset of only `type_` objects we can make the function generic over `type_` and have the list match the type of `type_`. I'm sorry if my explanations are not super clear, for more information check out mypy's [page](https://mypy.readthedocs.io/en/stable/generics.html#generic-functions) about generic functions.

sloretz (approved):

sloretz:
Pulls: https://github.com/ros2/rosidl/pull/791
Gist: https://gist.githubusercontent.com/sloretz/2c0fe5761de1071f7ded2b57f31a82d9/raw/c5235f4fdd81c7d9d5e27a042385ff870be0cee4/ros2.repos
BUILD args:  --packages-above-and-dependencies rosidl_parser
TEST args:  --packages-above rosidl_parser
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14723
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22021/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16277/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1539/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22751/)

sloretz:
Thank you for the PR!

PR #793, opened 2024-03-26
Fixes for modern uncrustify.

clalancette:
Mostly just redoing macros to make them fit on one line so we can avoid problems between old and new uncrustify.

### DIFF
--- a/rosidl_generator_tests/test/rosidl_generator_c/test_interfaces.c
+++ b/rosidl_generator_tests/test/rosidl_generator_c/test_interfaces.c
@@ -45,21 +45,16 @@
   u"Deep into that darkness peering, long I stood there wondering, fearing \u2122"
 #define ARR_SIZE 3
 
-#define STRINGIFY(x) _STRINGIFY(x)
 #define _STRINGIFY(x) #x
+#define STRINGIFY(x) _STRINGIFY(x)
 
-#define EXPECT_FALSE(arg) if (arg) { \
-    fputs(STRINGIFY(arg) " is not false\n", stderr); \
-    return 1; \
-}
-#define EXPECT_TRUE(arg) if (!(arg)) { \
-    fputs(STRINGIFY(arg) " is not true\n", stderr); \
-    return 1; \
-}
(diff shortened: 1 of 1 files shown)


ahcorde (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20611/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14964/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21332/)

rosidl_generator_tests/test/rosidl_generator_c/test_interfaces.c
```diff
@@ -45,21 +45,16 @@
...
-}
-#define EXPECT_TRUE(arg) if (!(arg)) { \
-    fputs(STRINGIFY(arg) " is not true\n", stderr); \
-    return 1; \
-}
-#define EXPECT_EQ(arg1, arg2) if ((arg1) != (arg2)) { \
-    fputs(STRINGIFY(arg1) " != " STRINGIFY(arg2) "\n", stderr); \
-    return 1; \
-}
+#define PUTS(arg, extra) fprintf(stderr, "%s%s\n", STRINGIFY(arg), extra)
```
fujitatomoya, line 51: Is that better to have space in `PUTS`, instead of `EXPECT_xxx` condition check?

```suggestion
#define PUTS(arg, extra) fprintf(stderr, "%s %s\n", STRINGIFY(arg), extra)
```

fujitatomoya (approved):
lgtm with green CI.

rosidl_generator_tests/test/rosidl_generator_c/test_interfaces.c
clalancette: I think it is fine either way.  Given that I've already run CI on this, and it is just in a test, I'm going to go with what is in here for now.  We can follow-up later if we want to make that change.

Issue #786, opened 2024-02-26
Add [[nodiscard]] attribute to C++ headers that return values, such as operators [Jazzy]


Ryanf55:
#### Feature description

Getters and operators could have the [[nodiscard]] attribute to prevent bugs.

#### Implementation considerations

See `/opt/ros/humble/include/geometry_msgs/geometry_msgs/msg/detail/point__struct.hpp`. This is ABI breaking because it increases the strictness of the API.

Before:
```c++
  bool operator!=(const Point_ & other) const
  {
    return !this->operator==(other);
  }
  

```

Proposed:
```c++
  [[nodiscard]] bool operator!=(const Point_ & other) const
  {
    return !this->operator==(other);
  }
```

Effect:

Using the `!=` operator without assigning the value would then issue a compile warning.

#### Related

Same as https://github.com/gazebosim/sdformat/issues/1358

wjwwood:
Reviewing from our issue triaging meeting, sounds like a nice improvement, help is appreciated.

jmachowinski:
Is this issue fixed and can be closed ?

clalancette:
It looks like it, thanks.  I'll go ahead and close it.

PR #801, opened 2024-04-21
Global use of nodiscard

CursedRock17:
This PR is meant to address #786 in the `rosidl` repo, so using [`nodiscard`](https://en.cppreference.com/w/cpp/language/attributes/nodiscard) on any getter functions and operators if possible. Adding `nodiscard` should prevent a lot of unnecessary bugs ahead of time. Though, I do have a couple things that we may need to address

1. Is this something that we want to do in any repository with C++, since it would reduce bugs projectwide? If that's the case, this issue should be expanded to [the ros2 repo](https://github.com/ros2/ros2), even if Jazzy is enduring a feature freeze.
2. To what extent should we be using `nodiscard`? At the original release of this PR I just went for getters and operators because of the implementation details:
> Getters and operators could have the [[nodiscard]] attribute to prevent bugs.

Though realistically many functions could utilize `nodiscard` to prevent weird issues.

3. Any operator that returns`*this` cannot use `nodiscard` because you must always assign that `this` value but each operator overload has `nodiscard`, so we would have to make extra functions which don't use `nodiscard`, getting rid of the purpose.

### DIFF
--- a/rosidl_runtime_cpp/include/rosidl_runtime_cpp/action_type_support_decl.hpp
+++ b/rosidl_runtime_cpp/include/rosidl_runtime_cpp/action_type_support_decl.hpp
@@ -27,6 +27,7 @@ namespace rosidl_runtime_cpp
  * \return Function handler for the action's typesupport.
  */
 template<typename T>
+[[nodiscard]]
 const rosidl_action_type_support_t * get_action_type_support_handle();
 
 }  // namespace rosidl_runtime_cpp
--- a/rosidl_runtime_cpp/include/rosidl_runtime_cpp/bounded_vector.hpp
+++ b/rosidl_runtime_cpp/include/rosidl_runtime_cpp/bounded_vector.hpp
@@ -714,6 +714,7 @@ private:
    * \param y A %BoundedVector of the same type as @a x
    * \return True if the size and elements of the vectors are equal
   */
+  [[nodiscard]]
   friend bool
   operator==(
     const BoundedVector & x,
@@ -734,6 +735,7 @@ private:
    * \param y A %BoundedVector of the same type as @a x
    * @return True if @a x is lexicographically less than @a y
   */
+  [[nodiscard]]
   friend bool
   operator<(
     const BoundedVector & x,
@@ -743,6 +745,7 @@ private:
   }
 
   /// Based on operator==
+  [[nodiscard]]
   friend bool
   operator!=(
     const BoundedVector & x,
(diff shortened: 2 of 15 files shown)


fujitatomoya (approved):
lgtm.

IMO this is better. Once we can have a consensus to go this path, i can help this.

clalancette:
Here's full CI with this in place:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20883/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15187/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21565/)

clalancette (approved):
It looks to me like this is all happy with the downstream.  And the issues on CI are known issues, unrelated to this PR.  So I'm going to go ahead and merge this one in, thanks @CursedRock17

clalancette:
> It looks to me like this is all happy with the downstream. And the issues on CI are known issues, unrelated to this PR. So I'm going to go ahead and merge this one in, thanks @CursedRock17

Ah, sorry, scratch that.

So, it turns out that we *cannot* update the `type_description/*` files like this.  In particular, those signatures have to match the ones that would have been generated by the rest of the rosidl pipeline.  What I'm going to suggest here is that we actually just remove those changes from this PR, and go with the rest of them.  In the future we can consider updating the `type_description/*` files when we update what the rosidl pipeline generates.

CursedRock17:
Interesting, one thing I was contemplating on touching on was inside of the `rosidl_generator_cpp`, in [msg__struct](https://github.com/ros2/rosidl/blob/rolling/rosidl_generator_cpp/resource/msg__struct.hpp.em#L344) would adding `nodiscard` here add problems throughout the pipeline?

clalancette:
> Interesting, one thing I was contemplating on touching on was inside of the `rosidl_generator_cpp`, in [msg__struct](https://github.com/ros2/rosidl/blob/rolling/rosidl_generator_cpp/resource/msg__struct.hpp.em#L344) would adding `nodiscard` here add problems throughout the pipeline?

It's not necessarily a problem; I'd be OK with adding that.  But we just have to make sure that the `type_description/*` files are updated in lockstep with whatever other changes we do.

Either way, I think we should do it in a separate PR; this one is almost good to go (just need to run CI on it one more time).

clalancette:
Another round of full CI for this:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20913/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15214/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21591/)

PR #805, opened 2024-04-26
Fix u16 docs and improve docs formatting

christophebedard:
1. I noticed that the docs for the `size` param for the u16 string sequence functions in `rosidl_runtime_c` are wrong. The `size` param refers to the size of the sequence, not the string. This is clear when looking at the implementations and at the docs of the other similar sequence functions.
2. I also modified the description of the `n` param for `rosidl_runtime_c__U16String__assignn_from_char` to clarify that it is the number of 8-bit chars in the string. I also mentioned that the function will return `false` if `n` is not even.
3. Finally, I also added some missing `*` after `/*` in docs (to get the usual `/**`).


fujitatomoya (approved):
lgtm

clalancette (approved):
Looks good to me.

Since this is only a documentation change, and the Rpr jobs are working again, I'm going to merge it in with just that passing.  Thanks for updating this!

PR #814, opened 2024-06-21
Silence one more gcc false-positive.

clalancette:
gcc 13 has false positives around array-bounds and stringop-overflow, so suppress them here while generating test cases.

@Crola1702 FYI, this should fix #813

PR #810, opened 2024-06-12
Suppress warnings in the benchmarks for upstream GCC false positives.

clalancette:
This will fix at least one of the warnings as seen in https://ci.ros2.org/view/nightly/job/nightly_linux_release/3103/gcc/new/

@Crola1702 FYI

PR #815, opened 2024-06-26
Patch warnings (#810) (#814)

Crola1702:
gcc 13 has false positives around array-bounds and stringop-overflow, so suppress them here while generating test cases. (#810)

This will fix at least one of the warnings as seen in https://ci.ros2.org/view/nightly/job/nightly_linux_release/3103/gcc/new/ (#814)

Warnings still happening on Jazzy ([reference build](https://build.ros2.org/view/Jci/job/Jci__nightly-release_ubuntu_noble_amd64/27/gcc/))


ahcorde:
Same here, I prefer to use mergify to track better commits https://github.com/ros2/rosidl/pull/810#issuecomment-2191701066 PR https://github.com/ros2/rosidl/pull/816

PR #821, opened 2024-08-01
Support empy3 and empy4

ahcorde:


rosidl_adapter/rosidl_adapter/resource/__init__.py
```diff
@@ -14,10 +14,13 @@
 
 from io import StringIO
 import os
+from packaging import version
 import sys
 
 import em
 
+if version.parse(em.__version__) >= version.parse('4.0.0'):
+    from em import Configuration
```
clalancette, line 23: Rather than parsing the version, I'm wondering if we should instead attempt to import the thing we want.  Something like:

```
try:
    from em import Configuration
    em_has_configuration = True
except ImportError:
    em_has_configuration = False
```

And then below, if em_has_configuration, we use that (and the newer Interpreter API).  Otherwise, use the older API.  What do you think?

clalancette:
I took this for a spin locally.  I have good news and bad news.

The good news is that this works, and seems to keep compatibility with both Empy 3 and Empy 4.

The bad news is that with Empy 4, there is at least one bug.  In particular, every time a rosidl_generator_py template file is expanded, it is printed to the console.  This makes the output when building with it extremely noisy.  I haven't dug into what is causing this, but I think we should find a way to fix/suppress this before we merge this in.

ahcorde:
> I took this for a spin locally. I have good news and bad news.
> 
> The good news is that this works, and seems to keep compatibility with both Empy 3 and Empy 4.
> 
> The bad news is that with Empy 4, there is at least one bug. In particular, every time a rosidl_generator_py template file is expanded, it is printed to the console. This makes the output when building with it extremely noisy. I haven't dug into what is causing this, but I think we should find a way to fix/suppress this before we merge this in.

I noticed this, That's why is still a draft

ahcorde:
@clalancette should be fixed now

rosidl_adapter/rosidl_adapter/resource/__init__.py
```diff
@@ -38,25 +44,39 @@ def expand_template(template_name, data, output_file, encoding='utf-8'):
 def evaluate_template(template_name, data):
     global _interpreter
     # create copy before manipulating
+    print('data', data)
```
clalancette, line 47:
```suggestion
```

clalancette (changes requested):

rosidl_pycommon/rosidl_pycommon/__init__.py
```diff
@@ -218,7 +239,10 @@ def _expand_template(template_name, **kwargs):
             'beforeInclude', name=str(template_path), file=h, locals=kwargs)
         content = h.read()
     try:
-        interpreter.string(content, str(template_path), kwargs)
+        if em_has_configuration:
+            interpreter.string(content, locals=kwargs)
+        else:
+            interpreter.string(content, str(template_path), kwargs)
```
clalancette, line 245: Sorry, I just noticed one more thing.  This looks different from both line 243, and line 193 above.  I'm wondering if we should just make this consistent with everything else:

```suggestion
            interpreter.string(content, template_path, locals=kwargs)
```

What do you think?

clalancette (commented):
One more thought, then I think we are good to run CI on this.

rosidl_pycommon/rosidl_pycommon/__init__.py
ahcorde: https://github.com/ros2/rosidl/pull/821/commits/761d831d430a32f26548f02876e9675414f7a444

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21692/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15967/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1193/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22397/)

clalancette (approved):
This looks good to me, but I'm going to suggest we do a full CI run here (i.e. no `--packages-above-and-dependencies`).  That's just because this is so deep in the rosidl pipeline, I want to make sure everything works.

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21696/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15971/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1198/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22401/)

yashi:
should we backport this to Jazzy, Iron, and Humble?

fujitatomoya:
AFAIK, ubuntu nobel and jammy both use `3.3.4` so basically we do not use empy4. besides https://github.com/ros2/rosidl/pull/821#issuecomment-2271098700, we should wait until we make sure `rolling` works with `em.Configuration` without any problems. and then, we can consider backporting?

or do you have any specific requirement for this backporting?

yashi:
No I don't. We can wait.

clalancette:
@Mergifyio backport jazzy

PR #825, opened 2024-09-05
Allow using NaN default values

henrygerardmoore:
Hello, this is my first attempt at a contribution here, so please let me know if I need to change something.

Anyway, I noticed that you cannot default to NaN in `.msg` definitions, which could be useful, especially when NaNs are used as a sentinel value, which is somewhat common. This PR is my attempt to add this capability.

Is the docstring I added correct? The `rosidl_generator_cpp` `primitive_value_to_cpp` function says that the type of value is a python builtin, but when I tried `isnan` in `rosidl_generator_c` it didn't work, while `== 'nan'` did. I'm not sure how to test `rosidl_generator_cpp` so I would appreciate a double check or someone telling me how to test it.

Oh also, I tried adding a bogus message in the `.msg` file (`float32 FLOAT32_FAKE=test`) to make sure it wouldn't just convert to NaN, and it didn't. Just failed to build, as you would expect.


fujitatomoya (commented):
this is a replacement of https://github.com/ros2/rosidl/pull/789.

@Ryanf55 since you are the original author for https://github.com/ros2/rosidl/pull/789, if you can review this, that would be really appreciated.

rosidl_generator_c/rosidl_generator_c/__init__.py
```diff
@@ -209,9 +220,13 @@ def basic_value_to_c(type_, value):
         return f'{value}ull'
 
     if 'float' == type_.typename:
+        if 'nan' == value:
```
fujitatomoya, line 223: Can we use `isnan(float(value))` instead? that would be more reliable check for it?

rosidl_generator_cpp/resource/idl__struct.hpp.em
```diff
@@ -31,6 +31,7 @@ include_directives = set()
 #include <memory>
 #include <string>
 #include <vector>
+#include <limits>
```
fujitatomoya, line 34: I think this needs to be line 31 to avoid linter warning.

fujitatomoya (changes requested):
@henrygerardmoore thank you for the PR.

IMO we can revise https://github.com/ros2/rosidl/pull/789 or borrow some fixes from there? what do you think?

rosidl_generator_c/rosidl_generator_c/__init__.py
henrygerardmoore: Yes, that makes sense. That way it would catch any `Nan` or `NAN` or similar :+1:

rosidl_generator_cpp/resource/idl__struct.hpp.em
henrygerardmoore: Yes, good point.

henrygerardmoore:
@fujitatomoya Thank you for the review! Apologies that I didn't see #789 before making this, I should've checked for similar existing PRs. I have applied your suggestions, they both made sense and seemed correct to me.

Ryanf55:
> this is a replacement of #789.
> 
> @Ryanf55 since you are the original author for #789, if you can review this, that would be really appreciated.

yea, compared to my PR, it's
* Missing the changes to the C generator
* Missing support of NAN constants

We've been using a variation of my branch internally for months in both python and C++, so I'll make sure my upstream is synced. 
I'd like to combine the best of both PR's. 

I never got an answer from the ROS maintainers on how they want to handle equality of messages that contain NaN fields. The lack of consenses is what held me up spending more time on this.

rosidl_generator_c/rosidl_generator_c/__init__.py
```diff
@@ -209,9 +222,13 @@ def basic_value_to_c(type_, value):
         return f'{value}ull'
 
     if 'float' == type_.typename:
+        if isnan(float(value)):
+            return 'NAN'
```
Ryanf55, line 226: I chose to use `'nanf(\"\")'` on my PR to avoid the c-style cast to a double later.

henrygerardmoore:
@Ryanf55 thank you for taking a look! I can close this PR now that I am aware of yours.

rosidl_generator_cpp/resource/msg__struct.hpp.em
```diff
@@ -287,7 +287,17 @@ non_defaulted_zero_initialized_members = [
...
 @[    end if]@
 @[   elif constant.type.typename == 'float']@
+@[    if constant.value == 'nan']
+    std::numeric_limits<float>::quiet_NaN()@
+@[    else]@
     @(constant.value)f@
+@[    end if]@
+@[   elif constant.type.typename == 'double']@
+@[    if constant.value == 'nan']
+    std::numeric_limits<double>::quiet_NaN()@
```
Ryanf55, line 297: Instead of coding this, I chose to use the `primitive_value_to_cpp` function to promote reuse.

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
```diff
@@ -198,13 +199,18 @@ def primitive_value_to_cpp(type_, value):
...
         'short', 'unsigned short',
         'char', 'wchar',
-        'double', 'long double',
         'octet',
         'int8', 'uint8',
         'int16', 'uint16',
     ]:
         return str(value)
 
+    if type_.typename in ['double', 'long double']:
```
Ryanf55, line 208: I also checked if it was a float here.

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
```diff
@@ -198,13 +199,18 @@ def primitive_value_to_cpp(type_, value):
...
+        if isnan(value):
+            return 'std::numeric_limits<double>::quiet_NaN()'
```
Ryanf55, line 210: Instead of hard-coding as double, I used the type name here.

rosidl_generator_cpp/rosidl_generator_cpp/__init__.py
```diff
@@ -198,13 +199,18 @@ def primitive_value_to_cpp(type_, value):
...
         'char', 'wchar',
-        'double', 'long double',
         'octet',
         'int8', 'uint8',
         'int16', 'uint16',
     ]:
         return str(value)
 
+    if type_.typename in ['double', 'long double']:
+        if isnan(value):
```
Ryanf55, line 209: For some reason, I needed to surroudn `value` in a `float()` cast otherwise it threw.

Ryanf55:
> @Ryanf55 thank you for taking a look! I can close this PR now that I am aware of yours.

Yea, please review mine and see if you can find any issues. A lot of the content is very similar.

PR #836, opened 2024-11-15
Update check-copied-type-description-sources.yaml

mosfet80:
fix node.js <20 deprecation


sloretz (approved):
Thank you for the PR!

PR #846, opened 2024-11-27
Deterministic iteration order for reproducible codegen

harrysarson:
Two small changes to get derministic iteration order which makes codegen reproducible

- Use OrderedDict for reproducible codegen

   - Python set's do not have a deterministic iteration order (unlike dicts),
therefore replace usage of set with OrderedDict's for the `includes`
variable in `full__description.c.em` to make codegen reproducible.

- Use deterministic iteration in type_support header

   - The `include_directives` set is used by the nested templates to
prevent generation of duplicate include directives, but is not iterated
over by those nested templates.

    - In `idl__type_support.c.em` rather that iterating over the
`include_directives` set (which introduces non-determinism) to generate
the includes for top level header files, we iterate over those top level
includes as a list and seperately construct the `include_directives` set
from the list. This ensures deterministic codegen.

### DIFF
--- a/rosidl_generator_c/resource/full__description.c.em
+++ b/rosidl_generator_c/resource/full__description.c.em
@@ -1,5 +1,6 @@
 @# Included from rosidl_generator_c/resource/idl__description.c.em
 @{
+from collections import OrderedDict
 from rosidl_generator_c import escape_string
 from rosidl_generator_c import idl_structure_type_to_c_include_prefix
 from rosidl_generator_c import type_hash_to_c_definition
@@ -31,7 +32,7 @@ def utf8_encode(value_string):
   return escape_string(repr(value_string.encode('utf-8'))[2:-1])
 
 implicit_type_names = set(td['type_description']['type_name'] for td, _ in implicit_type_descriptions)
-includes = set()
+includes = OrderedDict()
 toplevel_msg, _ = toplevel_type_description
 
 for referenced_td in toplevel_msg['referenced_type_descriptions']:
@@ -40,7 +41,7 @@ for referenced_td in toplevel_msg['referenced_type_descriptions']:
     names = referenced_td['type_name'].split('/')
     _type = NamespacedType(names[:-1], names[-1])
     include_prefix = idl_structure_type_to_c_include_prefix(_type, 'detail')
-    includes.add(include_prefix + '__functions.h')
+    includes[include_prefix + '__functions.h'] = None
 
 full_type_descriptions = [toplevel_type_description] + implicit_type_descriptions
 full_type_names = [t['type_description']['type_name'] for t, _ in full_type_descriptions]
--- a/rosidl_generator_c/resource/idl__type_support.c.em
+++ b/rosidl_generator_c/resource/idl__type_support.c.em
@@ -16,16 +16,19 @@ include_parts = [package_name] + list(interface_path.parents[0].parts) + [
     'detail', convert_camel_case_to_lower_case_underscore(interface_path.stem)]
 include_base = '/'.join(include_parts)
 
-include_directives = {
+top_level_includes = [
   'rosidl_typesupport_interface/macros.h',
   include_base + '__type_support.h',
   include_base + '__struct.h',
-  include_base + '__functions.h'}
+  include_base + '__functions.h']
+
+include_directives = set(top_level_includes)
+
 }@
 
 #include <string.h>
 
-@[for header_file in include_directives]@
+@[for header_file in top_level_includes]@
 #include "@(header_file)"
 @[end for]@
 


clalancette (commented):
While I'm all for making code generation reproducible, does this actually have an effect?  My reading of this is that the `include_directives` set() is used just to tell if we've every included a particular header before.  That is, we don't actually iterate over `include_directives` at any point, so this is basically a no-op.  Can you give an example of what changes with this in place?

harrysarson:
> we don't actually iterate over include_directives at any point

We iterated over `include_directives` here https://github.com/ros2/rosidl/blob/dd0530023e758914c9e8e48ef0ddc1f797161a64/rosidl_generator_c/resource/idl__type_support.c.em#L28

There are other `include_directives` variables (e.g the one defined at here https://github.com/ros2/rosidl/blob/dd0530023e758914c9e8e48ef0ddc1f797161a64/rosidl_generator_c/resource/idl__struct.h.em) and for those we _could_ keep using a set. However, I think it is worth being consistent and using an `OrderedDict` for all the different  `include_directives` variables.

clalancette:
> We iterated over `include_directives` here

Fair enough.  I think then we should just replace that one with a list.  We shouldn't need the rest of the changes here.

harrysarson:
The `include_directives` defined here https://github.com/ros2/rosidl/blob/dd0530023e758914c9e8e48ef0ddc1f797161a64/rosidl_generator_c/resource/idl__type_support.c.em#L19

is passed to other templates https://github.com/ros2/rosidl/blob/dd0530023e758914c9e8e48ef0ddc1f797161a64/rosidl_generator_c/resource/idl__type_support.c.em#L45-L49 and https://github.com/ros2/rosidl/blob/dd0530023e758914c9e8e48ef0ddc1f797161a64/rosidl_generator_c/resource/idl__type_support.c.em#L61-L65

and those templates then pass `include_directives` to other templates.

---

I guess we could do 

```patch
--- a/rosidl_generator_c/resource/idl__type_support.c.em
+++ b/rosidl_generator_c/resource/idl__type_support.c.em
@@ -16,16 +16,19 @@ include_parts = [package_name] + list(interface_path.parents[0].parts) + [
     'detail', convert_camel_case_to_lower_case_underscore(interface_path.stem)]
 include_base = '/'.join(include_parts)
 
-include_directives = {
+include_directives_ordered = [
   'rosidl_typesupport_interface/macros.h',
   include_base + '__type_support.h',
   include_base + '__struct.h',
-  include_base + '__functions.h'}
+  include_base + '__functions.h']
+
+include_directives = set(include_directives_ordered)
+
 }@
 
 #include <string.h>
 
-@[for header_file in include_directives]@
+@[for header_file in include_directives_ordered]@
 #include "@(header_file)"
 @[end for]@
 ```

so we iterate over the ordered list but and use the set in the other templates

**Comment by clalancette** [comment]
> so we iterate over the ordered list but and use the set in the other templates

Yeah, I like this.  I'd say let's go ahead with that change.  Thanks!

**Comment by harrysarson** [comment]
I have had a go at this, please see latest commits. I kept the `OrderedDict` in `rosidl_generator_c/resource/full__description.c.em` for the `includes` variable (this matches what we do here for example https://github.com/ros2/rosidl/blob/2f900108d0bf234139c85b25b7531f20b1cb5a51/rosidl_generator_cpp/resource/msg__struct.hpp.em#L36).

File: rosidl_generator_c/resource/full__description.c.em [hunk 1]
```diff
@@ -31,7 +32,7 @@ def utf8_encode(value_string):
   return escape_string(repr(value_string.encode('utf-8'))[2:-1])
 
 implicit_type_names = set(td['type_description']['type_name'] for td, _ in implicit_type_descriptions)
-includes = set()
+includes = OrderedDict()
```
clalancette, line 35: For this one, I don't think we need a dictionary at all.  We are never using this beyond the one time we iterate, so we never need the lookup performance.  Thus I think this can be a list:

```suggestion
includes = []
```

rosidl_generator_c/resource/full__description.c.em
```diff
@@ -40,7 +41,7 @@ for referenced_td in toplevel_msg['referenced_type_descriptions']:
     names = referenced_td['type_name'].split('/')
     _type = NamespacedType(names[:-1], names[-1])
     include_prefix = idl_structure_type_to_c_include_prefix(_type, 'detail')
-    includes.add(include_prefix + '__functions.h')
+    includes[include_prefix + '__functions.h'] = None
```
clalancette, line 44: Following from above, we can just append to the list:

```suggestion
    includes.append(include_prefix + '__functions.h')
```

clalancette (changes requested):
One more change, then this will be ready for CI.  Thanks for iterating.

rosidl_generator_c/resource/full__description.c.em
harrysarson: I believe this line might add the same include twice, the `set` (or `OrderedDict`) is used to deduplicate. So I am not sure a list can be used here.

harrysarson: Saying that I tried a (very) basic test and didn't see any duplicate `#includes` so it _might_ be that `toplevel_msg['referenced_type_descriptions']` is already de-deduplicated?

clalancette: That's actually a fair point, I hadn't considered that.  In that case, I think we can stick with the set, but instead change the iteration over the set to be:

```
@[for header_file in sorted(includes)]
```

harrysarson: Yep I can do this, just want to check we are happy with the performance of `sorted` (copying all the items plus `O(n log n)` sorting).

I originally went with `OrderedDict` because its already used in `*.em` files[1] (although admittedly `OrderedDict` is never used with `None` values) and I was happy that switching from a `set` to an `OrderedDict` would not add any meaningful overhead to compute required for codegen.

---

1: e.g. https://github.com/ros2/rosidl/blob/2f900108d0bf234139c85b25b7531f20b1cb5a51/rosidl_generator_cpp/resource/msg__struct.hpp.em#L35

clalancette: > Yep I can do this, just want to check we are happy with the performance of `sorted` (copying all the items plus `O(n log n)` sorting).

Another good point.

> I originally went with `OrderedDict` because its already used in `*.em` files[1] (although admittedly `OrderedDict` is never used with `None` values) and I was happy that switching from a `set` to an `OrderedDict` would not add any meaningful overhead to compute required for codegen.

The reason that this is causing me to try other things is that an `OrderedDict` with `None` as the values feels like the incorrect structure.  But fair enough, I've tried other things that have other drawbacks, so I'm OK with this.  Thanks for iterating with me, I'm going to withdraw/resolve this comment and run CI here.

clalancette (approved):

clalancette:
Pulls: https://github.com/ros2/rosidl/pull/846
Gist: https://gist.githubusercontent.com/clalancette/3b0f4ae9440be6b9395a03b5fb6fdfa0/raw/b854cb8c05665050c8ef1c8f0a5c5ecd0630c6a4/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14929
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22241/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16501/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1763/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23009/)

PR #851, opened 2025-03-07
Specify rosidl version dependency in CMake

Ryanf55:
Avoid annoying issues when you accidentally mix up versions of ROS 2. Jazzy accidentally pulling in humble packages due to docker cache issues cause us a world of hurt when rosidl is built with mixed distros. This is also common when working in vcs and one of the repos fails to switch branches without you noticing.

A trivial change like this allows us to catch it at configure time.

Sponsored by AeroVironment, Inc.


fujitatomoya (commented):
out of curiosity.

> Jazzy accidentally pulling in humble packages due to docker cache issues cause us a world of hurt 

how come this could happen? Jazzy is for Noble, and Humble is for Jammy, this can be managed under ubuntu codename via apt-repository?

on the same ubuntu codename, we can see the same packages with different distro. (for example, `ros-humble-demo-nodes-cpp`, `ros-iron-demo-nodes-cpp` and `ros-rolling-demo-nodes-cpp`)

Ryanf55:
As I mentioned, there's two ways this happens. 

Docker caching done poorly, or vcs failing to update image tags. It's a pretty regular occurrence for me working on a mixed distribution product team and frequently needing to switch environments with flaky development workflows. Anything that can be done that makes problems like this immediately obvious is extremely helpful. It's exacerbated in ROSIDL because we have a few forks from unmerged ROSIDL features.

This has no negative effect on current users, and only a slight increase in maintenance, however a large potential benefit for people like me.

christophebedard:
Would using version constraints in `package.xml` files work? https://ros.org/reps/rep-0149.html#dependency-tags

Ryanf55:
No, if you have a different version installed than required by rosdep, and only run colcon build, it won't be caught by colcon/CMake.

Ryanf55:
> Would using version constraints in `package.xml` files work? https://ros.org/reps/rep-0149.html#dependency-tags

Also, are these even implemented yet?
https://github.com/ros-infrastructure/rosdep/issues/803

christophebedard:
I think they are implemented in `catkin_pkg` (for `package.xml` parsing), but probably not used at build-time or at rosdep/dependency installation-time. They are used by bloom for packaging (e.g., to include the corresponding debian version constraints).

PR #881, opened 2025-07-29
Use `std::byte` in cpp files for bytes

InvincibleRMC:
## Description

Resolves TODO to use std::byte. Downstream users will need to update if they use IDL bytes. The message generation documentation would also need to be updated. Will also likely need a ROS Discourse post.

Yes users will need to switch to using std::byte rather than the existing conversion like int.

No


rosidl_generator_tests/test/rosidl_generator_cpp/test_array_generator.hpp
```diff
@@ -44,6 +44,37 @@ void test_vector_fill(
...
+  typename std::enable_if<
+    std::is_same<typename C::value_type, std::byte>::value
+  >::type * = nullptr
+>
+void test_vector_fill(
+  C * container, size_t size,
+  typename C::value_type min, typename C::value_type max)
+{
+  if (size > 0 && min < max) {
+    int step = (std::to_integer<int>(max) - std::to_integer<int>(min)) / static_cast<int>(size);
```
ahcorde, line 68: include `<cstdint>`

rosidl_generator_tests/test/rosidl_generator_cpp/test_array_generator.hpp
```diff
@@ -44,6 +44,37 @@ void test_vector_fill(
...
+ * Minimum and maximum values for the type and distributed values in the middle.
+ * @param C Container (vector, array, etc) to be filled
+ * @param size How many elements to fill in. Must size<=container_size
+ * @param min Minimum value in the range to fill.
+ * @param max Maximum value in the range to fill.
+ */
+template<
+  typename C,
+  typename std::enable_if<
+    std::is_same<typename C::value_type, std::byte>::value
```
ahcorde, line 60: include `<cstddef>`

rosidl_generator_tests/test/rosidl_generator_cpp/test_array_generator.hpp
```diff
@@ -44,6 +44,37 @@ void test_vector_fill(
...
+>
+void test_vector_fill(
+  C * container, size_t size,
+  typename C::value_type min, typename C::value_type max)
+{
+  if (size > 0 && min < max) {
+    int step = (std::to_integer<int>(max) - std::to_integer<int>(min)) / static_cast<int>(size);
+
+    (*container)[0] = min;
+    for (size_t i = 1; i < size - 1; i++) {
```
ahcorde, line 71: include `<cstddef>`

rosidl_generator_tests/test/rosidl_generator_cpp/test_msg_builder.cpp
```diff
@@ -24,7 +24,7 @@ TEST(Test_msg_initialization, build) {
   ::rosidl_generator_tests::msg::BasicTypes basic =
     ::rosidl_generator_tests::build<::rosidl_generator_tests::msg::BasicTypes>()
     .bool_value(true)
-    .byte_value(5)
+    .byte_value(std::byte{5})
```
ahcorde, line 27: include `<cstddef>`

rosidl_generator_tests/test/rosidl_generator_cpp/test_interfaces.cpp
```diff
@@ -168,7 +168,7 @@ void test_message_basic_types(rosidl_generator_tests::msg::BasicTypes message)
 #ifdef __linux__
 #pragma GCC diagnostic pop
 #endif
-  TEST_BASIC_TYPE_FIELD_ASSIGNMENT(message, byte_value, 0, 255)
+  TEST_BASIC_TYPE_FIELD_ASSIGNMENT(message, byte_value, std::byte{0}, std::byte{255})
```
ahcorde: include `<cstddef>`

rosidl_generator_tests/test/rosidl_generator_cpp/test_msg_initialization.cpp
```diff
@@ -47,7 +47,7 @@ make_scope_exit(Callable callable)
 TEST(Test_msg_initialization, no_arg_constructor) {
   rosidl_generator_tests::msg::Defaults def;
   ASSERT_TRUE(def.bool_value);
-  ASSERT_EQ(50, def.byte_value);
+  ASSERT_EQ(std::byte{50}, def.byte_value);
```
ahcorde: include `<cstddef>`

rosidl_generator_tests/test/rosidl_generator_cpp/test_srv_initialization.cpp
```diff
@@ -22,7 +22,7 @@ TEST(Test_srv_initialization, no_arg_request_constructor) {
 
   rosidl_generator_tests::srv::BasicTypes::Request basic_types;
   EXPECT_EQ(false, basic_types.bool_value);
-  EXPECT_EQ(0, basic_types.byte_value);
+  EXPECT_EQ(std::byte{0}, basic_types.byte_value);
```
ahcorde: include `<cstddef>`

rosidl_typesupport_introspection_tests/test/introspection_libraries_under_test.hpp
```diff
@@ -933,7 +933,7 @@ struct Example<rosidl_typesupport_introspection_tests::msg::BoundedSequences>
     auto message =
       std::make_unique<rosidl_typesupport_introspection_tests::msg::BoundedSequences>();
     message->bool_values.push_back(true);
-    message->byte_values.push_back(0x1B);
+    message->byte_values.push_back(std::byte{0x1B});
```
ahcorde, line 937: include `<cstddef>`

ahcorde (changes requested):

fujitatomoya (approved):
lgtm with green CI. i would like to have another review.

christophebedard:
Note that this is a breaking change.

InvincibleRMC:
While working on #882 Found a big use for std::byte simplifying to msg__traits.hpp. Will probably split it out of #882.

Edit: My cool simplification will need to wait for C++ 20 and the addition of `char8_t` since it needs a way to distinguish between `unsigned char` and `uint8_t`. Or I guess hypothetically we could add a custom `roschar8_t` until C++ 20 similar to `BoundedVector`.

InvincibleRMC:
> Note that this is a breaking change.

If/When this gets merged I can help fix downstream bugs.

fujitatomoya:
Pulls: ros2/rosidl#881
Gist: https://gist.githubusercontent.com/fujitatomoya/c0b962717673b135a1a582f9282c0413/raw/96d7a6596347cf690185cc4c41a1172af1b89aec/ros2.repos
BUILD args:  --packages-above-and-dependencies rosidl_generator_cpp rosidl_generator_tests rosidl_runtime_cpp rosidl_typesupport_introspection_tests
TEST args:  --packages-above rosidl_generator_cpp rosidl_generator_tests rosidl_runtime_cpp rosidl_typesupport_introspection_tests
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16671
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24543/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18591/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3999/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25085/)

InvincibleRMC:
Unsurprisingly this causes errors in `rclcpp` and presumably other packages further downstream. Should I try and add things to mitigate the pain of migration e.j. add in way to set bytes from the old `unsigned char` type with a deprecation warning and/or add a retrieval method of the bytes attributes which is the old `unsigned char` type. Or is the rip off bandaid approach fine?

christophebedard:
We usually avoid breaking changes and try to deprecate something first in one release and then remove it in the next release, see https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Developer-Guide.html#deprecation-strategy. That's what I really meant with my "breaking change" comment above; sorry I wasn't clear.

So yeah, we should try to support setting/getting an `unsigned char` while keeping `std::byte` underneath and let users migrate to it, which would let users migrate to `std::byte` without breaking their packages.

InvincibleRMC:
The following works for everything except compound assignments e.j. `+=` those overloads would need to be added. Something similar for adding support for containers of byte would also need to be implemented. Feel free to give feedback or if there is a more clever approach let me know.

EDIT: Updated so size matches byte size

```C++
#include <cstddef>
#include <iostream>

struct DeprecatedHelperByte {
    std::byte storage_{};

    [[deprecated("Assign std::byte instead")]]
    operator unsigned char() const noexcept {
        std::cout << "getter called char\n";
        return static_cast<unsigned char>(storage_);
    }

    operator std::byte() const noexcept {
        std::cout << "getter called byte\n";
        return storage_;
    }

    DeprecatedHelperByte& operator=(std::byte b) noexcept {
        std::cout << "setter called byte\n";
        storage_ = b;
        return *this;
    }

    [[deprecated("Assign std::byte instead")]]
    DeprecatedHelperByte& operator=(unsigned char b) noexcept {
        std::cout << "setter called char\n";
        storage_ = std::byte{b};
        return *this;
    }
};

struct Foo {
    DeprecatedHelperByte data{};
};

int main() {
    Foo f;

    f.data = std::byte{1};
    unsigned char b = f.data;
    f.data = 1;
    std::byte c = f.data;

    std::cout << "\nSize of Foo: " << sizeof(Foo) << '\n';
}
```

InvincibleRMC:
Hmm while this works great it would need a specialization for all the containers (vector, array, bounded_sequence) ROS uses. 

Will take into a look into just adding conversion methods.

PR #882, opened 2025-07-29
Add @optional support to `rosidl_generator_cpp`

### LINKED ISSUE #850: Feature Request: Optional Attributes
DDS supports the [optional attribute in IDL 4.2 specification](https://www.omg.org/spec/IDL/4.2/PDF) attribute section 8.3.1.3. In C++, now that we have C++17, we can use std::optional.

A prior dicussion is here: https://discourse.ros.org/t/optional-fields-in-message/991/16

Consider adding a `@optional` specifier to rosidl that translates to the DDS optional .

Amend the ROS 2 design documentation for `Standardized Annotations` to add optional:
https://design.ros2.org/articles/idl_interface_definition.html

It's supported by:
* [FastDDS](https://fast-dds.docs.eprosima.com/en/latest/fastddsgen/dataTypes/dataTypes.html#optional-members)
* [RTI](https://community.rti.com/static/documentation/connext-dds/current/doc/manuals/connext_dds_professional/extensible_types_guide/extensible_types/Optional_Members.htm)
* [CycloneDDS](https://cyclonedds.io/content/guides/supported-idl.html#optional)
* [Zenoh](https://github.com/colinhacks/zod/issues/310#issuecomment-794533682)

Workarounds: Bounded array with a max length of 1, but the syntax is funky.

InvincibleRMC:
## Description

In Draft state since there are downstream `rmw` problems (tested with fastrps). I assume similar problems will occur with other rmw implementations.

For example calling .size() on the `const class std::optional<std::__cxx11::basic_string<char> >`

```C++
// Member: optional_string
  current_alignment += padding +
    eprosima::fastcdr::Cdr::alignment(current_alignment, padding) +
    (ros_message.optional_string.size() + 1);
  ```
  
  A similar problem occurs with containers.

Fixes #850 

Since this adds optionals users will not have any breaking changes.


InvincibleRMC:
I believe this is as far this PR can go without updating each of the dds `rosidl_typesupport_**_cpp`. I'm not sure if it is better to merge this and then update them. Update them then merge this. Or try and so some simultaneous merge. The only downstream change would effect if people had already added the `@optional` annotation.

InvincibleRMC:
Found a bug in container type_traits working on a fix.

PR #905, opened 2025-09-24
Fix `@optional` for string literals

### LINKED ISSUE #904: `@optional` parsing in `.msg` is backwards incompatible
The recently introduced `@optional` syntax in the `.msg` format breaks backwards compatibility: https://github.com/ros2/rosidl/blob/4edcac1c0aa162d4c231389811c105214c74a439/rosidl_adapter/rosidl_adapter/parser.py#L533-L543

Before the above change, `string s @optional` would be interpreted as a string field with default value being the literal string `@optional`, but now it would be interpreted as an optional field. This creates problems for writing parsers, because there is no way to know which interpretation to use without also knowing the ROS2 version.

This ambiguity can be easily avoided with a slightly different syntax. I suggest one of the following:

1. Require `@optional` be required to appear first on the line, so instead we have `@optional string s`. This avoids any ambiguity since `@` is not permitted to appear as part of a type name.
2. Require it to appear on its own line like so:
```
@optional
string s
```

InvincibleRMC:
## Description

Changes a flaw implementation to allow `@optional` in defaults and constants.

Fixes #904

Should unbreak backwards compatibility.

### DIFF
--- a/rosidl_adapter/rosidl_adapter/parser.py
+++ b/rosidl_adapter/rosidl_adapter/parser.py
@@ -530,8 +530,7 @@ def parse_message_string(pkg_name: str, msg_name: str,
             if not line:
                 continue
 
-        annotation_index = line.rfind(OPTIONAL_ANNOTATION)
-        if annotation_index >= 0:
+        if line.startswith(OPTIONAL_ANNOTATION):
             if is_optional:
                 raise MultipleOptionalAnnotations(
                     f'Already declared @optional. Error detected with {line}.')
--- a/rosidl_adapter/test/data/msg/Test.expected.idl
+++ b/rosidl_adapter/test/data/msg/Test.expected.idl
@@ -8,6 +8,8 @@ module test_msgs {
     module Test_Constants {
       @optional
       const float INLINE_CONSTANT = 32.0;
+      @optional
+      const string OPTIONAL_STRING_CONST = "@optional";
     };
     @verbatim (language="comment", text=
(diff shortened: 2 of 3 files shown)


InvincibleRMC:
cc: @christophebedard

fujitatomoya (approved):
with stricter check by this PR, only lines beginning with `@optional` will trigger the related logic. lgtm with green CI.

fujitatomoya:
Pulls: ros2/rosidl#905
Gist: https://gist.githubusercontent.com/fujitatomoya/5f8a56a176a8a79da67447b701142df6/raw/228cbba699201fb8f10117fcd2561091479d55ea/ros2.repos
BUILD args:  --packages-above-and-dependencies rosidl_adapter
TEST args:  --packages-above rosidl_adapter
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17050
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/25030/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19091/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4504/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25484/)

mjcarroll (approved):
Thanks @InvincibleRMC for the quick fix.

christophebedard (approved):

Issue #859, opened 2025-04-01
Speeding up rosidl and related projects


TonyWelte:
I'm starting this issue to discuss a few strategies that could be used to speed up build times for interface packages. Instead of spamming issues, I have listed here four different strategies that could help accelerate builds. Also most/all of these affect other generators too (rosidl_typesupport, rosidl_typesupport_fastrtps, etc).

First, for reference, let's look at the current build process for interfaces (here px4_msgs, `.ninja_log` viewed with [https://ui.perfetto.dev/](https://ui.perfetto.dev/)):

![Image](https://github.com/user-attachments/assets/25b3fc77-9f7e-46ee-80d7-d8f0e1368654)

The build process consists of two parts: generating and compiling files. Both can be optimized.

## Use of precompiled headers

For some source files, most of the "compilation" time is actually spent parsing the file. This is caused by includes resulting in large files. All messages have the same set of includes that are always present and need to be parsed for each translation unit.

Precompiled headers can significantly reduce the compilation time of these files.

Issues with precompiled headers:
- Dealing with indirect includes. For instance, `rosidl_typesupport_cpp` includes `__struct.hpp` files (generated by `rosidl_generator_cpp`) which themselves include standard library headers that expand into thousands of lines of code. Those indirect dependencies would have to be directly included in the list of headers to precompile, or `rosidl_generator_cpp` would need to provide a header file containing them (that `rosidl_typesupport_cpp` could add to its list of precompiled headers).
- For small projects, the time spent generating the precompiled header might be longer than the time saved by using it. Precompiled headers are processed by a single thread. For some projects, the time spent parsing the headers in parallel for each message might be shorter than the time spent generating the precompiled header.

On my machine, the gain from just using precompiled headers is not that high since most of the time is spent generating files (10% speedup for `px4_msgs`).

Forks with precompiled headers added:
- [rosidl](https://github.com/TonyWelte/rosidl/tree/feature/precompiled-headers)
- [rosidl_python](https://github.com/TonyWelte/rosidl_python/tree/feature/precompiled-headers)
- [rosidl_typesupport](https://github.com/TonyWelte/rosidl_typesupport/tree/feature/precompiled-headers)
- [rosidl_typesupport_fastrtps](https://github.com/TonyWelte/rosidl_typesupport_fastrtps/tree/feature/precompiled-headers)

## Propagate the update time of the message files

[...]

PR #910, opened 2025-10-01
Add DEPENDS_EXPLICIT_ONLY to remove implicit dependencies

TonyWelte:
## Description

This PR addresses one of the build speedup mentioned in #859. 

When `add_custom_command` is used to generate files that are then used by a target, the `add_custom_command` becomes implicitly dependent on the target dependencies. In rosidl `add_custom_command` is used to generate files from the IDL files, the generator does not depend on anything else (except rosidl_generator_c[pp] whose generators depend on rosidl_generator_type_description).

This PR (and the equivalent for [rosidl_typesupport](https://github.com/ros2/rosidl_typesupport/pull/168), [rosidl_typesupport_fastrtps](https://github.com/ros2/rosidl_typesupport_fastrtps/pull/136) and [rosidl_python](https://github.com/ros2/rosidl_python/pull/238)) adds the `DEPENDS_EXPLICIT_ONLY` option of [add_custom_command](https://cmake.org/cmake/help/latest/command/add_custom_command.html) to most generators (except for C/CPP).

> [!note]
> This option is only useful when using Ninja. When testing use `--cmake-args -GNinja`

Since the generators are no longer constrained by unnecessary dependencies, they can start immediately. This speeds up the build significantly (2x for px4_msgs on my machine).

| Package | Now | rolling |
|--------|--------|--------|
| px4_msgs | ninja_px4_pr | ninja_px4_rolling |
| test_msgs | ninja_test_pr | ninja_test_rolling | 

> [!warning]
> DEPENDS_EXPLICIT_ONLY requires cmake 3.27. The version that ships with Ubuntu 24.04 is 3.28, not sure about other supported distributions.
> ~~Right now I've bumped the required version to 3.27 but I could instead only add the option when the cmake version is sufficient (if that's preferable ?).~~

> [!note]
> I wanted to also add the option to rosidl_generator_c[pp] and explicitly specify the dependency to rosidl_generator_type_description. I couldn't figure out how to make it work. 

no

### Additional Information

I've already created PRs for the other packages. I've set then to Draft until this PR is reviewed.
- https://github.com/ros2/rosidl_typesupport/pull/168
- https://github.com/ros2/rosidl_typesupport_fastrtps/pull/136
- https://github.com/ros2/rosidl_python/pull/238


InvincibleRMC:
Looking at `rhel9`  it only supports `3.26.5`. The upcoming `rhel10` is on `3.30.5`. So you would probably have to make it conditional or wait for the switch to `rhel10`.

Refs: 
[rhel9](https://almalinux.pkgs.org/9/almalinux-appstream-x86_64/cmake-3.26.5-2.el9.x86_64.rpm.html)
[rhel10](
https://almalinux.pkgs.org/10/almalinux-appstream-x86_64/cmake-3.30.5-2.el10.x86_64.rpm.html)

TonyWelte:
I've updated the PR and tested on rhel9 and Ubuntu 24.04. The options is only used on Ubuntu (with cmake 3.28), on rhel9 the behavior is unchanged.

fujitatomoya (approved):
lgtm. this will track only the listed dependencies in DEPENDS list, so that i can reduce possible spurious rebuilds due to inferred dependencies.

@InvincibleRMC @gavanderhoorn WDYT?

InvincibleRMC:
I'm no expert of cmake but from my understanding this looks good.

rosidl_generator_type_description/cmake/rosidl_generator_type_description_generate_interfaces.cmake
```diff
@@ -88,6 +94,7 @@ add_custom_command(
   DEPENDS ${target_dependencies}
   COMMENT "Generating type hashes for ROS interfaces"
   VERBATIM
+  ${_dep_explicit_only}
```
gavanderhoorn, line 97: have you considered using a CMake generator expression here?

TonyWelte: I actually did but it doesn't work. I tried setting it to `$<$<VERSION_GREATER_EQUAL:${CMAKE_VERSION},3.27>:DEPENDS_EXPLICIT_ONLY>` but it's treated as just another argument to the previous option. Also tried to use it to set the variable (`set(_dep_explicit_only "$<$<VERSION_GREATER_EQUAL:${CMAKE_VERSION},3.27>:DEPENDS_EXPLICIT_ONLY>")`) but it also didn't work.

gavanderhoorn: yeah, I missed this is an actual argument keyword, not a value. `/ignore`.

TonyWelte:
I noticed this PR hasn’t had activity in a while. Could you let me know if you have any feedback or if there’s anything I can do to help move it forward?

sloretz:
@mergifyio update

sloretz:
If I understand correctly, This looks great to me!

> I wanted to also add the option to rosidl_generator_c[pp] and explicitly specify the dependency to rosidl_generator_type_description. I couldn't figure out how to make it work.

That would be great to explicitly specify the dependency. I think we can merge these PRs first. Afterwards, would you be willing to open a draft showing what you tried?

sloretz (approved):

sloretz:
Pulls: https://github.com/ros2/rosidl/pull/910, https://github.com/ros2/rosidl_typesupport/pull/168, https://github.com/ros2/rosidl_typesupport_fastrtps/pull/136, https://github.com/ros2/rosidl_python/pull/238
Gist: https://gist.githubusercontent.com/sloretz/f923f1982b5ea7f83805e42acdcbf5d8/raw/e3f7d1598fdf6a30f5f3c0e9ad8a7a595659c6a2/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17975
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27087/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20482/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6859/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26436/) (Flaky tests, I see same failures in https://ci.ros2.org/view/nightly/job/nightly_win_rel/3675/#showFailuresLink )

sloretz:
Thank you for the PR!

Issue #923, opened 2025-12-12
[rosidl_generator_cpp] Adding member arity and field names to generated structures


oysstu:
### Description

I propose adding two members to the generated structure for C++. Naming of the members are just suggestions. The first member can strictly speaking be accessed through the .size() method of the second member.

```
static constexpr size_t _member_arity = X;
static constexpr std::array<std::string_view, _member_arity> _member_names = {"member1", "member2", ...};
```

### Motivation

The motivation for this is primarily to allow for some generic programming. Since C++17 it's possible to use structural bindings to decompose a structure into its individual members. E.g., 
```c++
builtin_interfaces::msg::Time time;
auto [sec, nanosec] = time;
```
To use this generically the number of members must be known at compile time. Before C++26 reflection the only way to do this is to test for how many members it can be initialized using brace initialization. This does not work for these structures due to the custom constructor.

Printing messages to console and generic implementations of the yaml functions in the struct__traits.hpp header are possible using this addition.

### Design / Implementation Considerations

I can make a PR if this sounds good. It's fairly small changes to the .em templates.

### Additional Information

The constructors probably seemed like a good idea at the time, but the message structs would be more flexible as aggregate types. Changing it now would break a lot of existing code.

kscottz:
Hi @oysstu, we took a look at this at the weekly Waffle meeting. We think it is a great idea but we don't have the bandwidth to do it ourselves for Lyric. If you are willing to send us a PR we would be happy to review it!

oysstu:
That's great, I'll have some free time to make a PR next week. For the record, it's possible to get the field names through the [introspection API](https://github.com/ros2/rosidl/blob/b9aebb32c317bbf8180ebbf8478f60e2bc98d6ba/rosidl_typesupport_introspection_cpp/include/rosidl_typesupport_introspection_cpp/message_introspection.hpp#L90-L110). There's two blockers for using this for generic programming; it's not available in a constexpr context and there's no way to get the structure from a message template type parameter (as far as I can tell).

Are there any other fields in the typesupport structure(s) that are convenient to have available directly from the message type?

oysstu:
Implemented in #928

PR #928, opened 2026-01-05
rosidl_generator_cpp: constexpr message traits and to_tuple_ref for generated structs

oysstu:
## Description

This PR adds a constexpr-capable MemberTraits structure such that the number of members and member names can be used in a compile-time context.  The PR also adds a function that forwards the members of the struct as a tuple of references. This makes it very simple to treat the members generically, using for example std::apply.

Implements #923 

It should not, as the new struct / function is contained within the appropriate namespaces.

No

### DIFF
--- a/rosidl_generator_cpp/resource/idl__traits.hpp.em
+++ b/rosidl_generator_cpp/resource/idl__traits.hpp.em
@@ -29,9 +29,14 @@ include_directives = set()
 
 #include <stdint.h>
 
+#include <array>
+#include <cstddef>
 #include <sstream>
 #include <string>
+#include <string_view>
+#include <tuple>
 #include <type_traits>
+#include <utility>
 
 #include "@(include_base)__struct.hpp"
 #include "rosidl_runtime_cpp/traits.hpp"
--- a/rosidl_generator_cpp/resource/msg__traits.hpp.em
+++ b/rosidl_generator_cpp/resource/msg__traits.hpp.em
@@ -214,6 +214,19 @@ inline std::string to_yaml(const @(message.structure.namespaced_type.name) & msg
   }
   return out.str();
 }
+
+template<typename T, std::enable_if_t<std::is_same_v<std::decay_t<T>, @(message_typename)>, int> = 0>
+constexpr auto as_tuple_ref(T&& msg) {
+@[if len(message.structure.members) == 0]@
+  return std::forward_as_tuple();
+@[elif len(message.structure.members) == 1]@
+  return std::forward_as_tuple(std::forward<T>(msg).@(message.structure.members[0].name));
+@[else]@
+  return std::forward_as_tuple(
+@{forward_args = [f'    std::forward<T>(msg).{member.name}' for member in message.structure.members]}@
+@(',\n'.join(forward_args)));
+@[end if]@
+}
 @[for ns in reversed(message.structure.namespaced_type.namespaces)]@
 
 }  // namespace @(ns)
@@ -285,5 +298,15 @@ struct has_bounded_size<@(message_typename)>
 template<>
 struct is_message<@(message_typename)>
   : std::true_type {};
+
+template<>
+struct MessageTraits<@(message_typename)> {
+  static constexpr std::size_t member_count = @(len(message.structure.members));
+  static constexpr std::array<std::string_view, member_count> member_names = {
+@[for member in message.structure.members]@
+    "@(member.name)",
+@[end for]@
+  };
+};
 
 }  // namespace rosidl_generator_traits
--- a/rosidl_generator_cpp/resource/msg__traits.hpp.em
+++ b/rosidl_generator_cpp/resource/msg__traits.hpp.em
@@ -216,7 +216,8 @@ inline std::string to_yaml(const @(message.structure.namespaced_type.name) & msg
 }
 
 template<typename T, std::enable_if_t<std::is_same_v<std::decay_t<T>, @(message_typename)>, int> = 0>
-constexpr auto as_tuple_ref(T&& msg) {
+constexpr auto as_tuple_ref(T && msg)
+{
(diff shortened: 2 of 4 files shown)


fujitatomoya (commented):
i think this is useful and non-breaking change, lgtm with green CI.

i also think that alternative approach would be traits class (like rosidl_generator_traits<T>) rather than adding members to the message itself. this keeps the message "cleaner" but requires more boilerplate for users.

a couple of comments,

(not blocking this PR) Does this also apply to service request/response structs and action types? The issue only mentions messages.
Users won't know this feature exists without documentation. Should be mentioned in the rosidl_generator_cpp README or official documentation?

oysstu:
> i also think that alternative approach would be traits class (like rosidl_generator_traits) rather than adding members to the message itself. this keeps the message "cleaner" but requires more boilerplate for users.

Yes, good point. What about something like the following in `msg__traits.hpp.em`?

```c++
namespace rosidl_generator_traits
{

template<>
struct MessageTraits<rosidl_generator_tests::msg::Defaults> {
  static constexpr std::size_t member_count = 13;
  static constexpr std::array<std::string_view, member_count> member_names = {"bool_value", ...};
}

}
```

I also propose the following addition, as this function greatly reduces the amount of boilerplate needed to do generic programming over the structs. Without it, one must implement a structured binding to tuple function for each possible number of members. It works well, but a maximum member count has to be decided on, which is avoided by defining the function below.
```c++
namespace rosidl_generator_traits
{

template<typename T, std::enable_if_t<std::is_same_v<std::decay_t<T>, rosidl_generator_tests::msg::Defaults>, int> = 0>
constexpr auto as_tuple_ref(T&& msg) {
    return std::forward_as_tuple(std::forward<T>(msg).bool_value, ..., std::forward<T>(msg).uint64_value);
}

}
```

> 
> Does this also apply to service request/response structs and action types? The issue only mentions messages.

When defined in the `msg__struct.hpp.em` or `msg__traits.hpp.em`  templates, yes, because that is used to generate the definitions for the underlying Response/Request messages. One would use the is_service/is_action traits and then access the underlying message structs.

> Users won't know this feature exists without documentation. Should be mentioned in the rosidl_generator_cpp README or official documentation?

Yes, should probably add some documentation once we agree on the design.

kscottz:
As per Waffle meeting this needs to go to the PMC.

oysstu:
Great, I've updated the example above to use forward_as_tuple, in order to be more flexible. I will convert this to a draft, and revise the implementation after the PMC has looked at it.

asymingt:
When this was raised in last week's PMC the consensus seemed to be that the change looks really useful. Thanks for putting up this PR.  Looks like all checks pass, and @fujitatomoya has already given feedback. So, I'm happy to mark it LGTM. Since `rosidl_generator_cpp` doesn't have any documentation, can we put something on [this page](https://docs.ros.org/en/rolling/Concepts/Advanced/About-Internal-Interfaces.html) making reference to the traits? ie. update the page [here](https://github.com/ros2/ros2_documentation/blob/rolling/source/Concepts/Advanced/About-Internal-Interfaces.rst) with an accompanying PR.

oysstu:
@asymingt I marked it as draft, because this PR does not yet contain the changes proposed by @fujitatomoya. I will update the PR and request another review.

oysstu:
Ok, I've updated the PR with the proposed changes, please take a look.

 I was hoping to make a PR to follow up this one to change the message struct constructors to be constexpr, but that is not possible until C++20 due to the if/else.

ros-discourse:
This pull request has been mentioned on **Open Robotics Discourse**. There might be relevant details there:

https://discourse.openrobotics.org/t/ros-pmc-minutes-for-january-27-2026/52151/1

asymingt:
> i think this is useful and non-breaking change, lgtm with green CI.
> 
> i also think that alternative approach would be traits class (like rosidl_generator_traits) rather than adding members to the message itself. this keeps the message "cleaner" but requires more boilerplate for users.
> 
> a couple of comments,
> 
> (not blocking this PR) Does this also apply to service request/response structs and action types? The issue only mentions messages. Users won't know this feature exists without documentation. Should be mentioned in the rosidl_generator_cpp README or official documentation?

I believe `srv_traits` [internally calls](https://github.com/oysstu/rosidl/blob/03a07f1e307a0b53213bdf26132585cfec45d735/rosidl_generator_cpp/resource/srv__traits.hpp.em#L3-L21) the `msg_traits`, and so we receive the benefit on the `Request` and `Response` sub-messages. A similar thing happens for action `Goals`, `Feedback` and `Result` [here](https://github.com/oysstu/rosidl/blob/03a07f1e307a0b53213bdf26132585cfec45d735/rosidl_generator_cpp/resource/action__traits.hpp.em#L11-L29).

asymingt (approved):
LGTM s.t a green build. Let's please update the documentation (PR against a separate repo) per my earlier [comment](https://github.com/ros2/rosidl/pull/928#issuecomment-3801835832).

asymingt:
Had some CI issues in the last 48 hours. Here's the re-trigger!

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27384/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20596/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7308/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26547/)

asymingt:
@fujitatomoya, I'm going to hold of merging until you approve the changes you requested. Feel free to merge, if the CI above all goes 🟢 🟢 🟢 🟢

fujitatomoya (approved):

fujitatomoya:
@asymingt thanks for reviewing this!

@oysstu thanks for the PR!

PR #938, opened 2026-03-10
Add rosidl_buffer for native Buffer type support

nvcyc:
## Description

This pull request adds the `rosidl_buffer` package - core C++ buffer types for the ROS 2 native buffer feature. This package introduces `rosidl::Buffer<T>`, a polymorphic container that's designed to replace `std::vector<T>` for `uint8[]` message fields (to be adopted in later pull requests; we focus on only the core buffer types in this pull request.) The native buffer type enables vendor-specific memory backends (CUDA, ROCm, etc.) while maintaining backward compatibility for existing CPU-based `std::vector<T>` user code.

This pull request consists of the following key components:
- **`Buffer<T>`**: PIMPL-based container providing `std::vector<T>`-compatible API. All vector-compatible operations (element access, iterators, modifiers) are CPU-only and throw `std::runtime_error` for non-CPU backends. Backend management APIs (`get_backend_type()`, `set_impl()`, `get_impl()`, `to_vector()`) and copy/move semantics work for all backends.
- **`BufferImplBase<T>`**: Minimal abstract base class of buffer implementation. All backend-specific APIs are to be provided by the vendor-specific backend implementations.
- **`CpuBufferImpl<T>`**: CPU-based buffer implementation wrapping `std::vector<T>`.

This pull request does not change existing behavior.
When adopted in later pull requests, message types with `uint8[]` fields (e.g., `sensor_msgs/msg/Image.data`) will use `rosidl::Buffer<uint8_t>` instead of `std::vector<uint8_t>`. For CPU backends (the default), `Buffer<T>` is a transparent drop-in replacement.

Yes. Claude (claude-4.6-opus) via Cursor was used to assist with the `std::vector<T>` compatibility feature in the `rosidl::Buffer<T>` class and generate portions of the unit tests.

### Additional Information

This package/PL is part of the broader ROS2 native buffer feature introduced in this [post](https://discourse.openrobotics.org/t/working-prototype-of-native-buffers-accelerated-memory-transport/52399).

### DIFF
--- a/rosidl_buffer/include/rosidl_buffer/buffer.hpp
+++ b/rosidl_buffer/include/rosidl_buffer/buffer.hpp
@@ -1,0 +1,428 @@
+// Copyright 2026 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ROSIDL_BUFFER__BUFFER_HPP_
+#define ROSIDL_BUFFER__BUFFER_HPP_
+
+#include <algorithm>
+#include <initializer_list>
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+#include "rosidl_buffer/buffer_impl_base.hpp"
+#include "rosidl_buffer/cpu_buffer_impl.hpp"
+
+namespace rosidl
+{
+
+/// Buffer<T> provides a drop-in replacement for std::vector<T> with support
+/// for vendor-specific memory backends (CPU, GPU, etc.).
+///
+/// For CPU backends, it provides implicit conversion to std::vector<T>&
+/// for seamless backward compatibility. For non-CPU backends, direct access
+/// throws exceptions, requiring explicit conversion via to_vector().
+template<typename T, typename Allocator = std::allocator<T>>
+class Buffer
+{
+public:
+  // Type aliases for std::vector compatibility
+  using value_type = T;
+  using allocator_type = Allocator;
+  using size_type = size_t;
+  using difference_type = std::ptrdiff_t;
+  using reference = T &;
+  using const_reference = const T &;
+  using pointer = T *;
+  using const_pointer = const T *;
+  using iterator = T *;
+  using const_iterator = const T *;
+
+  /// Default constructor creates CPU buffer
+  Buffer()
+  : backend_type_("cpu"),
+    impl_(std::make_unique<CpuBufferImpl<T>>())
+  {
+  }
+
+  /// Construct with initial size (CPU backend)
+  explicit Buffer(size_t count)
+  : backend_type_("cpu"),
+    impl_(std::make_unique<CpuBufferImpl<T>>())
+  {
+    get_cpu_impl()->get_storage().resize(count);
+  }
+
+  /// Construct with initial size and value (CPU backend)
+  Buffer(size_t count, const T & value)
+  : backend_type_("cpu"),
+    impl_(std::make_unique<CpuBufferImpl<T>>())
+  {
+    get_cpu_impl()->get_storage().assign(count, value);
+  }
+
+  /// Construct from std::vector (copy) - for backward compatibility
+  Buffer(const std::vector<T> & vec)  // NOLINT(runtime/explicit) - intentionally implicit
+  : backend_type_("cpu"),
+    impl_(std::make_unique<CpuBufferImpl<T>>())
+  {
+    get_cpu_impl()->get_storage() = vec;
+  }
+
+  /// Construct from std::vector (move) - for backward compatibility
+  Buffer(std::vector<T> && vec)  // NOLINT(runtime/explicit) - intentionally implicit
+  : backend_type_("cpu"),
+    impl_(std::make_unique<CpuBufferImpl<T>>())
+  {
+    get_cpu_impl()->get_storage() = std::move(vec);
+  }
+
+  /// Construct from initializer list - for backward compatibility
+  Buffer(std::initializer_list<T> init)
+  : backend_type_("cpu"),
+    impl_(std::make_unique<CpuBufferImpl<T>>())
+  {
+    get_cpu_impl()->get_storage() = init;
+  }
+
+  /// Copy constructor (deep copy via clone())
+  Buffer(const Buffer & other)
+  : backend_type_(other.backend_type_),
+    impl_(other.impl_ ? other.impl_->clone() : nullptr)
+  {
+  }
+
+  /// Move constructor
+  Buffer(Buffer && other) noexcept
+  : backend_type_(std::move(other.backend_type_)),
+    impl_(std::move(other.impl_))
+  {
+  }
+
+  /// Copy assignment (deep copy via clone())
+  Buffer & operator=(const Buffer & other)
+  {
+    if (this != &other) {
+      backend_type_ = other.backend_type_;
+      impl_ = other.impl_ ? other.impl_->clone() : nullptr;
+    }
+    return *this;
+  }
+
+  /// Move assignment
+  Buffer & operator=(Buffer && other) noexcept
+  {
+    if (this != &other) {
+      backend_type_ = std::move(other.backend_type_);
+      impl_ = std::move(other.impl_);
+    }
+    return *this;
+  }
+
+  /// Assignment from initializer list - for backward compatibility
+  /// This must come before vector assignment to resolve ambiguity with {{...}} syntax
+  Buffer & operator=(std::initializer_list<T> init)
+  {
+    backend_type_ = "cpu";
+    impl_ = std::make_unique<CpuBufferImpl<T>>();
+    get_cpu_impl()->get_storage() = init;
+    return *this;
+  }
+
+  /// Assignment from std::vector (copy) - for backward compatibility
+  /// Uses SFINAE to avoid ambiguity with initializer lists
+  template<typename U = std::vector<T>,
+    typename std::enable_if<std::is_same<U, std::vector<T>>::value, int>::type = 0>
+  Buffer & operator=(const U & vec)
+  {
+    backend_type_ = "cpu";
+    impl_ = std::make_unique<CpuBufferImpl<T>>();
+    get_cpu_impl()->get_storage() = vec;
+    return *this;
+  }
+
+  /// Assignment from std::vector (move) - for backward compatibility
+  /// Uses SFINAE to avoid ambiguity with initializer lists
+  template<typename U = std::vector<T>,
+    typename std::enable_if<std::is_same<U, std::vector<T>>::value, int>::type = 0>
+  Buffer & operator=(U && vec)
+  {
+    backend_type_ = "cpu";
+    impl_ = std::make_unique<CpuBufferImpl<T>>();
+    get_cpu_impl()->get_storage() = std::move(vec);
+    return *this;
+  }
+
+  // ========== Element Access (CPU only) ==========
+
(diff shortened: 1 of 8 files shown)


rosidl_buffer/test/test_buffer.cpp
```diff
@@ -0,0 +1,488 @@
...
+#include "rosidl_buffer/buffer_impl_base.hpp"
+#include "rosidl_buffer/cpu_buffer_impl.hpp"
+
+using rosidl::Buffer;
+using rosidl::BufferImplBase;
+using rosidl::CpuBufferImpl;
+
+// Test default construction
+TEST(TestBuffer, default_construction) {
+  Buffer<int> buffer;
```
MiguelCompany, line 32: Since the goal of the feature is to handle buffers of `uint8_t`, I think it makes sense to test with that type instead of `int`.

rosidl_buffer/test/test_buffer.cpp
```diff
@@ -0,0 +1,488 @@
...
+
+// Test copy construction
+TEST(TestBuffer, copy_construction) {
+  Buffer<int> buffer1(3, 100);
+  Buffer<int> buffer2(buffer1);
+
+  EXPECT_EQ(buffer1.size(), buffer2.size());
+  EXPECT_EQ(buffer1.get_backend_type(), buffer2.get_backend_type());
+  for (size_t i = 0; i < buffer1.size(); ++i) {
+    EXPECT_EQ(buffer1[i], buffer2[i]);
```
MiguelCompany, line 63: Shall we check that they are in different addresses?


```suggestion
    EXPECT_EQ(buffer1[i], buffer2[i]);
    EXPECT_NE(&buffer1[i], &buffer2[i]);
```

rosidl_buffer/test/test_buffer.cpp
```diff
@@ -0,0 +1,488 @@
...
+
+// Test copy assignment
+TEST(TestBuffer, copy_assignment) {
+  Buffer<int> buffer1(3, 100);
+  Buffer<int> buffer2;
+  buffer2 = buffer1;
+
+  EXPECT_EQ(buffer1.size(), buffer2.size());
+  for (size_t i = 0; i < buffer1.size(); ++i) {
+    EXPECT_EQ(buffer1[i], buffer2[i]);
```
MiguelCompany, line 84: Same here

rosidl_buffer/test/test_buffer.cpp
```diff
@@ -0,0 +1,488 @@
...
+  buffer.push_back({4.0, 5.0, 6.0});
+
+  EXPECT_EQ(2u, buffer.size());
+  EXPECT_DOUBLE_EQ(1.0, buffer[0].x);
+  EXPECT_DOUBLE_EQ(5.0, buffer[1].y);
+
+  std::vector<Point> copied = buffer.to_vector();
+  EXPECT_DOUBLE_EQ(3.0, copied[0].z);
+}
+
```
MiguelCompany, line 483: Add some tests that check `std::runtime_exception` is thrown for certain operations when the implementation is not `cpu`

rosidl_buffer/src/c_helpers.cpp
MiguelCompany, line 1: The functions in this file shall be checked in tests.

MiguelCompany (changes requested):

rosidl_buffer/include/rosidl_buffer/buffer.hpp
```diff
@@ -0,0 +1,428 @@
...
+  /// implementations.
+  /// @param impl Backend implementation instance.
+  /// @param backend_type Backend identifier string.
+  void set_impl(
+    std::unique_ptr<BufferImplBase<T>> impl,
+    const std::string & backend_type)
+  {
+    impl_ = std::move(impl);
+    backend_type_ = backend_type;
+  }
```
MiguelCompany, line 397: If this is called with `"cpu"` as the backend type, we should check that the implementation can be cast to `CpuBufferImpl`


```suggestion
  void set_impl(
    std::unique_ptr<BufferImplBase<T>> impl,
    const std::string & backend_type)
  {
    if ((backend_type == "cpu") && (nullptr == dynamic_cast<CpuBufferImpl<T>*>(impl.get()))) {
      throw std::runtime_error("CPU backend shall inherit from CpuBufferImpl.");
    }

    impl_ = std::move(impl);
    backend_type_ = backend_type;
  }
```

MiguelCompany (changes requested):

rosidl_buffer/include/rosidl_buffer/buffer.hpp
hidmic: A follow up question would be, why store the backend identifier here instead of querying the underlying implementation for it? Why not let the underlying implementation decide what it does and doesn't support? You can follow that line of thought and say that `rosidl::Buffer` could simply forward calls to the implementation +/- CPU backend checks.

rosidl_buffer/include/rosidl_buffer/buffer.hpp
```diff
@@ -0,0 +1,428 @@
...
+/// for seamless backward compatibility. For non-CPU backends, direct access
+/// throws exceptions, requiring explicit conversion via to_vector().
+template<typename T, typename Allocator = std::allocator<T>>
+class Buffer
+{
+public:
+  // Type aliases for std::vector compatibility
+  using value_type = T;
+  using allocator_type = Allocator;
+  using size_type = size_t;
```
ahcorde, line 46: include `<cstddef>`

ahcorde:
Pulls: ros2/rosidl#938
Gist: https://gist.githubusercontent.com/ahcorde/189205809e28fadd8f0c809e165b7f09/raw/f345f50215a60fba5be8164ecd41c2d1ff9df91b/ros2.repos
BUILD args:  --packages-above-and-dependencies rosidl_buffer
TEST args:  --packages-above rosidl_buffer
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18438
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27985/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20968/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7842/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26940/)

rosidl_buffer/package.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>rosidl_buffer</name>
+  <version>1.0.0</version>
```
ahcorde, line 5: version should match with the current packages `5.1.2`

rosidl_buffer/package.xml
```diff
@@ -0,0 +1,24 @@
...
+  <description>
+    Core Buffer types and implementations for ROS2 native buffer feature.
+    Provides Buffer container type with support for multiple memory backends (CPU, GPU, custom).
+  </description>
+  
+  <maintainer email="cyc@nvidia.com">CY Chen</maintainer>
```
ahcorde, line 11: missing `<author>` tag

ahcorde (changes requested):

rosidl_buffer/include/rosidl_buffer/buffer.hpp
```diff
@@ -0,0 +1,428 @@
...
+
+  /// Get the backend type identifier (e.g., "cpu", "cuda").
+  std::string get_backend_type() const {return backend_type_;}
+
+  /// Set buffer implementation (for backend libraries).
+  /// This allows vendor-specific backend libraries to inject their
+  /// implementations.
+  /// @param impl Backend implementation instance.
+  /// @param backend_type Backend identifier string.
+  void set_impl(
```
hidmic, line 391: @nvcyc meta: hmm, have we considered passing implementation on buffer construction? Allowing swapping implementations on the fly creates race conditions in the other methods.

rosidl_buffer/include/rosidl_buffer/c_helpers.h
```diff
@@ -0,0 +1,50 @@
...
+const uint8_t * rosidl_buffer_uint8_data(const void * buffer_ptr);
+
+/// Get a mutable pointer to the underlying CPU data. Throws for non-CPU backends.
+/// @param buffer_ptr Opaque pointer to an rosidl::Buffer<uint8_t>
+uint8_t * rosidl_buffer_uint8_data_mut(void * buffer_ptr);
+
+/// Resize a Buffer<uint8_t>. Throws for non-CPU backends.
+/// @param buffer_ptr Opaque pointer to an rosidl::Buffer<uint8_t>
+/// @param size New size
+void rosidl_buffer_uint8_resize(void * buffer_ptr, size_t size);
```
hidmic, line 40: @nvcyc hmm, I would've expected a `rosidl_uint8_buffer_t`, even if it's just a type erased wrapper for `rosidl::Buffer<uint8_t>`. Mind to explain the use case?

hidmic (commented):
First pass, meta comments.

rosidl_buffer/include/rosidl_buffer/buffer.hpp
nvcyc: That is a valid suggestion. I'm updating the code so that the buffer implementation is the source of truth for the buffer backend type.

rosidl_buffer/test/test_buffer.cpp
nvcyc: Thanks for the suggestions. Tests are updated to test `uint8_t`.

rosidl_buffer/test/test_buffer.cpp
nvcyc: Added in new update.

rosidl_buffer/test/test_buffer.cpp
nvcyc: Added in new update.

rosidl_buffer/test/test_buffer.cpp
nvcyc: Added a test implementation class `NonCpuBufferImpl` and tests to check the throwing exception cases.

rosidl_buffer/src/c_helpers.cpp
nvcyc: Added `test_c_helpers.cpp` for testing functions in `c_helpers.cpp`.

rosidl_buffer/include/rosidl_buffer/buffer.hpp
nvcyc: Added in new update.

rosidl_buffer/package.xml
nvcyc: Fixed in new update.

rosidl_buffer/package.xml
nvcyc: Added in new update.

rosidl_buffer/include/rosidl_buffer/buffer.hpp
nvcyc: You are right and this is also not allowed by my initial design. Removing this function to keep it clean and aligned.

rosidl_buffer/include/rosidl_buffer/c_helpers.h
nvcyc: Thanks for the comment.
These C helpers are mainly for the C introspection typesupport path (in later PLs) but also they are not intended to be used against non-CPU-based buffers (and hence we throw exception if accessed with these APIs). In C path, the current design is that CPU-based uint8[] will always be the data field in `rosidl_runtime_c__uint8__Sequence` as it originally does, so we won't have the case where CPU-based utin8[] is stored in rosidl::Buffer in the C path. 

For this reason, I think we actually don't need these introspection C helper functions here as they are not supported by backend buffers. I'm optimizing from the C introspection typesupport side to simply call `rosidl_buffer_uint8_throw_if_not_cpu()` as guards there.

I'll make the changes to keep them simple in my new update.

nvcyc:
Thanks for your review.

To facilitate the review process with cleaner future PR dependencies, I'm creating a new PR (#941) that's based on a branch in this repo (in contrast to a fork) to replace this one. All the changes for addressing the comments in this PR have been included in the new PR.

Please continue our review in #941 and I'll close this PR subsequently.

PR #940, opened 2026-03-12
Add @key annotation

### LINKED ISSUE #939: Support for `@key` in ROS Interface files
### Description

Currently keyed topics require the use of `IDL` messages.

However, ROS Interface files (`.msg`, `.srv` & `.action`) already support `@optional`.
Maybe this could be extended to include `@key` as well?

### Motivation

Simplify the use of keyed topics, without requiring the developer to create `.idl` message definitions. 

### Design / Implementation Considerations

Maybe the annotation system can be expanded in such a way it allows for more future additions easily.

### Additional Information

_No response_

SuperJappie08:
## Description

Add parsing for `@key` in `rosidl_adapter`.
This enables the use of keyed topics with `msg`, `srv` and `action` files.

Fixes #939 

Yes, users can now use keyed topics with `msg`, `srv` and `action` interface files instead of `idl`. 

No

### DIFF
--- a/rosidl_adapter/rosidl_adapter/parser.py
+++ b/rosidl_adapter/rosidl_adapter/parser.py
@@ -19,18 +19,25 @@ import textwrap
 from typing import Final, Iterable, List, Optional, Tuple, TYPE_CHECKING, TypedDict, Union
 
 if TYPE_CHECKING:
+    from typing import Dict, Literal
     from typing_extensions import TypeAlias
 
     PrimitiveType: TypeAlias = Union[bool, float, int, str]
+    AnnotationType: TypeAlias = Literal['optional', 'key']
+    AnnotantionsDict: TypeAlias = Dict[AnnotationType, bool]
 
     class Annotations(TypedDict, total=False):
         comment: List[str]
         unit: str
         optional: bool
+        key: bool
 
 PACKAGE_NAME_MESSAGE_TYPE_SEPARATOR: Final = '/'
 ANNOTATION_DELIMITER: Final = '@'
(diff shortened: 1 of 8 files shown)


asymingt:
Thanks for the contribution :) We've marked this PR for discussion in next week's ROS PMC meeting. We should be able to get back to you within the week.

asymingt:
This was [discussed](https://docs.google.com/document/d/1TGXYp33N00ZJtilsRzfClFXcSyMa4QKVNmA03Lx9D7w/edit?tab=t.0) in the PMC meeting on 2026-03-24. It was pointed out that support for this keyed topics Zenoh is unknown, plus [CycloneDDS](https://docs.ros.org/en/rolling/Tutorials/Advanced/Topic-Keys/Topic-Keys-Tutorial.html) for example doesn't appear to support keyed topics. Members wanted to ensure that we do our best not to leak DDS-specific features into the ROS domain, and if we choose to do so that we understand the consequence of our decision to do so! That being said, they weren't opposed to this PR, but they do want to put out a [REP](https://reps.openrobotics.org) to solicit feedback from the wider community. Is this something you'd be willing to lead?

SuperJappie08:
@asymingt, sorry for the delayed response.

I understand the concerns from the PMC.

When I made this PR I was not aware of the implications it would have to the ROS communication setup.
I just noticed that `@optional` was supported in the `rosidl` conversion, while `@key` was not, which felt off.

I agree creating a REP for this makes sense, however I don't think I have enough understanding of the RMW layer(s) and DDS mechanisms to be able to lead this effort.

PR #941, opened 2026-03-13
Add rosidl_buffer and rosidl_buffer_backend for native Buffer type support

nvcyc:
## Description

This pull request adds `rosidl_buffer` and `rosidl_buffer_backend` packages - core C++ buffer types for the ROS 2 native buffer feature. This package introduces `rosidl::Buffer<T>`, a polymorphic container that's designed to replace `std::vector<T>` for `uint8[]` message fields (to be adopted in later pull requests; we focus on only the core buffer types in this pull request.) The native buffer type and its backend class enable vendor-specific memory buffer implementation (CUDA, ROCm, etc.) while maintaining backward compatibility for existing CPU-based `std::vector<T>` user code.

This pull request consists of the following key components:
- **`Buffer<T>`**: PIMPL-based container providing `std::vector<T>`-compatible API. All vector-compatible operations (element access, iterators, modifiers) are CPU-only and throw `std::runtime_error` for non-CPU backends. Backend management APIs (`get_backend_type()`, `get_impl()`, `to_vector()`) work for all backends.
- **`BufferImplBase<T>`**: Minimal abstract base class of buffer implementation. All backend-specific APIs are to be provided by the vendor-specific backend implementations.
- **`CpuBufferImpl<T>`**: CPU-based buffer implementation wrapping `std::vector<T>`.
- **`BufferBackend`**: Abstract interface for vendor-specific buffer backend implementations. 

This pull request does not change existing behavior.
When adopted in later pull requests, message types with `uint8[]` fields (e.g., `sensor_msgs/msg/Image.data`) will use `rosidl::Buffer<uint8_t>` instead of `std::vector<uint8_t>`. For CPU backends (the default), `Buffer<T>` is a transparent drop-in replacement.

Yes. Claude (claude-4.6-opus) via Cursor was used to assist with the `std::vector<T>` compatibility feature in the `rosidl::Buffer<T>` class and generate portions of the unit tests.

### Additional Information

This package/PL is part of the broader ROS2 native buffer feature introduced in this [post](https://discourse.openrobotics.org/t/working-prototype-of-native-buffers-accelerated-memory-transport/52399).

### DIFF
--- a/rosidl_buffer/include/rosidl_buffer/buffer.hpp
+++ b/rosidl_buffer/include/rosidl_buffer/buffer.hpp
@@ -1,0 +1,531 @@
+// Copyright 2026 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ROSIDL_BUFFER__BUFFER_HPP_
+#define ROSIDL_BUFFER__BUFFER_HPP_
+
+#include <algorithm>
+#include <cstddef>
+#include <initializer_list>
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+#include "rosidl_buffer/buffer_impl_base.hpp"
+#include "rosidl_buffer/cpu_buffer_impl.hpp"
+
+namespace rosidl
+{
+
+/// Buffer<T> provides a drop-in replacement for std::vector<T> with support
+/// for vendor-specific memory backends (CPU, GPU, etc.).
+///
+/// For CPU backends, it provides implicit conversion to std::vector<T>&
+/// for seamless backward compatibility. For non-CPU backends, direct access
+/// throws exceptions, requiring explicit conversion via to_vector().
+template<typename T, typename Allocator = std::allocator<T>>
+class Buffer
+{
+public:
+  // Type aliases for std::vector compatibility
+  using value_type = T;
+  using allocator_type = Allocator;
+  using size_type = size_t;
+  using difference_type = std::ptrdiff_t;
+  using reference = T &;
+  using const_reference = const T &;
+  using pointer = T *;
+  using const_pointer = const T *;
+  using iterator = T *;
+  using const_iterator = const T *;
+
+  /// Default constructor creates CPU buffer
+  Buffer()
+  : impl_(std::make_unique<CpuBufferImpl<T>>())
+  {
+  }
+
+  /// Construct with a backend implementation (set once at construction).
+  /// This is the only way to set a non-CPU backend; there is no post-
+  /// construction setter, which avoids race conditions with concurrent reads.
+  explicit Buffer(std::unique_ptr<BufferImplBase<T>> impl)
+  : impl_(std::move(impl))
+  {
+    if (!impl_) {
+      throw std::invalid_argument("Buffer implementation must not be null");
+    }
+  }
+
+  /// Construct with initial size (CPU backend)
+  explicit Buffer(size_t count)
+  : impl_(std::make_unique<CpuBufferImpl<T>>())
+  {
+    get_cpu_impl()->get_storage().resize(count);
+  }
+
+  /// Construct with initial size and value (CPU backend)
+  Buffer(size_t count, const T & value)
+  : impl_(std::make_unique<CpuBufferImpl<T>>())
+  {
+    get_cpu_impl()->get_storage().assign(count, value);
+  }
+
+  /// Construct from std::vector (copy) - for backward compatibility
+  Buffer(const std::vector<T> & vec)  // NOLINT(runtime/explicit) - intentionally implicit
+  : impl_(std::make_unique<CpuBufferImpl<T>>())
+  {
+    get_cpu_impl()->get_storage() = vec;
+  }
+
+  /// Construct from std::vector (move) - for backward compatibility
+  Buffer(std::vector<T> && vec)  // NOLINT(runtime/explicit) - intentionally implicit
+  : impl_(std::make_unique<CpuBufferImpl<T>>())
+  {
+    get_cpu_impl()->get_storage() = std::move(vec);
+  }
+
+  /// Construct from initializer list - for backward compatibility
+  Buffer(std::initializer_list<T> init)
+  : impl_(std::make_unique<CpuBufferImpl<T>>())
+  {
+    get_cpu_impl()->get_storage() = init;
+  }
+
+  /// Copy constructor (deep copy via clone())
+  Buffer(const Buffer & other)
+  : impl_(other.impl_ ? other.impl_->clone() : nullptr)
+  {
+  }
+
+  /// Move constructor
+  Buffer(Buffer && other) noexcept
+  : impl_(std::move(other.impl_))
+  {
+  }
+
+  /// Copy assignment (deep copy via clone())
+  Buffer & operator=(const Buffer & other)
+  {
+    if (this != &other) {
+      impl_ = other.impl_ ? other.impl_->clone() : nullptr;
+    }
+    return *this;
+  }
+
+  /// Move assignment
+  Buffer & operator=(Buffer && other) noexcept
+  {
+    if (this != &other) {
+      impl_ = std::move(other.impl_);
+    }
+    return *this;
+  }
+
+  /// Assignment from initializer list - for backward compatibility
+  /// This must come before vector assignment to resolve ambiguity with {{...}} syntax
+  Buffer & operator=(std::initializer_list<T> init)
+  {
+    impl_ = std::make_unique<CpuBufferImpl<T>>();
+    get_cpu_impl()->get_storage() = init;
+    return *this;
+  }
+
+  /// Assignment from std::vector (copy) - for backward compatibility
+  /// Uses SFINAE to avoid ambiguity with initializer lists
+  template<typename U = std::vector<T>,
+    typename std::enable_if<std::is_same<U, std::vector<T>>::value, int>::type = 0>
+  Buffer & operator=(const U & vec)
+  {
+    impl_ = std::make_unique<CpuBufferImpl<T>>();
+    get_cpu_impl()->get_storage() = vec;
+    return *this;
+  }
+
+  /// Assignment from std::vector (move) - for backward compatibility
+  /// Uses SFINAE to avoid ambiguity with initializer lists
+  template<typename U = std::vector<T>,
+    typename std::enable_if<std::is_same<U, std::vector<T>>::value, int>::type = 0>
+  Buffer & operator=(U && vec)
+  {
+    impl_ = std::make_unique<CpuBufferImpl<T>>();
+    get_cpu_impl()->get_storage() = std::move(vec);
+    return *this;
+  }
+
+  // ========== Element Access (CPU only) ==========
+
+  /// Access element at position (CPU only)
+  reference operator[](size_t pos)
+  {
+    throw_if_not_cpu_backend();
+    return get_cpu_impl()->get_storage()[pos];
+  }
+
+  const_reference operator[](size_t pos) const
+  {
+    throw_if_not_cpu_backend();
+    return get_cpu_impl()->get_storage()[pos];
+  }
+
+  /// Access element with bounds checking (CPU only)
+  reference at(size_t pos)
+  {
+    throw_if_not_cpu_backend();
+    return get_cpu_impl()->get_storage().at(pos);
+  }
+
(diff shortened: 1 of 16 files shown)


rosidl_buffer/include/rosidl_buffer/buffer.hpp
```diff
@@ -0,0 +1,531 @@
...
+    }
+    return *this;
+  }
+
+  /// Assignment from initializer list - for backward compatibility
+  /// This must come before vector assignment to resolve ambiguity with {{...}} syntax
+  Buffer & operator=(std::initializer_list<T> init)
+  {
+    impl_ = std::make_unique<CpuBufferImpl<T>>();
+    get_cpu_impl()->get_storage() = init;
```
hidmic: @nvcyc do we need to recreate the CPU buffer? Isn't storage assignment enough?

rosidl_buffer/include/rosidl_buffer/buffer.hpp
```diff
@@ -0,0 +1,531 @@
...
+
+  /// Copy constructor (deep copy via clone())
+  Buffer(const Buffer & other)
+  : impl_(other.impl_ ? other.impl_->clone() : nullptr)
+  {
+  }
+
+  /// Move constructor
+  Buffer(Buffer && other) noexcept
+  : impl_(std::move(other.impl_))
```
hidmic: @nvcyc meta: I think we need to review move semantics. Either a null `impl_` means a null buffer and every API call throws, or `impl_` is never null but replaced by an empty CPU implementation to keep the buffer in a valid state. As it stands, I believe calling `push_back` on a moved buffer will SEGFAULT (because backend type defaults to CPU on a null `impl_`, so CPU backend checks pass).

Maybe the right answer is to delegate the move to the implementation, like we do for copy with `clone`. CPU buffers can internally move their vectors, which keeps the moved buffer valid and functional even if empty.

rosidl_buffer/include/rosidl_buffer/cpu_buffer_impl.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+  std::string get_backend_type() const override {return "cpu";}
+
+  size_t size() const override {return storage_.size();}
+
+  std::unique_ptr<BufferImplBase<T>> to_cpu() const override
+  {
+    auto copy = std::make_unique<CpuBufferImpl<T>>();
+    copy->storage_ = storage_;
+    return copy;
```
hidmic: @nvcyc could this be:
```suggestion
    return clone();
```
?

rosidl_buffer/include/rosidl_buffer/buffer_impl_base.hpp
```diff
@@ -0,0 +1,58 @@
...
+/// CpuBufferImpl directly.
+template<typename T>
+class BufferImplBase
+{
+public:
+  virtual ~BufferImplBase() = default;
+
+  /// Get the backend type identifier (e.g., "cpu", "cuda", "demo").
+  /// Each concrete implementation returns its own fixed identifier.
+  virtual std::string get_backend_type() const = 0;
```
hidmic, line 40: @nvcyc should this be:
```suggestion
  virtual std::string_view get_backend_type() const = 0;
```
? Same for the backend interface.

hidmic: Also, I may be missing the whole picture, but why doesn't the implementation hold a reference to the backend? Where does the connection happen? How? 

Feel free to not answer and just paste a link to code.

hidmic (commented):
Amazing work @nvcyc!

rosidl_buffer/include/rosidl_buffer/buffer.hpp
hidmic: For reference on move semantics, I like [cppref](https://en.cppreference.com/w/cpp/language/move_assignment.html):

> _Move assignment operators typically transfer the resources held by the argument  [...], rather than make copies of them, and leave the argument in **some valid but otherwise indeterminate state**_.

rosidl_buffer/include/rosidl_buffer/buffer_impl_base.hpp
nvcyc: Thanks for the question!
We intentionally designed `rosidl_buffer`/`rosidl::Buffer` to be only a user-facing package/class and separated the backend part that provides ways to interact with typesupport/RMW layers.
This way, a backend provider can offer user-facing libraries for creating backend-backed buffers with cleaner dependency if desired.
Then, as you may have already discovered from other relevant PRs, the place where a buffer and its backend instance meet is when the buffer is about to be serialized/deserialized. You can refer to the following two places in `rosidl_typesupport_fastrtps` for where they become relevant:
- [serialize_buffer_with_endpoint()](https://github.com/ros2/rosidl_typesupport_fastrtps/blob/native_buffer/3-buffer_serialization/rosidl_typesupport_fastrtps_cpp/include/rosidl_typesupport_fastrtps_cpp/buffer_serialization.hpp#L132C13-L132C43)
- [deserialize_buffer_with_endpoint()](https://github.com/ros2/rosidl_typesupport_fastrtps/blob/native_buffer/3-buffer_serialization/rosidl_typesupport_fastrtps_cpp/include/rosidl_typesupport_fastrtps_cpp/buffer_serialization.hpp#L132C13-L132C43)

nvcyc: For the `string_view` case, most of our downstream code use `c_str()` and string concatenation with the obtained backend type string, so keeping it `std::string` would actually make sense I think.

rosidl_buffer/include/rosidl_buffer/buffer.hpp
nvcyc: You are correct. We don't need to recreate a CPU buffer here. Fixed in my update.

rosidl_buffer/include/rosidl_buffer/cpu_buffer_impl.hpp
nvcyc: Good suggestion! Updated.

rosidl_buffer/include/rosidl_buffer/buffer.hpp
nvcyc: Great catch, thanks! I went with the "impl_ is never null" approach: after a move, the moved-from buffer gets a fresh empty `CpuBufferImpl`, keeping it in a valid, empty, CPU-backed state. 

For delegating the move to the implementation, it can get complicated for cross-backend moves (e.g., moving a CUDA buffer impl into a CPU buffer impl), so I think the aforementioned approach (unique_ptr move + replacement with an empty CPU impl) is simple enough and achieves the cppref "valid but otherwise indeterminate state" contract.

rosidl_buffer/include/rosidl_buffer/buffer_impl_base.hpp
hidmic: Fair enough.

hidmic (approved):
Looks correct.

rosidl_buffer/include/rosidl_buffer/buffer.hpp
```diff
@@ -0,0 +1,533 @@
...
+  using reference = T &;
+  using const_reference = const T &;
+  using pointer = T *;
+  using const_pointer = const T *;
+  using iterator = T *;
+  using const_iterator = const T *;
+
+  /// Default constructor creates CPU buffer
+  Buffer()
+  : impl_(std::make_unique<CpuBufferImpl<T>>())
```
MiguelCompany: Shall we pass the allocator to the CpuBufferImpl ?

MiguelCompany: Same for all CPU constructors

rosidl_buffer/include/rosidl_buffer/buffer.hpp
```diff
@@ -0,0 +1,533 @@
...
+  /// Get the implementation pointer (for serialization) - returns raw pointer
+  /// for read-only access
+  const BufferImplBase<T> * get_impl() const {return impl_.get();}
+
+  /// Throw exception if not CPU backend.
+  /// @throws std::runtime_error if backend is not CPU.
+  void throw_if_not_cpu_backend() const
+  {
+    const std::string bt = get_backend_type();
+    if (bt != "cpu") {
```
MiguelCompany: What will happen if the buffer is constructed with a backend that returns `"cpu"` as the backend type, but does not inherit from `CpuBufferImpl` ?

rosidl_buffer/include/rosidl_buffer/buffer.hpp
```diff
@@ -0,0 +1,533 @@
...
+      throw std::runtime_error(
+              "Operation requires CPU backend. Current backend: " + bt +
+              ". Use to_vector() for explicit conversion to CPU.");
+    }
+  }
+
+private:
+  /// Unique pointer for proper ownership and value semantics.
+  /// The implementation is the sole source of truth for the backend type.
+  std::unique_ptr<BufferImplBase<T>> impl_;
```
MiguelCompany, line 499: Perhaps we could have something like the following:

```suggestion
  /// Unique pointer for proper ownership and value semantics.
  /// The implementation is the sole source of truth for the backend type.
  std::unique_ptr<BufferImplBase<T>> impl_;
  
  /// Used to check if the backend is a CPU one
  CpuBufferImpl<T> * cpu_impl_ = nullptr;
  
  // Used whenever we want to set the value of `impl_`
  void set_impl(std::unique_ptr<BufferImplBase<T>> impl) {
    impl_ = std::move(impl);
    cpu_impl_ = dynamic_cast<CpuBufferImpl<T> *>(impl_.get());
  }
```

And then use `if (nullptr == cpu_impl_)` in `throw_if_not_cpu_backend()`, and change `get_cpu_impl()` to just `return cpu_impl_;`

MiguelCompany (changes requested):

rosidl_buffer/include/rosidl_buffer/buffer.hpp
nvcyc: Thanks for bringing this up. Yes, I think the allocator should be passed to `CpuBufferImpl`.

Updated the PR with a new commit: https://github.com/ros2/rosidl/pull/941/changes/326b4f8bd992d847cd57b84c0c5aa1c4af0e3d87

rosidl_buffer/include/rosidl_buffer/buffer.hpp
nvcyc: Good idea! I've included your suggested changes in a new commit pushed to this PR: https://github.com/ros2/rosidl/pull/941/changes/326b4f8bd992d847cd57b84c0c5aa1c4af0e3d87

rosidl_buffer/include/rosidl_buffer/buffer.hpp
nvcyc: That is indeed a case to be taken care of.
The solution you suggested in the response below using a `CpuBufferImpl` pointer looks good to me.
I've pushed a commit to this PR for addressing this issue: https://github.com/ros2/rosidl/pull/941/changes/326b4f8bd992d847cd57b84c0c5aa1c4af0e3d87

MiguelCompany (approved):
LGTM

ahcorde (approved):

ahcorde:
Pulls: ros2/rosidl#941
Gist: https://gist.githubusercontent.com/ahcorde/d14f1064681a43ab6af3d7e1c2bbd285/raw/17346b358df7d0e4cab5d800a790e46b6c26dea0/ros2.repos
BUILD args:  --packages-above-and-dependencies rosidl_buffer rosidl_buffer_backend
TEST args:  --packages-above rosidl_buffer rosidl_buffer_backend
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18681
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28355/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21231/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8215/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27233/)

nvcyc:
Pulls: ros2/rosidl#941
Gist: https://gist.githubusercontent.com/nvcyc/874d4eb64eb8543bcd349cd9e355a788/raw/17346b358df7d0e4cab5d800a790e46b6c26dea0/ros2.repos
BUILD args: --continue-on-error --packages-above-and-dependencies rosidl_buffer rosidl_buffer_backend
TEST args:  --packages-above rosidl_buffer rosidl_buffer_backend
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18701
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28375/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21251/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8235/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27253/)

hidmic (approved):

nvcyc:
Re-triggered the CI just for the Windows platform:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27259/)

mjcarroll (approved):
LGTM!

PR #944, opened 2026-03-16
Add rosidl_buffer_backend_registry

nvcyc:
## Description

This pull request adds the `rosidl_buffer_backend_registry` package — a singleton registry using pluginlib to discover and load `BufferBackend` plugins at runtime. This package consolidates buffer backend initialization logic that RMW implementations need during `rmw_init`/`rmw_shutdown`, eliminating the need for each RMW to implement its own backend loading.

This pull request consists of the following key changes:

- **Buffer backend registry** (`buffer_backend_registry.hpp/cpp`): Singleton that uses pluginlib to discover all installed `BufferBackend` plugin classes and manage their lifecycle.
- **Buffer backend loader** (`buffer_backend_loader.hpp/cpp`): Provides `initialize_buffer_backends()` and `shutdown_buffer_backends()` — consolidated entry points that load backend plugins, populate the global `BufferDescriptorOps` and `DescriptorSerializers` maps in `rosidl_typesupport_fastrtps_cpp`, and clean up on shutdown. RMW implementations call these during `rmw_init`/`rmw_shutdown` instead of maintaining their own copies.

This pull request does not change existing behavior. It provides infrastructure for loading buffer backend plugins, which has no effect until an RMW implementation calls the initialization functions and a backend plugin is installed.

Yes. Claude (claude-4.6-opus) via Cursor was used to assist with creating an initial prototype version of the changes contained in this PR.

### Additional Information

This PR is part of the broader ROS 2 native buffer feature introduced in this [post](https://discourse.openrobotics.org/t/working-prototype-of-native-buffers-accelerated-memory-transport/52399).


rosidl_buffer_backend_registry/include/rosidl_buffer_backend_registry/buffer_backend_loader.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include "rosidl_buffer_backend_registry/visibility_control.h"
+
+namespace rosidl_buffer_backend_registry
+{
+
+/// Load buffer backend plugins and register them with FastCDR serialization.
+/// Populates the global BackendDescriptorOps and DescriptorSerializers maps
+/// in rosidl_typesupport_fastrtps_cpp so that buffer-aware message types can
+/// serialize/deserialize vendor-specific descriptors.
```
hidmic: @nvcyc why are there references to FastRTPS in here?

rosidl_buffer_backend_registry/include/rosidl_buffer_backend_registry/buffer_backend_registry.hpp
```diff
@@ -0,0 +1,132 @@
...
+  ROSIDL_BUFFER_BACKEND_REGISTRY_PUBLIC
+  std::vector<std::string> get_backend_names() const;
+
+  /// Get backend type strings for all registered backends, always including "cpu".
+  /// Each backend's get_backend_type() is queried (e.g., "cuda", "rocm").
+  ROSIDL_BUFFER_BACKEND_REGISTRY_PUBLIC
+  std::vector<std::string> get_backend_types();
+
+  /// Collect aux info from all registered backends.
+  /// @return Map of backend name to aux info string.
```
hidmic: @nvcyc meta: aux info is not particularly descriptive. What kind of metadata do you have in mind?

rosidl_buffer_backend_registry/src/buffer_backend_loader.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include <string>
+
+#include "rcutils/logging_macros.h"
+#include "rosidl_buffer_backend_registry/buffer_backend_registry.hpp"
+#include "rosidl_typesupport_fastrtps_cpp/buffer_serialization.hpp"
+
+namespace rosidl_buffer_backend_registry
+{
+
+static const char * kLoggerName = "rosidl_buffer_backend_registry";
```
hidmic: @nvcyc nit: 
```suggestion
static constexpr std::string_view kLoggerName = "rosidl_buffer_backend_registry";
```

rosidl_buffer_backend_registry/src/buffer_backend_loader.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+void initialize_buffer_backends()
+{
+  auto & registry = BufferBackendRegistry::get_instance();
+  registry.load_plugins();
+
+  auto & backend_ops = rosidl_typesupport_fastrtps_cpp::get_backend_descriptor_ops();
```
hidmic: @nvcyc meta: hmm, `rosidl` packages depending on `rosidl_typesupport_fastrtps_cpp` is... unusual. I actually don't know if `bloom` will like the circular dependency at the repository level.

As a counterproposal, consider making this a reusable function that can called by middlewares to configure backends for the typesupport plugins those need instead of blanket initializing this here for FastRTPS only.

rosidl_buffer_backend_registry/src/buffer_backend_registry.cpp
```diff
@@ -0,0 +1,257 @@
...
+  // Always include CPU backend
+  backend_types.push_back("cpu");
+
+  // Get additional backends from loaded plugins
+  for (const auto & pair : backends_) {
+    auto & backend = pair.second;
+    if (backend) {
+      std::string backend_type = backend->get_backend_type();
+      // Avoid duplicating CPU if it's in the registry
+      if (backend_type != "cpu") {
```
hidmic: @nvcyc will CPU always be in the registry? Do we need all the special casing?

rosidl_buffer_backend_registry/src/buffer_backend_registry.cpp
```diff
@@ -0,0 +1,257 @@
...
+    load_plugins();
+  }
+
+  std::vector<std::string> backend_types;
+
+  // Always include CPU backend
+  backend_types.push_back("cpu");
+
+  // Get additional backends from loaded plugins
+  for (const auto & pair : backends_) {
```
hidmic: @nvcyc nit^2: structured assignment FTW
```suggestion
  for (const auto & [_, backend] : backends_) {
```
Same elsewhere.

rosidl_buffer_backend_registry/src/buffer_backend_registry.cpp
```diff
@@ -0,0 +1,257 @@
...
+std::vector<std::string> BufferBackendRegistry::get_common_backends(
+  const std::vector<std::string> & a,
+  const std::vector<std::string> & b)
+{
+  std::vector<std::string> common;
+  for (const auto & backend_a : a) {
+    for (const auto & backend_b : b) {
+      if (backend_a == backend_b) {
+        bool already_added = false;
+        for (const auto & c : common) {
```
hidmic: @nvcyc fyi: there's `std::any_of`

rosidl_buffer_backend_registry/src/buffer_backend_registry.cpp
```diff
@@ -0,0 +1,257 @@
...
+}
+
+void BufferBackendRegistry::clear_global_state()
+{
+  // CRITICAL: Clear backends_ map to release shared_ptr to plugin instances
+  // before ClassLoader is destroyed.
+  backends_.clear();
+
+  // Note: Global serialization maps in rosidl_typesupport_fastrtps_cpp are cleared
+  // by rmw_shutdown() which is called before this destructor runs.
```
hidmic: @nvcyc this I think reinforces the idea that tying the lifetime of backends and middleware may be best.

rosidl_buffer_backend_registry/include/rosidl_buffer_backend_registry/buffer_backend_registry.hpp
```diff
@@ -0,0 +1,132 @@
...
+  /// @param existing_endpoints List of existing endpoints for grouping decisions.
+  /// @param backend_endpoint_groups In/out parameter for backend-specific grouping information.
+  /// @param endpoint_supported_backends Backend aux info from the discovered endpoint.
+  /// @return Map of backend name to compatibility boolean.
+  ROSIDL_BUFFER_BACKEND_REGISTRY_PUBLIC
+  std::unordered_map<std::string, bool> notify_endpoint_discovered(
+    const rmw_topic_endpoint_info_t & endpoint_info,
+    const std::vector<rmw_topic_endpoint_info_t> & existing_endpoints,
+    std::unordered_map<std::string, std::vector<std::set<uint32_t>>> & backend_endpoint_groups,
+    const std::unordered_map<std::string, std::string> & endpoint_supported_backends);
```
hidmic: @nvcyc meta: these are odd APIs for a registry, does it hint to the need for another object? It's _mostly_ call forwarding to the backend collection.

hidmic (commented):
First pass.

rosidl_buffer_backend_registry/src/buffer_backend_registry.cpp
```diff
@@ -0,0 +1,257 @@
...
+          "Buffer backend plugin discovery error: %s",
+          e.what());
+      }
+
+      plugins_loaded_ = true;
+    });
+}
+
+std::vector<std::string> BufferBackendRegistry::get_backend_names() const
+{
```
mjcarroll: The function documentation seems to indicate that this is thread safe, but I don't see any protection here.  Is the idea that this should not be mutating outside of the initial run?  Is there a chance that more backends plugins could be loaded after?

rosidl_buffer_backend_registry/src/buffer_backend_registry.cpp
mjcarroll: We discussed in the PMC meeting this week that we should probably be storing these things in the context rather than a singleton, and I think that would clear up some of the logic here?

rosidl_buffer_backend_registry/src/buffer_backend_loader.cpp
nvcyc: I've moved `buffer_backend_loader.cpp/.hpp` to RMWs so now `rosidl_buffer_backend_registry` doesn't depend on `rosidl_typesupport_fastrtps_cpp`. For now each RMW (`rmw_fastrtps_cpp` and `rmw_zenoh`) has its own copy of `buffer_backend_loader.cpp/.hpp` (through https://github.com/ros2/rmw_fastrtps/pull/867/changes/9ac694a6d43662a18d3cbc530f59859b5525023f and https://github.com/ros2/rmw_zenoh/pull/930/changes/807604b0cd689ec853ca9402437a2cd0f58446a1)  to keep dependency simple. We will discuss whether we want to reduce this duplication in their PRs (https://github.com/ros2/rmw_fastrtps/pull/867 and https://github.com/ros2/rmw_zenoh/pull/930) if needed.

hidmic: Aha, you didn't like the idea of dedupllcating via `BackendOps`. That's fine.

rosidl_buffer_backend_registry/include/rosidl_buffer_backend_registry/buffer_backend_registry.hpp
nvcyc: The aux info is supposed to be generic as it is extra backend-specific information (represented in a string) a backend would like to share with other endpoints. So a backend can optionally put anything meaningful to it. For example, a GPU-based backend may add IDs of GPUs and host information that it has access to in the aux info so the same type of backends in other endpoints can determine how to export/import the memory for/from this endpoint based on this information.

Another name option is "backend metadata", or do you have some suggestions?

rosidl_buffer_backend_registry/include/rosidl_buffer_backend_registry/buffer_backend_loader.hpp
nvcyc: Relocated this file and hence we don't depend on FastRTPS in this package.

rosidl_buffer_backend_registry/src/buffer_backend_registry.cpp
nvcyc: The design intent is that `load_plugins()` is called once (guarded by `std::call_once`) during RMW initialization, and after that the `backends_` map is effectively immutable -- no new plugins are loaded dynamically after initialization.

rosidl_buffer_backend_registry/src/buffer_backend_registry.cpp
nvcyc: Updated based on your suggestions.

rosidl_buffer_backend_registry/src/buffer_backend_registry.cpp
nvcyc: Thanks! Updated based on the suggestion.

rosidl_buffer_backend_registry/src/buffer_backend_registry.cpp
nvcyc: Yes, we always put "cpu" in the registry as default installed backend. This part of the code has been updated so we don't need the special casing here.

rosidl_buffer_backend_registry/include/rosidl_buffer_backend_registry/buffer_backend_registry.hpp
hidmic: Backend metadata or backend info sound good to me.

hidmic (approved):
Barring aux info naming, this seems correct.

rosidl_buffer_backend_registry/include/rosidl_buffer_backend_registry/buffer_backend_registry.hpp
nvcyc: Ok. I've pushed new commits to rename "aux info" to "backend metadata" in all places (functions/variables/comments).

rosidl_buffer_backend_registry/include/rosidl_buffer_backend_registry/buffer_backend_registry.hpp
nvcyc: Agreed. These backend utility functions are now moved to a separate util header [backend_utils.hpp‎](https://github.com/ros2/rosidl/pull/944/changes#diff-0ad01bc4f56b034b18a2002f0e788d2fadbbd9b37744fcdc77450c3a8c515a06).

rosidl_buffer_backend_registry/src/buffer_backend_loader.cpp
nvcyc: This part has been moved to the RMW. Yet, using `constexpr std::string_view` would mean that it needs to be converted with `kLoggerName.data()` when passing to `RCUTILS_LOG_INFO_NAMED()` while `const char *` can be passed as it is. Therefore keeping it `const char *` might be a good choice here.

hidmic (approved):

rosidl_buffer_backend_registry/src/buffer_backend_registry.cpp
nvcyc: Thanks for the comments.

As you suggested, I've restructured the buffer backend registry and RMW layer so that the registry is no longer a singleton.

The registry, the instantiated backends and relevant descriptor ops and serializers are now stored under the RMW's context, so the backend buffer support is now fully contained within a RMW instance.

rosidl_buffer_backend_registry/src/buffer_backend_loader.cpp
nvcyc: I've restructured the design to remove the singleton registry and global descriptor ops/serializers.

As described in this [response](https://github.com/ros2/rosidl/pull/944#discussion_r3007178419) below, a registry instance, backend instances and those ops are now stored in a RMW context. Therefore it makes sense to let the RMW handle the initialization altogether [here](https://github.com/ros2/rmw_fastrtps/pull/867/changes#diff-3058005e6377b04163ef10abe4c6d7f41211632561fec41bb081db1574eefd51R36).

mjcarroll (approved):

nvcyc:
Pulls: ros2/rosidl#944
Gist: https://gist.githubusercontent.com/nvcyc/b5146169437bac2f8205e8d36149b312/raw/16d2d0c4118905ab27327d1e68c6c6baa825d220/ros2.repos
BUILD args: --continue-on-error --packages-above-and-dependencies rosidl_buffer_backend_registry
TEST args:  --packages-above rosidl_buffer_backend_registry
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18774
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28463/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21327/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8315/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27331/)

nvcyc:
Pulls: ros2/rosidl#944
Gist: https://gist.githubusercontent.com/nvcyc/97b010866c790660160b6b3315ff4f10/raw/16d2d0c4118905ab27327d1e68c6c6baa825d220/ros2.repos
BUILD args: --continue-on-error --packages-above-and-dependencies rosidl_buffer_backend_registry
TEST args:  --packages-above rosidl_buffer_backend_registry
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18791
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28480/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21344/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8332/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27348/)

ahcorde (approved):

PR #957, opened 2026-04-10
Make codegen of rosidl_generator_c deterministic for jazzy

marlamb:
Applying the patch merged in #846 to `rolling` branch also to `jazzy` to
keep the implementations as close as possible.

### DIFF
--- a/rosidl_generator_c/resource/full__description.c.em
+++ b/rosidl_generator_c/resource/full__description.c.em
@@ -1,5 +1,6 @@
 @# Included from rosidl_generator_c/resource/idl__description.c.em
 @{
+from collections import OrderedDict
 from rosidl_generator_c import escape_string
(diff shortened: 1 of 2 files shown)


fujitatomoya (approved):
i think that that is okay to backport this fix to jazzy, i would like to have another review before merge.

fujitatomoya:
actually maybe if we do this, creating backports PR by mergifyio from https://github.com/ros2/rosidl/pull/846 to kilted and jazzy would be better to track the history.

marlamb:
> actually maybe if we do this, creating backports PR by mergifyio from #846 to kilted and jazzy would be better to track the history.

In case you decide so, feel free to close this PR. I don't mind, as long as the functionality becomes available :smiley: .

PR #958, opened 2026-04-20
Clarify return of get_descriptor_type_support() in rosidl_buffer_backend

nvcyc:
## Description

This pull request updates description for the function `get_descriptor_type_support()` in the `BufferBackend` class to clarify that the returned type support handle from the function should be generic to keep the backend plugin RMW-agnostic. This will match description in a documentation introducing `BufferBackend` to be added to ROS2 documentation.

No, this change only touches comments for a funciton.

No.


hidmic (approved):
💯

PR #961, opened 2026-04-29
Add include guards to generated IDL

### LINKED ISSUE #911: `rosidl_adapter` does not emit include guards
Converting a `.msg` to ROS IDL using `rosidl_adapter` does not add include guards. This means that, after preprocessing (using GCC for example), definitions may appear multiple times in the source. This is a problem for using OMG IDL tooling with ROS IDL. According to OMG IDL v4.2:

- Section 7.5.2 "Scoping Rules and Name Resolution" states "An identifier can only be defined once in a scope."
- Section 7.5.3 "Special Scoping Rules for Type Names" states "Once a type has been defined anywhere within the scope of a module, interface or value type, it may not be redefined except within the scope of a nested module, interface or value type, or within the scope of a derived interface or value type."

I'm not sure whether this is a deliberate difference, but if it is, it should be documented.

Lidang-Jiang:
## Description

Fixes #911.

This adds include guards to IDL generated by `rosidl_adapter` for `.msg`, `.srv`, and `.action` inputs. The guard name is derived from the package name and relative input file path, with non-alphanumeric characters normalized to underscores.

This also teaches `rosidl_parser` to accept the preprocessor lines emitted by the generated guarded IDL (`#ifndef`, `#define`, and `#endif`) and updates the generated IDL test fixtures.

Yes. Generated IDL now includes include guards, so downstream consumers that include generated IDL more than once can avoid duplicate definition issues. The generated message, service, and action interfaces otherwise keep the same IDL content.

Yes. I used OpenAI Codex (GPT-5) to help draft and update this pull request and its tests.

### Additional Information

The previous `Before / After` section was intended to show the concrete generated IDL output change for `test_msgs/msg/Test.msg`: before this PR, the generated IDL started directly with the `module test_msgs` declaration; after this PR, the same generated IDL is wrapped by the new `#ifndef` / `#define` / `#endif` include guard. The parser regression test confirms that guarded IDL still parses.

Tests run locally:

```bash
PYTHONPATH=rosidl_parser python3 -m pytest rosidl_parser/test/test_parser.py -q
```

Output:

```text
...........                                                              [100%]
11 passed in 0.68s
```

```bash
PYTHONPATH=rosidl_adapter:rosidl_parser python3 -m pytest rosidl_adapter/test/test_parse_message_string.py rosidl_adapter/test/test_parse_service_string.py rosidl_adapter/test/test_parse_action_string.py -q
```

Output:

```text
.......                                                                  [100%]
7 passed in 0.02s
```

```bash
git diff --check
```

Output:

```text
<no output>
```

### DIFF
--- a/rosidl_adapter/rosidl_adapter/resource/action.idl.em
+++ b/rosidl_adapter/rosidl_adapter/resource/action.idl.em
@@ -6,10 +6,13 @@
 import re
 from rosidl_adapter.msg import get_include_file
 
-include_guard = re.sub(r'[^0-9A-Za-z]+', '_', f'{pkg_name}_{relative_input_file}')
-include_guard = include_guard.strip('_').upper()
-if not include_guard or not include_guard[0].isalpha():
-    include_guard = f'ROSIDL_{include_guard}'
+include_guard_parts = [
+    re.sub(r'[^0-9A-Za-z]+', '_', part).strip('_').upper()
+    for part in (pkg_name, *relative_input_file.parts)
+]
+assert all(include_guard_parts)
+include_guard = '__'.join(include_guard_parts)
+assert include_guard[0].isalpha()
 include_files = set()
 fields = action.goal.fields + action.result.fields + action.feedback.fields
 for field in fields:
@@ -20,12 +23,10 @@ for field in fields:
 #ifndef @(include_guard)
 #define @(include_guard)
 
-@[if include_files]@
 @[for include_file in sorted(include_files)]@
 #include "@(include_file)"
 @[end for]@
 
-@[end if]@
 module @(pkg_name) {
   module action {
 @{
--- a/rosidl_adapter/rosidl_adapter/resource/action.idl.em
+++ b/rosidl_adapter/rosidl_adapter/resource/action.idl.em
@@ -3,7 +3,13 @@
 // generated code does not contain a copyright notice
 
 @{
+import re
 from rosidl_adapter.msg import get_include_file
+
+include_guard = re.sub(r'[^0-9A-Za-z]+', '_', f'{pkg_name}_{relative_input_file}')
+include_guard = include_guard.strip('_').upper()
+if not include_guard or not include_guard[0].isalpha():
+    include_guard = f'ROSIDL_{include_guard}'
 include_files = set()
 fields = action.goal.fields + action.result.fields + action.feedback.fields
 for field in fields:
@@ -11,10 +17,15 @@ for field in fields:
     if include_file is not None:
         include_files.add(include_file)
 }@
+#ifndef @(include_guard)
+#define @(include_guard)
+
+@[if include_files]@
 @[for include_file in sorted(include_files)]@
 #include "@(include_file)"
 @[end for]@
 
+@[end if]@
 module @(pkg_name) {
   module action {
 @{
@@ -37,3 +48,5 @@ TEMPLATE(
 }@
   };
 };
+
+#endif  // @(include_guard)
--- a/rosidl_parser/rosidl_parser/grammar.lark
+++ b/rosidl_parser/rosidl_parser/grammar.lark
@@ -3,7 +3,8 @@
 // 7.2.3 Identifiers
 // 7.2.6 Literals
 // 7.3 Preprocessing
-//   only #include and simple include guard directives are supported
+//   only #include and simple include guard directives
+//   (#ifndef / #define / #endif) are supported
 // 7.4.1 Building Block Core Data Types
 // 7.4.15.4.2 Applying Annotations
 //   Only supported on modules, structs, members, enums and enum identifiers
--- a/rosidl_parser/rosidl_parser/grammar.lark
+++ b/rosidl_parser/rosidl_parser/grammar.lark
@@ -3,7 +3,7 @@
 // 7.2.3 Identifiers
 // 7.2.6 Literals
 // 7.3 Preprocessing
-//   only #include is supported
+//   only #include and simple include guard directives are supported
 // 7.4.1 Building Block Core Data Types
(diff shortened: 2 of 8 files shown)


fujitatomoya:
same here, did you use generate AI or this PR is created by Agent AI? please use PR template instead.
test looks good to me, can you explain `Before/After` what it is trying to show here?

Lidang-Jiang:
Thanks, updated the PR description to use the ROS 2 PR template and filled in the Generative AI section. The previous `Before / After` section was intended to show the generated IDL output change: before this PR the IDL started directly with the `module` declaration, and after this PR the same generated IDL is wrapped with the new `#ifndef` / `#define` / `#endif` include guard. I moved that explanation into the template’s Additional Information section.

rosidl_adapter/rosidl_adapter/resource/action.idl.em
```diff
@@ -3,18 +3,29 @@
 // generated code does not contain a copyright notice
 
 @{
+import re
 from rosidl_adapter.msg import get_include_file
+
+include_guard = re.sub(r'[^0-9A-Za-z]+', '_', f'{pkg_name}_{relative_input_file}')
```
christophebedard: I've always used double underscores to separate package name/directory name/file name. That way the file hierarchy is easier to visually extract from the include guard. I think it would be worth using that here too.

For example:

```c
#ifndef TEST_MSGS__ACTION__TEST_ACTION
```

Same for `msg.idl.em` and `srv.idl.em`.

rosidl_parser/rosidl_parser/grammar.lark
```diff
@@ -3,7 +3,7 @@
 // 7.2.3 Identifiers
 // 7.2.6 Literals
 // 7.3 Preprocessing
-//   only #include is supported
+//   only #include and simple include guard directives are supported
```
christophebedard:
```suggestion
//   only #include and simple include guard directives (#ifndef / #define / #endif) are supported
```

Please double-check the line length.

rosidl_adapter/rosidl_adapter/resource/action.idl.em
```diff
@@ -3,18 +3,29 @@
...
+include_guard = include_guard.strip('_').upper()
+if not include_guard or not include_guard[0].isalpha():
+    include_guard = f'ROSIDL_{include_guard}'
```
christophebedard: This feels dangerous. When could `not include_guard` return `True`? Only if it's empty (IIUC); when would `include_guard` be empty here? In that case, `include_guard` would be `'ROSIDL_'`. If this happens with multiple messages that get includes in the same file, the latter includes will not produce/include anything, most likely resulting in weird errors.

I would either just remove `not include_guard` here, or make it an assert so that we don't continue if `include_guard` is empty, although I don't see how `include_guard` could be empty.

Same for `msg.idl.em` and `srv.idl.em`.

rosidl_adapter/rosidl_adapter/resource/action.idl.em
```diff
@@ -3,18 +3,29 @@
...
 fields = action.goal.fields + action.result.fields + action.feedback.fields
 for field in fields:
     include_file = get_include_file(field.type)
     if include_file is not None:
         include_files.add(include_file)
 }@
+#ifndef @(include_guard)
+#define @(include_guard)
+
+@[if include_files]@
```
christophebedard: This if isn't really necessary, so I would remove it. Same for the `@[end if]@ below.

Same for `msg.idl.em` and `srv.idl.em`.

rosidl_parser/test/test_parser.py
```diff
@@ -92,6 +93,35 @@ def test_message_parser_includes(message_idl_file: IdlFile) -> None:
     assert includes[1].locator == 'pkgname/msg/OtherMessage.idl'
 
 
+def test_message_parser_include_guard() -> None:
+    content = parse_idl_string("""
+#ifndef _ROSIDL_PARSER_MSG_GUARDED_IDL
+#define _ROSIDL_PARSER_MSG_GUARDED_IDL
```
christophebedard: I'm curious, where is this leading underscore coming from? The code wouldn't add one AFAIK.

christophebedard (changes requested):

rosidl_adapter/rosidl_adapter/resource/action.idl.em
Lidang-Jiang: Good catch, fixed in 7d4877d. The generated guards now use double underscores between package, directory, and file-name components, e.g. TEST_MSGS__ACTION__TEST_ACTION.

rosidl_parser/rosidl_parser/grammar.lark
Lidang-Jiang: Fixed in 7d4877d by wrapping the comment line.

rosidl_adapter/rosidl_adapter/resource/action.idl.em
Lidang-Jiang: Agreed, fixed in 7d4877d. I removed the fallback and now assert that all sanitized guard components are non-empty and that the final guard starts with a letter.

rosidl_adapter/rosidl_adapter/resource/action.idl.em
Lidang-Jiang: Fixed in 7d4877d. I removed the unnecessary include_files conditional in the msg, srv, and action templates.

rosidl_parser/test/test_parser.py
Lidang-Jiang: Good point, fixed in 7d4877d. The parser regression test now uses ROSIDL_PARSER__MSG__GUARDED_IDL without the leading underscore.

PR #965, opened 2026-05-16
Re-evaluate topic type name and type hash

### LINKED ISSUE #1682: Re-evaluate topic type name and type hash
In ROS 2, each publisher, subscription, and service is associated with a message type name (e.g., sensor_msgs/msg/Image) and a corresponding type hash (e.g., RIHS01_...). These identifiers are crucial for ensuring type safety and compatibility across a distributed system. However, the current implementation, where type names can be verbose and type hashes (specifically RIHS01 using SHA256) are a fixed 71 bytes, results in lengthy strings.

When combined with fully qualified topic or service names and potentially domain IDs, the total string length for identifying an endpoint can become substantial. For instance, in use cases like rmw_zenoh_cpp with zenoh-pico on microcontrollers, the Zenoh topic key expression incorporates domain_id/topic_name/message_type/RIHS01_hash, directly reflecting this combined length.

The design decision for type hashing (REP-2011) opted for SHA256 for its cryptographic strength, despite suggestions for shorter hashes like MD5 or SHA1, acknowledging a trade-off. While ensuring robustness, this contributes to the length issue impacting embedded use cases.

For ROS 2 to be more pervasive in deeply embedded systems, mechanisms to reduce the overhead of type names and hashes are needed. This might involve exploring:

* Shorter hash options or configurable hash lengths for resource-constrained environments.
* Numeric or compressed representations for type names/hashes.
* Alternative discovery or type negotiation mechanisms that minimize the transmission of long string identifiers.
* Compile-time options to optimize string representations for specific embedded targets.

Ref: https://www.reddit.com/r/ROS/comments/1ix5u7q/mapping_ros2_to_small_embedded_systems_why_are/

v8p1197:
## Description

Fixes https://github.com/ros2/ros2/issues/1682.

This PR introduces a new hash scheme based on BLAKE2.

If users want to build messages with the new scheme, they should specify the `-DROSIDL_GENERATOR_USE_RIHS02=ON` flag (see below).

### Additional Information

This PR can be tested in two ways:

1. Formal tests:

  ```bash
  mkdir ~/ros2_ws/
  cd ~/ros2_ws/
  git clone https://github.com/v8p1197/rosidl.git src/rosidl/
  colcon build
  colcon test
  ```

2. Building a custom message:

  ```bash
  mkdir ~/ros2_ws/
  cd ~/ros2_ws/
  git clone https://github.com/v8p1197/rosidl.git src/rosidl/
  colcon build
  source install/setup.bash
  git clone https://github.com/v8p1197/my_test_msgs.git src/my_test_msgs/
  colcon build --packages-select my_test_msgs --cmake-clean-cache --cmake-args -DROSIDL_GENERATOR_USE_RIHS02=OFF
  cat build/my_test_msgs/rosidl_generator_type_description/my_test_msgs/msg/MyMessage.json | grep hash_string
  # Expected output: RIHS01_aa6fa4c8d287fd38ac88d666c540b4485a93f0d6118911d513528b77f1eaf27c
  colcon build --packages-select my_test_msgs --cmake-clean-cache --cmake-args -DROSIDL_GENERATOR_USE_RIHS02=ON
  cat build/my_test_msgs/rosidl_generator_type_description/my_test_msgs/msg/MyMessage.json | grep hash_string
  # Expected output: RIHS02_17a6e69ee67f1252c4fc5b1ab5173340
  ```


fujitatomoya (commented):
blake2 would be our option here, out of curiosity do we have any discussion on PMC meeting including pros/cons?

v8p1197:
Hi @fujitatomoya, personally I don't usually take part in PMC meetings, but a discussion on this hashing scheme is reported in the referenced issue.
For sure, a pro is the shortest encoded string length.

mjcarroll:
Neat!  I think the idea was that we needed a much shorter string length so that we don't use excessive amounts of memory in resource constrained devices, as reported in the issue upstream.

Changing this hash will have some impact on previously recorded bag files, so we probably want to make sure that we get it correct and not introduce a lot of churn around that.

Rather than having discussion at the PMC meetings, I will just flag it in the next meeting so that people know to take a look at it.

mjcarroll:
Ah, but on closer inspection, you made it opt in, so maybe that takes care of some of the immediate concern there.

v8p1197:
Hi @mjcarroll, yes I made this flag optional but `ON` by default. Do you suggest changing it to `OFF`?

Is there anything else to do to address the referenced issue?

mjcarroll:
Thanks for your patience @v8p1197, and sorry for the long delay in getting back to you.

My preference would be yes, please flip the default to `OFF`.

One big question would be do we also need to update `rcl_calculate_type_hash` and gate it accordingly?  I think the implementations probably need to be kept in sync?  I will review further here, but that's my first observation.

PR #981, opened 2026-07-19
Avoid wstring_convert for ASCII UTF-16 YAML output

Tobias-Fischer:
This PR is part of an effort to contribute RoboStack downstream patches back upstream.

Origin: RoboStack `patch/ros-rolling-rosidl-runtime-cpp.osx.patch`, authored by Daisuke Nishimatsu.

Best-guess rationale: std::wstring_convert and codecvt are deprecated in C++17 and can be problematic on newer standard libraries; ASCII char16_t values can be emitted directly without using the converter.

### DIFF
--- a/rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
+++ b/rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
@@ -15,7 +15,6 @@
 #ifndef ROSIDL_RUNTIME_CPP__TRAITS_HPP_
 #define ROSIDL_RUNTIME_CPP__TRAITS_HPP_
 
-#include <codecvt>
 #include <cstdint>
 #include <iomanip>
 #include <iosfwd>
@@ -129,14 +128,12 @@ inline void value_to_yaml(const std::string & value, std::ostream & out)
 inline void value_to_yaml(const std::u16string & value, std::ostream & out)
 {
   out << "\"";
-  std::wstring_convert<std::codecvt_utf8_utf16<char16_t>, char16_t> convert;
   auto flags = out.flags();
   size_t index = 0;
   while (index < value.size()) {
     uint_least16_t character = static_cast<uint_least16_t>(value[index]);
     if (!(character & 0xff80)) {  // ASCII
-      std::string character_as_string = convert.to_bytes(character);
-      out << std::hex << character_as_string.c_str();
+      out << static_cast<char>(character);
     } else if (!(character & 0xff00)) {  // only 1 byte set
       out << "\\x" << std::hex << std::setw(2) << std::setfill('0') << \
         character;


Tobias-Fischer:
FYI @wep21 and @traversaro, this upstream PR comes from the RoboStack patch upstreaming tracker: https://github.com/RoboStack/robostack.github.io/issues/16.

Origin patch: RoboStack `patch/ros-rolling-rosidl-runtime-cpp.osx.patch`, authored by Daisuke Nishimatsu.

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
```diff
@@ -129,14 +129,20 @@ inline void value_to_yaml(const std::string & value, std::ostream & out)
 inline void value_to_yaml(const std::u16string & value, std::ostream & out)
 {
   out << "\"";
+#if __cplusplus < 201703L
```
fujitatomoya: Why keep the old path at all? ROS 2 humble requires C++17 as the minimum standard, so this branch is dead code in every supported configuration?

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
```diff
@@ -129,14 +129,20 @@ inline void value_to_yaml(const std::string & value, std::ostream & out)
...
   std::wstring_convert<std::codecvt_utf8_utf16<char16_t>, char16_t> convert;
+#endif
   auto flags = out.flags();
   size_t index = 0;
   while (index < value.size()) {
     uint_least16_t character = static_cast<uint_least16_t>(value[index]);
     if (!(character & 0xff80)) {  // ASCII
+#if __cplusplus < 201703L
```
fujitatomoya: does this work with Windows?

Tobias-Fischer: @traversaro do you have any idea?

traversaro: It depends on the command line options passed to the compiler (see https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170), so I guess in the default build that is not passed, so `__cplusplus` as a really old value, and so teh old variant is always used.

rosidl_runtime_cpp/include/rosidl_runtime_cpp/traits.hpp
traversaro: I agree with @fujitatomoya, we can probably just drop the back-compat code, and just keep the one working in C++17 .

Tobias-Fischer:
Hi @fujitatomoya - thanks for the review. I removed the legacy code. Codex tells me that the new code should work fine on Windows as well. It would be great if you could take another look.

wjwwood:
Pulls: ros2/rosidl#981
Gist: https://gist.githubusercontent.com/wjwwood/bb89de8e3c31295da3352025855fb2fe/raw/f0bb693f4aee01b868e284eec366273f53181f45/ros2.repos
BUILD args:  --packages-above-and-dependencies rosidl_runtime_cpp
TEST args:  --packages-above rosidl_runtime_cpp
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/20083
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30074/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22841/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9959/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28994/)

wjwwood (approved):
lgtm, and sounds like @fujitatomoya's concerns were addressed. I started CI.

wjwwood:
Actually I just noticed that https://github.com/ros2/rosidl/pull/982 addressed this, so I think can be closed. I canceled CI.

PR #995, opened 2026-09-09
Handle comment delimiters in generated C headers

### LINKED ISSUE #994: `rosidl_generator_c`: `*/` in a message or constant comment closes the generated `/** */` block and breaks compilation
### Generated by Generative AI

Claude-code using  Fable 5.1

### Operating System:

``Ubuntu 24.04 (Docker)```

### ROS version or commit hash:

Kilted (`rosidl_generator_c` 4.x);

### RMW implementation (if applicable):

n/a

### RMW Configuration (if applicable):

_No response_

### Client library (if applicable):

n/a

### 'ros2 doctor --report' output

<details><summary>ros2 doctor --report</summary>

```console
<COPY OUTPUT HERE>
```
</details>

### Steps to reproduce issue

`msg/Foo.msg` in any interface package:

```text
# See the definitions under other_pkg/**/action for the field layout.
float32 x
```

```bash
colcon build --packages-select my_msgs
```

### Expected behavior

The comment is documentation; it should be copied into the header without
affecting the C code that follows.

### Actual behavior

```text
foo__struct.h:27:67: warning: "/*" within comment [-Wcomment]
foo__struct.h:27:77: error: expected '=', ',', ';', 'asm' or '__attribute__' before ')' token
```

followed by hundreds of cascading errors: the `*/` inside `**/` terminates the
block comment early and the tail of the line (`action for the field layout.`)
is compiled as C.

### Additional information

`msg__struct.h.em` emits message-level and constant-level comments verbatim
inside a block comment:

```text
@{comments = message.structure.get_comment_lines()}@
@[if comments]@
/**
@[  for line in comments]@
@[    if line]@
  * @(line)
...
 */
```

(same pattern for `constant.get_comment_lines()` a few lines above). Member
comments are emitted as `/// @(line)` and are immune, which is why the same
text is harmless on a field and fatal at message level. The Python generator
and the `.idl` adapter are unaffected (the text lives in a `@verbatim` string
there), so the failure only appears when a C header is compiled.

Two possible fixes:

1. Emit the message/constant comments as `///` line comments like the member
   comments (simplest, no escaping problem at all).
2. Keep the block and neutralise the terminator, e.g.
   `line.replace('*/', '* /')`.

Related: #563 is the same kind of problem one stage earlier (comment text
copied unescaped into the `.idl` annotation string by `rosidl_adapter`); the
two are independent, this one only concerns the C header template.

Miko997:
## Description

Embedded comment delimiters in message or constant documentation can break generated C headers. Escape both `/*` and `*/` inside their existing block comments, retaining documentation and field-comment behavior.

Fixes #994.

These comments now generate valid C, including when documentation ends with a backslash.

Codex (GPT-6) wrote the fix and regression coverage and assisted with local testing and code review.

### Additional Information

Verified failure before and success after with the actual adapter/parser/C generator. Ten generated-header cases pass GCC C11 with `-Wall -Wextra -Werror -Wpedantic`. Added a message fixture to the existing C compilation test.

Local ROS Rolling container: built and tested `rosidl_adapter`, `rosidl_parser`, `rosidl_pycommon`, `rosidl_generator_type_description`, `rosidl_generator_c`, `rosidl_generator_cpp`, and `rosidl_generator_tests`, including package lint checks. All checks passed after installing the missing `fastjsonschema` test dependency.


kscottz:
@shru

fujitatomoya (approved):
lgtm with green CI.

Shru (approved):
Tested locally on Rolling. Reverting the change breaks on `Comments.msg` (as expected), and with the fix the tests pass without any issues. 

Maybe down the line we could switch these docblocks to `///` like member fields so we don't have to change the text with `/ *`, but this works as intended.

@kscottz : Ready for CI whenever you have a moment to trigger it!

Miko997:
@kscottz, following up on the CI request. Both reviewers have approved, and Shru verified the fix locally on Rolling. Could you trigger CI when you have a chance? Thanks!

Shru:
Thank you so much for following up and staying on top of this, @Miko997! Really appreciate it.

Just a quick heads up. [ROSCon](https://roscon.ros.org/2026/) is happening this week, so maintainer activity is a bit paused for a couple of days. We can give it a little nudge once they wrap up the conference.

